PR #20, opened 2024-09-04
19 type declarations

### LINKED ISSUE #19: Investigate using .d.ts files to improve type support

JPMeehan:
Implements type declarations to augment classes. Note: There is *not* filtering for `system` based on the value of `type` due to limitations in how TS discriminates unions.

Closes #19


src/module/data/helpers.mjs
```diff
@@ -6,8 +6,8 @@ const {NumberField, SchemaField} = foundry.data.fields;
  * @returns A Schema with a value and max
  */
 export const barAttribute = (initial = 0) => new SchemaField({
-  value: new NumberField({initial}),
-  max: new NumberField({initial})
+  value: new NumberField({initial, nullable: false}),
+  max: new NumberField({initial, nullable: false})
```
krbz999, line 10: `integer: true`? Presumably these should have a minimum as well

krbz999 (approved):

src/module/data/helpers.mjs
JPMeehan: Realizing I'm only using this twice; Stamina *can* go negative, while Recoveries can't. I'm not sure how many other times we'll be using this helper, maybe I should just declare the properties and call it good.

PR #28, opened 2024-09-07
15 kit data

JPMeehan:
- Edit mode now largely uses the _source property of fields
- Kits now apply their bonuses to characters (still no sheet for kits, have to use console/API)
- Created derived property, abilityBonuses, that abilities will pull from in *their* data prep.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -78,9 +78,6 @@
             "size": {
               "label": "Size"
             },
-            "reach": {
-              "label": "Reach"
-            },
             "stability": {
               "label": "Stability"
             }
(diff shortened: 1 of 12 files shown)


krbz999 (approved):
I don't personally believe the method of disabling fields due to effects is the newest hotness.

Rather, since the sheets have toggles, making use of `_source` when editable would be more than enough. But I think we can consider this when the sheets are eventually overhauled.

JPMeehan:
> I don't personally believe the method of disabling fields due to effects is the newest hotness.
> 
> Rather, since the sheets have toggles, making use of `_source` when editable would be more than enough. But I think we can consider this when the sheets are eventually overhauled.

Most fields just use _source which is why I have that dataset selector. However, I have it as a failsafe for the fields that *are* editable in play mode like current HP and resources.

PR #31, opened 2024-09-08
Kit Bonuses for Abilities

JPMeehan:
Still no UI, but did a lot of data prep work plus various improvements to CONFIG and data model options

### DIFF
--- a/src/module/data/helpers.mjs
+++ b/src/module/data/helpers.mjs
@@ -1,3 +1,5 @@
+import {DSRoll} from "../helpers/rolls.mjs";
+
 const {NumberField, SchemaField, StringField} = foundry.data.fields;
 
 /**
@@ -67,5 +69,37 @@ export class SizeModel extends foundry.abstract.DataModel {
   toString() {
     const letter = this.value === 1 ? this.letter ?? "" : "";
     return this.value + letter;
+  }
+}
+
+/**
+ * Special case StringField which represents a formula.
+ */
+export class FormulaField extends foundry.data.fields.StringField {
+
+  /** @override */
+  static get _defaults() {
+    return foundry.utils.mergeObject(super._defaults, {
+      required: true,
+      deterministic: false
+    });
+  }
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  _validateType(value) {
+    DSRoll.validate(value);
+    if (this.options.deterministic) {
+      const roll = new Roll(value);
+      if (!roll.isDeterministic) throw new Error("must not contain dice terms");
+    }
+    super._validateType(value);
+  }
+
+  /** @override */
+  _applyChangeAdd(value, delta, model, change) {
+    if (value) return value.concat(" + ", delta);
+    else return delta;
   }
 }
--- a/src/module/data/helpers.mjs
+++ b/src/module/data/helpers.mjs
@@ -97,9 +97,27 @@ export class FormulaField extends foundry.data.fields.StringField {
     super._validateType(value);
   }
 
+  /* -------------------------------------------- */
+
   /** @override */
+  _castChangeDelta(delta) {
+    // super just calls `_cast`
+    return this._cast(delta).trim();
+  }
+
+  /**
+   * @param {string} value
+   * @param {string} delta
+   * @param {InstanceType<foundry["abstract"]["DataModel"]>} model
+   * @param {import("../../../foundry/common/types.mjs").EffectChangeData} change
+   * @override
+   */
   _applyChangeAdd(value, delta, model, change) {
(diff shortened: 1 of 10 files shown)


src/module/data/helpers.mjs
```diff
@@ -69,3 +71,35 @@ export class SizeModel extends foundry.abstract.DataModel {
...
+  _validateType(value) {
+    DSRoll.validate(value);
+    if (this.options.deterministic) {
+      const roll = new Roll(value);
+      if (!roll.isDeterministic) throw new Error("must not contain dice terms");
+    }
+    super._validateType(value);
+  }
+
+  /** @override */
```
krbz999, line 100: Missing divider.

src/module/data/helpers.mjs
```diff
@@ -69,3 +71,35 @@ export class SizeModel extends foundry.abstract.DataModel {
...
+  _applyChangeAdd(value, delta, model, change) {
```
krbz999, line 101: I think we can add methods for multiplying and upgrading/downgrading as well - in a follow-up PR.

src/module/data/helpers.mjs
```diff
@@ -69,3 +71,35 @@ export class SizeModel extends foundry.abstract.DataModel {
...
+    if (value) return value.concat(" + ", delta);
```
krbz999, line 102: We should remove and leading sign (and if the leading sign is "-" use that instead of "+").

src/module/data/item/ability.mjs
```diff
@@ -30,9 +32,10 @@ export default class AbilityModel extends BaseItemModel {
     schema.type = new fields.StringField(requiredChoice(config.types, "action"));
     schema.distance = new fields.SchemaField({
       type: new fields.StringField(requiredChoice(config.distances, "self")),
-      primary: new fields.NumberField(),
-      secondary: new fields.NumberField()
+      primary: new fields.NumberField({integer: true, required: true}),
+      secondary: new fields.NumberField({integer: true, required: true})
```
krbz999, line 36: I don't believe `required: true` does anything for a NumberField; it cannot be `undefined`.

src/module/data/item/ability.mjs
```diff
@@ -42,13 +45,13 @@ export default class AbilityModel extends BaseItemModel {
 
     const powerRollSchema = () => ({
       damage: new fields.SchemaField({
-        value: new fields.StringField({validate: DSRoll.validate, validationError: "Must be a valid roll formula"}),
-        type: new fields.StringField()
+        value: new FormulaField(),
+        type: new fields.StringField({nullable: false})
```
krbz999, line 49:
```suggestion
        type: new fields.StringField({required: true})
```

src/module/data/item/ability.mjs
```diff
@@ -42,13 +45,13 @@ export default class AbilityModel extends BaseItemModel {
...
       }),
       ae: new fields.StringField({validate: foundry.data.validators.isValidId}),
       forced: new fields.SchemaField({
         type: new fields.StringField({choices: config.forcedMovement, blank: false}),
-        value: new fields.NumberField(),
+        value: new fields.NumberField({required: true}),
```
krbz999, line 54: same comment here.

src/module/data/item/ability.mjs
```diff
@@ -70,4 +73,100 @@ export default class AbilityModel extends BaseItemModel {
     description.flavor = new foundry.data.fields.StringField({required: true, blank: true});
     return description;
   }
+
+  /** @override */
```
krbz999, line 77: Missing divider.

src/module/data/item/ability.mjs
```diff
@@ -70,4 +73,100 @@ export default class AbilityModel extends BaseItemModel {
...
+  prepareDerivedData() {
+    super.prepareDerivedData();
+
+    if (this.actor && (this.actor.system instanceof CharacterModel)) {
```
krbz999, line 81:
```suggestion
    if (this.actor?.type !== "character") return;
```

src/module/data/item/base.mjs
```diff
@@ -63,6 +63,13 @@ export default class BaseItemModel extends foundry.abstract.TypeDataModel {
     return {};
   }
 
+  /**
+   * Convenient access to the item's actor.
+   */
+  get actor() {
+    return this.parent.actor;
+  }
+
```
krbz999, line 72: Missing dividers, and
```suggestion
  /**
   * Convenient access to the item's actor.
   * @type {DrawSteelActor}
   */
  get actor() {
    return this.parent.actor;
  }
```

krbz999 (approved):

PR #35, opened 2024-09-10
Refine status effect list

### LINKED ISSUE #29: Refine status effect list
- Prune unneeded conditions
- Find icons for new conditions

Taroc0:
This resolves #29 by removing the status effects that might not be needed by DrawSteel and adding some icons for the new status effects introduced by DrawSteel.
I'm not 100% sure if some of the removed status effects might make sense in DrawSteel. Will be easy to add them again by editing the toRemove array in draw-steel.mjs

This is my first GitHub contribution ever. I hope I did everything right.


JPMeehan (approved):
This looks great! Could you also fix the i18n strings in `DRAW_STEEL.conditions` - specifically, the names are missing `.name` at the end.

PR #36, opened 2024-09-12
Refine status effect list

Taroc0:
This resolves #35 to fix the i18n strings in `DRAW_STEEL.conditions`.

I just had do add `.name` to all condition names as @JPMeehan had suggested. Couldn't see anything else to change.


JPMeehan (approved):
Thank you! And I think you noticed - making your changes on a branch that you submit the PR from makes it a lot easier to manage your fork on *your* end since we do squish and merge.

Issue #37, opened 2024-09-12
npm install createSymlinks error


Taroc0:
When I run `npm install` on a newly cloned repo I get an error (did all the setup stuff mentioned in CONTRIBUTING.md)

```
Reforging Symlinks
node:internal/errors:497
    ErrorCaptureStackTrace(err);
    ^

TypeError [ERR_INVALID_ARG_TYPE]: The "cb" argument must be of type function. Received undefined
    at maybeCallback (node:fs:178:3)
    at Module.readFile (node:fs:360:14)
    at file:///home/taroco/Foundry/Data/systems/draw-steel/tools/create-symlinks.mjs:8:23
    at ModuleJob.run (node:internal/modules/esm/module_job:217:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:316:24)
    at async loadESM (node:internal/process/esm_loader:34:7)
    at async handleMainPromise (node:internal/modules/run_main:66:12) {
  code: 'ERR_INVALID_ARG_TYPE'
}
```
Ubuntu 20.04
Node.js v20.9.0
npm 10.8.3

I figured out that there is a synchronous and an async version of `fs`. Will send a PR that might solve this.

JPMeehan:
Yeah might need to change some pieces from `fs.func` to `fs.promises.func`

PR #38, opened 2024-09-12
fix create-symlink error by importing and using promise based fs

Taroc0:
This solves issue #37. At least it worked for me on my machine.

### DIFF
--- a/tools/create-symlinks.mjs
+++ b/tools/create-symlinks.mjs
@@ -1,18 +1,24 @@
 import * as fs from "fs";
+import {promises as fsPromises} from "fs";
 import yaml from "js-yaml";
 import path from "path";
 
 console.log("Reforging Symlinks");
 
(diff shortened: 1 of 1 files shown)


tools/create-symlinks.mjs
```diff
@@ -1,34 +1,40 @@
 import * as fs from "fs";
+import {promises as fsPromises} from "fs";
```
JPMeehan, line 2: I think we can clean up the import handling to a single line - maybe switch to something like `import {promises as fsp, existsSync} from "fs"`? Or just use `fs.promises` instead of `fsPromises` in your function calls and remove this import line.

JPMeehan (changes requested):

JPMeehan (approved):

Taroc0:
You are right. Using `fs.promises` is much cleaner and straight forward. Changed it. Still works for me.

Issue #11, opened 2024-08-30
Add descriptions to CONFIG entries


JPMeehan:
The backer packet license allows the full text, so we can include things like the descriptions of keywords

Taroc0:
As my knowledge on how to build a game system in Foundry is very limited, I could work on this issue.

### What keywords should be described? 
I could start with conditions like dazed, frightend and so on.
All skills could be next.

### What would be the best way to do it?
Would it make sense to add the description to `en.json` first like this:

```json
      "Conditions": {
        "Dazed": {
          "name": "Dazed",
          "description": "While you are dazed, you can do only one thing on your turn: use a maneuver, use an action, or take a move action. You also can’t use triggered actions, free triggered actions,or free maneuvers."
        },
```

And add it to `config.mjs` like this:
```mjs
/**
 * @type {Record<string, {img: string, name: string}>}
 */
DRAW_STEEL.conditions = {
  // bleeding: {},
  dazed: {
    name: "DRAW_STEEL.Effect.Conditions.Dazed.name",
    img: "icons/svg/daze.svg",
    desc: "DRAW_STEEL.Effect.Conditions.Dazed.description"
  },

```

Taroc0:
Sure. That var instead of const is a stupid mistake of mine. I'll push with the next PR.

Does that separate localization pass through relate to Babele or is it something else? Is it ok to begin adding the `description` values now or is it better to wait and figure out that localization pass through first?

JPMeehan:
You can start adding values. To briefly explain, there is a `preLocalize` helper that I copied from dnd5e that allows us to define this CONFIG in `init` and then make a pass updating the values in `i18nInit`. However, the status effects are an exception because we're adding them somewhere else in a way that just iterating over CONFIG.DRAW_STEEL won't help — need to loop through that `CONFIG.statusEffects` and localize the descriptions.

krbz999:
I think it would be better to include a journal in the system and have the status condition point to those pages. Then we can use the `@Embed` enricher. This would mean the text is not copied for every effect but is drawn dynamically from a single journal.

JPMeehan:
So in this case rather than an i18n string we'd have the `description` in the status effects be set to the `@Embed`. I can get the journals set up in a PR.

Also, the spot to override would be `ActiveEffect._fromStatusEffect` which was added in v12.

krbz999:
[...]

PR #42, opened 2024-09-19
Added Conditions journal

Taroc0:
First part for #11 
Added a 'Conditions' journal to the packs/journals folder, containing pages for each condition available in Draw Steel. I thought a separate journal would keep the system journals organised.

Please let me know if this works for you. I'll wait for your feedback before I proceeding.

### DIFF
--- a/draw-steel.mjs
+++ b/draw-steel.mjs
@@ -35,13 +35,13 @@ Hooks.once("init", function () {
     }
   }
 
+  //Remove Status Effects Not Available in DrawSteel
+  const toRemove = ["bleeding", "bless", "burrow", "corrode", "curse", "degen", "disease", "upgrade", "fireShield", "fear", "holyShield", "hover", "coldShield", "magicShield", "paralysis", "poison", "prone", "regen", "restrain", "shock", "silence", "stun", "unconscious", "downgrade"];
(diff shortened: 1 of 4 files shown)


JPMeehan:
Yeah, this makes sense to me. Are these all level 1? Maybe should have the first "Conditions" be level 1 and then the following be level 2. Then, the CONFIG object can point to the UUIDs of each of these journals.

Taroc0:
I'm not sure what you mean by level 1 and level 2. I added a journal called "Conditions". That's on the same level as the other journals (Backer Packet and System Info). The journal "Conditions" consists of ten pages. The first page is a general description of conditions in Draw Steel. The following nine pages are describing one condition at a time. (See screenshot below) All pages are on the same level and I don't know a way to put pages to level 2. The only way to have them at different levels would be to put them on one page and use something like Heading 1 and Heading 2. But that makes referencing them a lot harder, I think.

I afraid I didn't got your point. Sorry.

That's how it looks at the moment:
image

JPMeehan:
When you edit the page, in the top right there's a Level property. You can also see it in the json for the individual pages, it's a top level property.

JPMeehan:
anyways the actual thing I want to see here is the UUID pointers in CONFIG

JPMeehan (approved):

PR #43, opened 2024-09-27
Power roll crits

### LINKED ISSUE #10: Adjust Power Roll Result determination for 19 and 20
Change from Patron packet
> Natural 19 or 20 always gives success with a reward on a test.

> A natural 19 or 20 automatically gives you a tier 3 result as well as being a critical hit now so these things are in parity.

Taroc0:
This fixes #10 

Nat19 and Nat20 are always a crit regardless of the Power Roll type.
Nat19 and Nat20 are always a tier 3 result

### DIFF
--- a/src/module/helpers/rolls.mjs
+++ b/src/module/helpers/rolls.mjs
@@ -230,6 +230,9 @@ export class PowerRoll extends DSRoll {
    */
   get product() {
     if (this._total === undefined) return undefined;
+    // Crits are always a tier 3 result
+    if (this.critical) return 3;
+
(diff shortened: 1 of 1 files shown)


src/module/helpers/rolls.mjs
```diff
@@ -268,7 +271,6 @@ export class PowerRoll extends DSRoll {
    * otherwise returns if the dice total is a 19 or higher
    */
   get critical() {
-    if (this.options.type !== "ability") return null;
```
krbz999, line 271: The above comment should be amended as well.

JPMeehan (approved):

JPMeehan:
Thank you!

PR #44, opened 2024-10-07
Pack and unpack with compendium folders that match real folders in so…

### LINKED ISSUE #41: Pack and unpack with real folders
The current pack/unpack setup has a flat structure and only implements folders inside foundry. While this is functional, it would be even better if content was packed and unpacked according to its actual folder.

Taroc0:
Fixes #41 
Implemented packing and unpacking with subfolders. Every subfolder in a compedium unpacks to a subfolder in the `src/packs/<compediumName>` directory. Every subfolder in `src/packs/<compendiumName>` packs to a folder in the corresponding compedium.

I removed `system-info` and `backer-packet` journals and added a `characteristics` journal inside a `rules` folder. Moved the existing `Conditions` journal to the new `combat` folder. This is just for demonstration purposes. We can still agree on a suitable folder structure.


JPMeehan:
Yeah, I fully expect the actual final structure to change whenever we get a new licensed packet.

tools/pushLDBtoJSON.mjs
```diff
@@ -1,30 +1,87 @@
 import {extractPack} from "@foundryvtt/foundryvtt-cli";
-import {promises as fs} from "fs";
+import * as fs from "fs";
 import path from "path";
 
 const MODULE_ID = process.cwd();
```
JPMeehan, line 5: Rename this too

tools/pullJSONtoLDB.mjs
```diff
@@ -1,14 +1,28 @@
 import {compilePack} from "@foundryvtt/foundryvtt-cli";
 import {promises as fs} from "fs";
+import path from "path";
 
 const MODULE_ID = process.cwd();
```
JPMeehan, line 5: We should probably take this opportunity to rename this variable to `SYSTEM_ID` - I had copied the original code from my 5e content packs.

JPMeehan (commented):
Overall job well done, just some minor edits

JPMeehan (approved):

JPMeehan:
Thank you so much!

PR #46, opened 2024-12-04
Suppress effects with expired durations

JPMeehan:
Leveraging the `isSuppressed` and `active` getters to maybe-cleanly handle effect expiration.

### DIFF
--- a/src/module/documents/active-effect.mjs
+++ b/src/module/documents/active-effect.mjs
@@ -2,10 +2,11 @@ export class DrawSteelActiveEffect extends ActiveEffect {
 
   /**
    * Automatically deactivate effects with expired durations
+   * @type {Boolean}
    */
   get isSuppressed() {
-    if (Number.isNumeric(this.duration?.remaining)) {
-      return this.duration?.remaining <= 0;
+    if (Number.isNumeric(this.duration.remaining)) {
+      return this.duration.remaining <= 0;
     }
     return false;
   }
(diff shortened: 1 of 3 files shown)


src/module/documents/active-effect.mjs
```diff
@@ -1,4 +1,15 @@
 export class DrawSteelActiveEffect extends ActiveEffect {
+
+  /**
+   * Automatically deactivate effects with expired durations
+   */
+  get isSuppressed() {
+    if (Number.isNumeric(this.duration?.remaining)) {
```
krbz999, line 7: Can this be simplified to just `if (this.isTemporary && !this.duration.remaining) return true;` ?

krbz999: Ah I immediately realize why not (effects with just `statuses`).

src/module/documents/active-effect.mjs
```diff
@@ -1,4 +1,15 @@
...
+      return this.duration?.remaining <= 0;
```
krbz999, line 8:
```suggestion
    if (Number.isNumeric(this.duration.remaining)) {
      return this.duration.remaining <= 0;
```

src/module/documents/active-effect.mjs
```diff
@@ -1,4 +1,15 @@
 export class DrawSteelActiveEffect extends ActiveEffect {
+
+  /**
+   * Automatically deactivate effects with expired durations
+   */
```
krbz999, line 5:
```suggestion
  /**
   * Automatically deactivate effects with expired durations
   * @type {Boolean}
   */
```

krbz999 (approved):

PR #50, opened 2024-12-20
Implemented Item Sheets

JPMeehan:
Also included a lot of general revisions, e.g. to the item models, based on packet 3 info. Will need to make further adjustments when Packet 4 drops in next week or so.


krbz999 (approved):
Well this required some thorough testing. LGTM.

PR #69, opened 2024-12-21
Implemented Project Roll

### LINKED ISSUE #65: Add Project Roll
New DSRoll Subclass

JPMeehan:
This roll has enough distinctions from the base power roll I decided to make it its own class

Closes #65

### DIFF
--- a/src/module/helpers/rolls.mjs
+++ b/src/module/helpers/rolls.mjs
@@ -310,3 +310,201 @@ export class PowerRoll extends DSRoll {
     return context;
   }
 }
+
+/**
+ * Special test
+ */
+export class ProjectRoll extends DSRoll {
+  constructor(formula = "2d10", data = {}, options = {}) {
+    super(formula, data, options);
+    foundry.utils.mergeObject(this.options, this.constructor.DEFAULT_OPTIONS, {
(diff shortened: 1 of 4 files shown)


src/module/helpers/rolls.mjs
```diff
@@ -310,3 +310,201 @@ export class PowerRoll extends DSRoll {
...
+  }
+
+  /**
+   * Determines if the natural result was a natural 20
+   * @returns {boolean | null} Null if not yet evaluated
+   */
+  get nat20() {
+    if ((this._total === undefined) || !this.validProjectRoll) return null;
+    return (this.dice[0].total >= 20);
+  }
```
krbz999, line 462: Is "nat 20" a term actually used in DS?

Regardless, could we adopt a convention for getters that ask a question like this - perhaps "isNat20" to make it seem less like a property and more like a consequence of other data on the roll.

krbz999 (approved):

src/module/helpers/rolls.mjs
JPMeehan: It is, although I need to double check if the current rules ever treat it differently from just being a critical (nat 19 or 20)

PR #70, opened 2024-12-21
Implemented Saving Throw roll class

### LINKED ISSUE #67: Add Saving Throw
> If an effect ends with “(save ends)” at the end of its
description, then a creature suffering the effect can make a
saving throw at the end of their turn to remove the effect.
A saving throw represents your luck in shaking off an effect.
You likely already got a chance to resist the effect using
your characteristic score to resist a potency, so now it’s
down to fate. To make a saving throw, roll a 1d10. On a
result of 6 or higher, the effect ends. Otherwise, it continues.

JPMeehan:
image

Closes #67

### DIFF
--- a/src/module/helpers/rolls.mjs
+++ b/src/module/helpers/rolls.mjs
@@ -14,8 +14,8 @@ export class DSRoll extends foundry.dice.Roll {
   /**
    * Helper function to generate render context in use with `static CHAT_TEMPLATE`
    * @param {object} options
-   * @param {string} [options.flavor]     - Flavor text to include
-   * @param {boolean} [options.isPrivate] - Is the Roll displayed privately?
+   * @param {string} [options.flavor]     Flavor text to include
+   * @param {boolean} [options.isPrivate] Is the Roll displayed privately?
    * @returns An object to be used in `renderTemplate`
    */
   async _prepareContext({flavor, isPrivate}) {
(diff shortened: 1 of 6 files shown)


src/module/helpers/rolls.mjs
```diff
@@ -509,3 +509,43 @@ export class ProjectRoll extends DSRoll {
     return context;
   }
 }
+
+export class SavingThrow extends DSRoll {
```
krbz999, line 513: Consider a more descriptive name like `SavingThrowRoll`.

src/module/helpers/rolls.mjs
```diff
@@ -509,3 +509,43 @@ export class ProjectRoll extends DSRoll {
...
+  /**
+   * @param {string} [formula="1d10"]        Default saving throw is a flat 1d10
+   * @param {Record<string, any>} [data]     Roll data
+   * @param {{ flavor?: string }} [options]  Options to modify roll display
+   */
+  constructor(formula = "1d10", data = {}, options = {}) {
+    super(formula, data, options);
+  }
```
krbz999, line 521: Do we foresee any instance of the formula being anything but `1d10`?

src/module/helpers/rolls.mjs
```diff
@@ -509,3 +509,43 @@ export class ProjectRoll extends DSRoll {
...
+    if (this._total === undefined) return undefined;
+    return this.total >= 6;
+  }
+
+  /**
+   * Helper function to generate render context in use with `static CHAT_TEMPLATE`
+   * @param {object} options
+   * @param {string} [options.flavor]     - Flavor text to include
+   * @param {boolean} [options.isPrivate] - Is the Roll displayed privately?
+   * @returns An object to be used in `renderTemplate`
```
krbz999, line 539: Align the formatting of the documentation here (remove the hyphens).

krbz999 (approved):

krbz999:
In the future, it would be good to be able to tie the roll to an effect as well to allow for (optionally) ending the effect immediately on a success as described in the issue.

src/module/helpers/rolls.mjs
JPMeehan: It looks like there are some effects that can give a bonus to saving throws, found a high level Censor feature that gives a +2 bonus to them.

JPMeehan:
> In the future, it would be good to be able to tie the roll to an effect as well to allow for (optionally) ending the effect immediately on a success as described in the issue.

Yeah, I think the next thing I wanted to tackle the base effect model & make some adjustments to stuff like isTemporary as well

src/module/helpers/rolls.mjs
krbz999: > It looks like there are some effects that can give a bonus to saving throws, found a high level Censor feature that gives a +2 bonus to them.

OK. Maybe it would be good to have a follow-up PR for constructing the roll's different parts; with bonuses, whichever modifiers, etc.

JPMeehan: Yeah, once we write functions like "use an ability" I'm sure we'll realize new ergonomics are needed

PR #77, opened 2024-12-22
Implemented system model based duration

### LINKED ISSUE #72: Implement Base Effect Model
At a minimum, need to specify the end condition from `ds.CONFIG.effectEnds`. Should probably also override ActiveEffect#isTemporary to check `system.isTemporary`, so that this `system.ends` property can *also* mark things as temporary without needing fixed durations

Depends on #67

JPMeehan:
Doesn't implement any automatic end handling - that will have to wait for implementing all the combat & initiative handling - but this at least gets all of the display data handled

Closes #72

### DIFF
--- a/src/module/apps/hooks/activeEffectConfig.mjs
+++ b/src/module/apps/hooks/activeEffectConfig.mjs
@@ -1,0 +1,24 @@
+/**
+ * A hook event that fires when the ActiveEffectConfig application is rendered
+ * @param {ActiveEffectConfig} app
+ * @param {JQuery<HTMLElement>} jquery
+ * @param {Record<string, any>} context
+ */
+export function renderActiveEffectConfig(app, [html], context) {
+  /**
+   * @type {import("../../documents/active-effect.mjs").DrawSteelActiveEffect}
+   */
+  const effect = app.document;
+
+  if (effect.type !== "base") return;
+
+  const endOptions = Object.entries(ds.CONFIG.effectEnds).map(([value, {label}]) => ({value, label}));
+
+  const endsInput = effect.system.schema.getField("end").toFormGroup({}, {options: endOptions, value: effect.system.end, blank: ""});
+
+  const durationTab = html.querySelector([".tab[data-tab=\"duration\"]"]);
+
+  durationTab.append(endsInput);
+
+  app.setPosition();
+}
--- a/src/module/apps/hooks/activeEffectConfig.mjs
+++ b/src/module/apps/hooks/activeEffectConfig.mjs
@@ -14,14 +14,18 @@ export function renderActiveEffectConfig(app, [html], context) {
 
   const endOptions = Object.entries(ds.CONFIG.effectEnds).map(([value, {label}]) => ({value, label}));
 
-  const endsInput = effect.system.schema.getField("end").toFormGroup(
+  const endsInput = effect.system.schema.getField("end.type").toFormGroup(
     {},
-    {options: endOptions, value: effect.system.end, blank: ""}
+    {options: endOptions, value: effect.system.end.type, blank: ""}
   );
+
+  const rollInput = effect.system.schema.getField("end.roll").toFormGroup({}, {value: effect.system.end.roll});
+
+  const hrBreak = document.createElement("hr");
 
   const durationTab = html.querySelector([".tab[data-tab=\"duration\"]"]);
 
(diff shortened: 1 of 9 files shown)


src/module/apps/hooks/activeEffectConfig.mjs
```diff
@@ -0,0 +1,24 @@
...
+  /**
+   * @type {import("../../documents/active-effect.mjs").DrawSteelActiveEffect}
+   */
+  const effect = app.document;
+
+  if (effect.type !== "base") return;
+
+  const endOptions = Object.entries(ds.CONFIG.effectEnds).map(([value, {label}]) => ({value, label}));
+
+  const endsInput = effect.system.schema.getField("end").toFormGroup({}, {options: endOptions, value: effect.system.end, blank: ""});
```
krbz999, line 17:
```suggestion
  const endsInput = effect.system.schema.getField("end").toFormGroup(
    {},
    {options: endOptions, value: effect.system.end, blank: ""}
  );
```

src/module/documents/active-effect.mjs
```diff
@@ -16,4 +17,22 @@ export class DrawSteelActiveEffect extends ActiveEffect {
...
+  _prepareDuration() {
+    return this.system._prepareDuration ?? super._prepareDuration();
+  }
+
+  /**
+   * Check if the effect's subtype has special handling, otherwise fallback to normal `duration` and `statuses` check
+   */
+  get isTemporary() {
+    return this.system.isTemporary ?? super.isTemporary;
+  }
```
krbz999, line 37: Since it does not rely solely on the system property, best not mark that as API. Whether an effect is temporary should still be derived from the document class (and whatever that calls):
```suggestion
  get isTemporary() {
    return this.system._isTemporary ?? super.isTemporary;
  }
```

src/module/data/effect/base.mjs
```diff
@@ -6,12 +6,49 @@ export default class BaseEffectModel extends foundry.abstract.TypeDataModel {
...
-      characteristic: new fields.StringField({choices: config.characteristics, blank: false})
+      end: new fields.StringField({choices: Object.keys(config.effectEnds), blank: true, required: true})
+    };
+  }
+
+  /**
+   * An effect is also temporary if it has the `end` property set even though they have indeterminate lengths
+   * @returns {true | null}
+   */
+  get isTemporary() {
```
krbz999, line 25: See my other comment.
```suggestion
  get _isTemporary() {
```

src/module/data/effect/base.mjs
```diff
@@ -6,12 +6,49 @@ export default class BaseEffectModel extends foundry.abstract.TypeDataModel {
...
+    if (this.end) return true;
+    else return null;
```
krbz999, line 27: Any specific reasoning for `null`? Seems to me it can return `false` just fine.

src/module/data/effect/base.mjs
```diff
@@ -6,12 +6,49 @@ export default class BaseEffectModel extends foundry.abstract.TypeDataModel {
...
+  static LOCALIZATION_PREFIXES = ["DRAW_STEEL.Effect.base"];
+
+  /** @override */
   static defineSchema() {
     const fields = foundry.data.fields;
     const config = ds.CONFIG;
     return {
-      end: new fields.StringField({choices: config.effectEnds, blank: false}),
-      characteristic: new fields.StringField({choices: config.characteristics, blank: false})
+      end: new fields.StringField({choices: Object.keys(config.effectEnds), blank: true, required: true})
```
krbz999, line 17: Do we expect this to remain simple enough that a mere StringField is warranted? How about further nesting:
```js
end: {
  type: "...",
}
```
which - I could imagine - may have further properties in the future, e.g., to determine conditions of ending the effect.

krbz999 (approved):
Looks good to me, just some thoughts that you may take or leave.

src/module/data/effect/base.mjs
JPMeehan: The document class is using null coalescing, so `false` would be a rejection of whatever the normal determination is.

src/module/data/effect/base.mjs
JPMeehan: At least in the current rules the conditions are one of the three choices, BUT should maybe include a roll modifier field for the Saving Throw type?

PR #79, opened 2024-12-22
Bugfix: Missed two property adjustments in refactor

JPMeehan:


krbz999 (approved):
Might actually consider making this a SetField.

PR #85, opened 2024-12-24
Updated Status Effects to backer packet 2

### LINKED ISSUE #82: Update status effects
1. Double check that everything is up to date with backer packet 4
2. Consider some kind of ready hook update where we check a reference UUID to then update the `description` of the status effects.

JPMeehan:
Room for future work
- Update the `changes` property of each condition
- Add any necessary automations related to checking if a character has a given status
- Create API for status interactions such as the human's Perserverence trait that makes slow downgrade to 3 instead of 2

Closes #82


draw-steel.mjs
```diff
@@ -77,9 +77,18 @@ Hooks.once("i18nInit", () => helpers.utils.performPreLocalization(CONFIG.DRAW_ST
...
+  // Hooks.on("hotbarDrop", (bar, data, slot) => helpers.macros.createDocMacro(data, slot));
+  for (const e of CONFIG.statusEffects) {
+    if (e.rule) {
+      /** @type {JournalEntryPage} */
+      const page = await fromUuid(e.rule);
+      if (page) {
+        e.description = page.text.content;
+      }
+    }
+  }
```
krbz999, line 91: Why is this? We can just use `@Embed` when the effect is created via `actor.toggleStatusEffect`.

krbz999 (changes requested):

draw-steel.mjs
JPMeehan: Oh that would be a lot cleaner all told

draw-steel.mjs
```diff
@@ -93,9 +93,9 @@ Hooks.once("i18nInit", () => {
 /*  Ready Hook                                  */
 /* -------------------------------------------- */
 
-Hooks.once("ready", function () {
+Hooks.once("ready", async function () {
```
krbz999, line 96:
```suggestion
Hooks.once("ready", function () {
```

src/module/documents/active-effect.mjs
```diff
@@ -1,4 +1,9 @@
 export class DrawSteelActiveEffect extends ActiveEffect {
+  /** @override */
+  static async _fromStatusEffect(statusId, effectData, options) {
+    if (effectData.rule) effectData.description ||= `@Embed[${effectData.rule} inline]`;
+    return new this(effectData, options);
+  }
```
krbz999, line 6:
```suggestion
  static async _fromStatusEffect(statusId, effectData, options) {
    const effect = await super._fromStatusEffect(statusId, effectData, options);
    if (effectData.rule) effect.updateSource({description: `@Embed[${effectData.rule} inline]`});
    return effect;
  }
```

krbz999 (approved):

PR #86, opened 2024-12-24
Migrated characteristics from 3 letters to 1

### LINKED ISSUE #84: Migrate Characteristics from three letters to one
The characteristics have changed from three letter abbreviations like MGT to single letters like M. This requires both adjusting CONFIG and the en.json file

JPMeehan:
Patreon Packet 3 changed the abbreviations, seems to have stuck. Fortunately all the handling adapts pretty well since I wasn't tying stuff to particularly data paths yet.

Closes #84


lang/en.json
```diff
@@ -33,33 +33,33 @@
           },
           "characteristics": {
             "label": "Characteristics",
-            "mgt": {
+            "M": {
```
krbz999, line 36: I think it's a lot cleaner to have internal keys be self-explanatory. We can use the full  labels, `might`, `agility`, etc., for the internal keys and leave the human-readable labels as 1 character.

JPMeehan: Sounds good will adjust

krbz999 (approved):

PR #87, opened 2024-12-25
Languages

### LINKED ISSUE #60: Implement Languages
Tied to #51 

Probably CONFIG object, but also people are likely to want to customize so maybe world setting that's accessible?

JPMeehan:
Possible future work
- Add "hint" property for tooltip usage in multi-checkbox (e.g. for Advancements)
- Add group support

Closes #60


lang/en.json
```diff
@@ -770,6 +769,49 @@
...
+      "Vaniric": "Vaniric",
+      "Vasloria": "Vasloria",
+      "HighRhyvian": "High Rhyvian",
+      "Khamish": "Khamish",
+      "Kheltivari": "Kheltivari",
+      "LowRhivian": "Low Rhivian",
+      "OldVariac": "Old Variac",
+      "Phorialtic": "Phorialtic",
+      "Rallarian": "Rallarian",
+      "Ullorvic": "Ullorvic"
```
krbz999, line 813: Can we sort these alphabetically.

src/module/config.mjs
```diff
@@ -449,7 +449,139 @@ Object.defineProperty(DRAW_STEEL.skills, "optgroups", {
   }
 });
 
-// TODO: Languages
+/**
+ * Configuration information for languages
+ * @type {Record<string, {label: string}>}
+ */
+DRAW_STEEL.languages = {
```
krbz999, line 456: Could be we should grab the `SelectChoices` class from dnd5e. Seems it might come in useful.

krbz999 (approved):

PR #94, opened 2024-12-26
Add characteristics to ability power roll

### LINKED ISSUE #92: Add characteristic(s) to ability power rolls
Need to define which characteristic(s) are used for the power roll. Some abilities allow a choice between multiple. Would need to determine the highest score to use when rolling and for damage.

cdverrett94:
Schema includes the characteristics that can be used for the ability. In _prepareCharacterData, the highest characteristic is chosen. Future work includes:

- Add checkbox to each tiers damage for adding characteristic to damage
- Add characteristic to damage if checked

Closes #92

### DIFF
--- a/src/module/data/item/ability.mjs
+++ b/src/module/data/item/ability.mjs
@@ -207,7 +207,7 @@ export default class AbilityModel extends BaseItemModel {
 
     context.showDamageDisplay = this.keywords.has("melee") && this.keywords.has("ranged");
 
-    context.damageType = Object.entries(ds.CONFIG.damageTypes).map(([value, {label}]) => ({value, label}));
+    context.damageTypes = Object.entries(ds.CONFIG.damageTypes).map(([value, {label}]) => ({value, label}));
     context.appliedEffects = this.parent.effects.filter(e => !e.transfer).map(e => ({label: e.name, value: e.id}));
 
     context.characteristics = Object.entries(ds.CONFIG.characteristics).map(([value], {label}) => ({value, label}));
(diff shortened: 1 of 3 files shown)


JPMeehan:
I think moving from using `choices` to providing the options as an argument to the form, as is done with `ClassModel` and `EquipmentModel`, is more conducive to people potentially modifying `characteristics`. Of course, it may be more conducive to just say "screw it" and move all that info into CONST and then hard lock choices.

src/module/data/item/ability.mjs
```diff
@@ -60,6 +60,7 @@ export default class AbilityModel extends BaseItemModel {
 
     schema.powerRoll = new fields.SchemaField({
       enabled: new fields.BooleanField(),
+      characteristics: new fields.SetField(new fields.StringField({choices: ds.CONFIG.characteristics})),
```
JPMeehan, line 63: Switch to `{required: true, blank: false}`

templates/item/partials/ability.hbs
```diff
@@ -32,6 +32,7 @@
 </fieldset>
 {{formGroup systemFields.powerRoll.fields.enabled value=system.powerRoll.enabled}}
 {{#if system.powerRoll.enabled}}
+{{formGroup systemFields.powerRoll.fields.characteristics value=system.powerRoll.characteristics}}
```
JPMeehan, line 35: Add `context.characteristics = Object.entries(ds.CONFIG.characteristics).map(([value], {label}) => ({value, label}));` to `getSheetContext` and then add `options=characteristics` in this

JPMeehan (changes requested):

cdverrett94:
I've made the changes you've requested. Good to know on options=characteristics and the rationale.

JPMeehan (approved):

PR #98, opened 2024-12-27
Add characteristic to damage

### LINKED ISSUE #97: Add Characteristic to ability damage
Follow up to #92 to use the selected characteristic in the damage formula

cdverrett94:
Closes #97 

Let me know if you have a better name for the property or the label.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -271,6 +271,9 @@
             "characteristics": {
               "label": "Characteristics"
             },
+            "characteristicDamage": {
+              "label": "Characteristic Damage"
+            },
             "tier1": {
               "label": "Tier 1 Result",
               "damage": {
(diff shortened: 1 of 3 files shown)


JPMeehan:
So, from a philosophical perspective, shouldn't we just have something like `@mod` available in the roll data that people can use to dynamically reference the characteristic?

cdverrett94:
So instead of adding the characteristic to the damage formula in data prep, are you suggesting that by adding @mod to the item's roll data, you could just input the damage formula as like "3 + @mod"?

JPMeehan:
> So instead of adding the characteristic to the damage formula in data prep, are you suggesting that by adding `@mod` to the item's roll data, you could just input the damage formula as like "3 + `@mod`"?

Yeah exactly, makes things more clear. I'd probably name it `@chr` upon consideration, but same basic idea. 

I was going to PR the use ability stuff soon, need to refactor the roll data generally so stuff like the subtype model has a chance to make adjustments beyond simply being included as a shallow copy

cdverrett94:
That's probably fair and less code to maintain. If you're already going to be refactoring the roll data stuff soon, then probably not beneficial for me to add this to the roll data right now. I can just close this one instead.

PR #99, opened 2024-12-27
Add triggered and free triggered/maneuver actions

### LINKED ISSUE #62: Expand Ability Types
> Most abilities require you to
use an action, a maneuver, a triggered action, a free
maneuver, or a free triggered actio

cdverrett94:
Closes #62 

Now that there are triggered types, I've updated the sheet to only show the triggered field when it's a triggered or free triggered action. I copied the pattern of putting sheet conditionals into the sheet context.

### DIFF
--- a/src/module/data/item/ability.mjs
+++ b/src/module/data/item/ability.mjs
@@ -165,7 +165,7 @@ export default class AbilityModel extends BaseItemModel {
     context.keywords = Object.entries(config.keywords).map(([value, {label}]) => ({value, label}));
     context.actionTypes = Object.entries(config.types).map(([value, {label}]) => ({value, label}));
 
-    context.triggeredAction = ["triggered", "freeTriggered"].includes(this.type);
+    context.triggeredAction = Object.entries(config.types).filter(type => type.triggered).includes(this.type);
 
     context.distanceTypes = Object.entries(config.distances).map(([value, {label}]) => ({value, label}));
     context.primaryDistance = config.distances[this.distance.type].primary;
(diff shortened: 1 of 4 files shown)


src/module/data/item/ability.mjs
```diff
@@ -165,6 +165,8 @@ export default class AbilityModel extends BaseItemModel {
     context.keywords = Object.entries(config.keywords).map(([value, {label}]) => ({value, label}));
     context.actionTypes = Object.entries(config.types).map(([value, {label}]) => ({value, label}));
 
+    context.triggeredAction = ["triggered", "freeTriggered"].includes(this.type);
```
JPMeehan, line 168: I would rather have this as an optional Boolean on the config object, same level as the label, something like `triggered: true`.

JPMeehan (changes requested):
Slight change in implementation but otherwise good work here

src/module/data/item/ability.mjs
cdverrett94: I just made the changes. I think this is what you were asking. I assumed the issue is the hard coding of what "triggered" is being in the context instead of the config.

cdverrett94: Also, I think I did the typing right? Not terribly familiar with that.

JPMeehan (approved):
Also works, although the simplest is to just directly check the Boolean - `context.actionTypes = !!ds.CONFIG.actionTypes[this.type]?.triggered`, no need to check against an array now that it's a property in the object.

cdverrett94:
> Also works, although the simplest is to just directly check the Boolean - `context.actionTypes = !!ds.CONFIG.actionTypes[this.type]?.triggered`, no need to check against an array now that it's a property in the object.

That's actually better and simpler. LOL. Not sure why I didn't think to do that. Thanks!

PR #104, opened 2024-12-28
Hero Tokens API

JPMeehan:
This implements the API portion of Hero Token support. Still need to figure out the UI portion; in the meanwhile, the following basic commands interact with the tokens as a player and GM respectively

```js
heroTokens = game.settings.get("draw-steel", "heroTokens")
heroTokens.spendToken("regainStamina")
```

```js
game.settings.set("draw-steel", "heroTokens", {value: 5})
```

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -738,12 +738,12 @@
       }
     },
     "Setting": {
-      "NoActiveGM": "This requires an active GM logged into the world",
+      "NoActiveGM": "This requires an active GM logged into the world.",
       "HeroTokens": {
         "Label": "Hero Tokens",
-        "Hint": "Hero tokens are a group resource that is tracked by the players and kept in a pool that is accessible to all their characters",
+        "Hint": "Hero tokens are a group resource that is tracked by the players and kept in a pool that is accessible to all their characters.",
         "NoHeroTokens": "No Hero Tokens available",
-        "WarnDirectorBadSpend": "{name} attempted to spend a hero token while none were available",
+        "WarnDirectorBadSpend": "{name} attempted to spend a hero token while none were available.",
         "GainSurges": {
           "label": "Gain Surges",
           "messageContent": "Spends a hero token to gain two surges."
@@ -758,7 +758,7 @@
         },
         "RegainStamina": {
           "label": "Regain Stamina",
-          "messageContent": "Spends two hero tokens to regain stamina equal to their recovery value"
+          "messageContent": "Spends two hero tokens to regain stamina equal to their recovery value."
         }
       }
     },
--- a/lang/en.json
+++ b/lang/en.json
@@ -738,11 +738,28 @@
       }
     },
     "Setting": {
+      "NoActiveGM": "This requires an active GM logged into the world",
       "HeroTokens": {
         "Label": "Hero Tokens",
         "Hint": "Hero tokens are a group resource that is tracked by the players and kept in a pool that is accessible to all their characters",
         "NoHeroTokens": "No Hero Tokens available",
-        "WarnDirectorBadSpend": "{name} attempted to spend a hero token while none were available"
+        "WarnDirectorBadSpend": "{name} attempted to spend a hero token while none were available",
+        "GainSurges": {
+          "label": "Gain Surges",
+          "messageContent": "Spends a hero token to gain two surges."
+        },
+        "SucceedSave": {
+          "label": "Succeed on a Saving Throw",
+          "messageContent": "Spends a hero token to succeed on a saving throw instead of failing."
+        },
+        "ImproveTest": {
+          "label": "Improve Test Result",
+          "messageContent": "Spends a hero token to succeed on a test and lose any consequences suffered."
+        },
+        "RegainStamina": {
+          "label": "Regain Stamina",
+          "messageContent": "Spends two hero tokens to regain stamina equal to their recovery value"
+        }
       }
     },
     "Skill": {
--- a/lang/en.json
+++ b/lang/en.json
@@ -737,6 +737,12 @@
         "Label": "Saving Throw"
       }
     },
+    "Setting": {
+      "HeroTokens": {
+        "Label": "Hero Tokens",
+        "Hint": "Hero tokens are a group resource that is tracked by the players and kept in a pool that is accessible to all their characters"
+      }
+    },
     "Skill": {
       "Group": {
         "Crafting": "Crafting",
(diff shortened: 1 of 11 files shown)


src/module/data/settings/hero-tokens.mjs
```diff
@@ -0,0 +1,45 @@
...
+   * @param {string} spendType - Key of ds.CONFIG.hero.tokenSpends
+   * @returns {boolean} Returns an explicit false if the socket message wasn't sent
+   */
+  spendToken(spendType) {
+    if (!game.users.activeGM) {
+      ui.notifications.error("DRAW_STEEL.Setting.NoActiveGM", {localize: true});
+      return false;
+    }
+    const tokenSpend = ds.CONFIG.hero.tokenSpends[spendType];
+    if (!tokenSpend) {
```
krbz999, line 34: I think it's fine to allow `null` and `0`.
```suggestion
    if (tokenSpend === undefined) {
```

src/module/helpers/sockets.mjs
```diff
@@ -28,4 +33,36 @@ export default class DrawSteelSocketHandler {
...
     game.socket.emit(this.identifier, {type, payload});
   }
+
+  /**
+   * Tell the GM to spend a hero token
+   * @param {object} payload
+   * @param {string} payload.userId
+   * @param {string} payload.spendType
+   */
+  async spendHeroToken({userId, spendType}) {
```
krbz999, line 43: Do we expect to add a lot more socket events? I can imagine we will, for example for forced movement of an enemy. In which case, what did you think of my PR for a socket manager for the dnd5e system?

src/module/data/settings/hero-tokens.mjs
JPMeehan: This is an object, but I'll rename the variable to something like tokenConfiguration

src/module/helpers/sockets.mjs
JPMeehan: Yes absolutely expect more socket events to get around the permission structure.

Oh, interesting work on enforcing certain kinds of data in the payloads by rebinding the entire structure. I also like tying together the emission & handling.

lang/en.json
```diff
@@ -737,6 +737,31 @@
         "Label": "Saving Throw"
       }
     },
+    "Setting": {
+      "NoActiveGM": "This requires an active GM logged into the world",
+      "HeroTokens": {
+        "Label": "Hero Tokens",
+        "Hint": "Hero tokens are a group resource that is tracked by the players and kept in a pool that is accessible to all their characters",
+        "NoHeroTokens": "No Hero Tokens available",
+        "WarnDirectorBadSpend": "{name} attempted to spend a hero token while none were available",
```
krbz999, line 746:
```suggestion
        "WarnDirectorBadSpend": "{name} attempted to spend a hero token while none were available.",
```

lang/en.json
```diff
@@ -737,6 +737,31 @@
         "Label": "Saving Throw"
       }
     },
+    "Setting": {
+      "NoActiveGM": "This requires an active GM logged into the world",
+      "HeroTokens": {
+        "Label": "Hero Tokens",
+        "Hint": "Hero tokens are a group resource that is tracked by the players and kept in a pool that is accessible to all their characters",
```
krbz999, line 744:
```suggestion
        "Hint": "Hero tokens are a group resource that is tracked by the players and kept in a pool that is accessible to all their characters.",
```

lang/en.json
```diff
@@ -737,6 +737,31 @@
         "Label": "Saving Throw"
       }
     },
+    "Setting": {
+      "NoActiveGM": "This requires an active GM logged into the world",
```
krbz999, line 741:
```suggestion
      "NoActiveGM": "This requires an active GM logged into the world.",
```

lang/en.json
```diff
@@ -737,6 +737,31 @@
...
+          "label": "Succeed on a Saving Throw",
+          "messageContent": "Spends a hero token to succeed on a saving throw instead of failing."
+        },
+        "ImproveTest": {
+          "label": "Improve Test Result",
+          "messageContent": "Spends a hero token to succeed on a test and lose any consequences suffered."
+        },
+        "RegainStamina": {
+          "label": "Regain Stamina",
+          "messageContent": "Spends two hero tokens to regain stamina equal to their recovery value"
```
krbz999, line 761:
```suggestion
          "messageContent": "Spends two hero tokens to regain stamina equal to their recovery value."
```

src/module/config.mjs
```diff
@@ -643,8 +643,39 @@ preLocalize("negotation.motivations", {key: "label"});
  * Configuration information for heros
  */
 DRAW_STEEL.hero = {
-  xp_track: [0, 16, 32, 48, 64, 80, 96, 112, 128, 144]
+  /**
+   * XP progression for heroes
+   * @type {number[]}
+   */
+  xp_track: [0, 16, 32, 48, 64, 80, 96, 112, 128, 144],
```
krbz999, line 650: I have a general dislike for arrays of "generic" data such as this but I think in this case it's likely the only, or best, way of recording it.

src/module/data/settings/hero-tokens.mjs
```diff
@@ -0,0 +1,45 @@
...
+const fields = foundry.data.fields;
+
+/**
+ * A data model to manage Hero Tokens in Draw Steel
+ */
+export class HeroTokenModel extends foundry.abstract.DataModel {
+  /** @override */
+  static defineSchema() {
+    return {
+      value: new fields.NumberField({required: true, nullable: false, initial: 0, integer: true, min: 0})
```
krbz999, line 12: I don't believe `required: true` is needed if we have `initial: 0`.

src/module/data/settings/hero-tokens.mjs
```diff
@@ -0,0 +1,45 @@
...
+  static label = "DRAW_STEEL.Setting.HeroTokens.Label";
+
+  /** Helper text for Hero Tokens */
+  static hint = "DRAW_STEEL.Setting.HeroTokens.Hint";
+
+  /**
+   * Send a socket message to the Director to spend a hero token
+   * Necessary because only game masters can modify world settings
+   * @param {string} spendType - Key of ds.CONFIG.hero.tokenSpends
+   * @returns {boolean} Returns an explicit false if the socket message wasn't sent
```
krbz999, line 26:
```suggestion
   * Send a socket message to the Director to spend a hero token
   * Necessary because only game masters can modify world settings
   * @param {string} spendType       Key of `ds.CONFIG.hero.tokenSpends`.
   * @returns {void|false}           An explicit `false` if the socket message was not sent.
```

krbz999 (approved):

PR #108, opened 2024-12-28
Add echelon to actor data and roll data

### LINKED ISSUE #106: Include character echelon as getter on BaseActorModel and in roll data
The derived value of a character's echelon - e.g. 1 for a level 1-3 character - should be available both in the system properties as well as roll data. This should be available for both characters and npcs.

cdverrett94:
Closes #106 

- I had to add level to `NPCModel`. 
- Added a level getter so that I could use that instead of actor type checking in the echelon getter. 
- Added echelon to roll data

### DIFF
--- a/src/module/data/actor/base.mjs
+++ b/src/module/data/actor/base.mjs
@@ -109,7 +109,7 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
   get echelon() {
     return Object.entries(ds.CONFIG.echelons).reduce((acc, current) => {
       return this.level >= current[1].threshold? Number(current[0]) : acc;
-    }, 0);
+    }, 1);
   }
 
   /**
--- a/src/module/data/actor/base.mjs
+++ b/src/module/data/actor/base.mjs
@@ -97,6 +97,22 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
   }
 
   /**
+   * The actor's level
+   */
+  get level() {
+    return 1;
+  }
+  
+  /**
+   * The actor's echelon based on their current level
+   */
+  get echelon() {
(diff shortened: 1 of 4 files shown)


src/module/data/actor/base.mjs
```diff
@@ -72,6 +72,22 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
...
+   */
+  get level() {
+    return 1;
+  }
+  
+  /**
+   * The actor's echelon based on their current level
+   */
+  get echelon() {
+    return Object.entries(ds.CONFIG.echelons).reduce((acc, current) => {
```
JPMeehan, line 86: You can destructure `current` to something like `entry[key, value]` so you can just have `value.threshold` and `Number(key)`

src/module/data/actor/base.mjs
```diff
@@ -72,6 +72,22 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
...
+      return this.level >= current[1].threshold? Number(current[0]) : acc;
```
JPMeehan, line 87:
```suggestion
      return this.level >= current[1].threshold ? Number(current[0]) : acc;
```

JPMeehan (approved):
I waffled a bit on whether the echelon stuff should go in `modifyRollData` - merge/rebase from main if you don't see it, I added it in the Potency PR that just got merged - but I think it's a sufficiently universal concept that every actor subtype should probably provide it.

src/module/data/actor/base.mjs
cdverrett94: Thanks. Didn't think to destructure there

cdverrett94:
Updated per your comments. Moved the echelon roll data to `modifyRollData`.

PR #109, opened 2024-12-28
Apply kit stamina bonus per echelon

### LINKED ISSUE #107: Update kit stamina bonus to be per echelon
Relies on #106

cdverrett94:
Closes #107


JPMeehan (approved):
Had to double check there wasn't any risk of NaN but looks like the previous implementation ensures we're good

PR #110, opened 2024-12-28
Adjust character prototype token on _preCreate

### LINKED ISSUE #90: Adjust Character prototype token data in _preCreate
1. Linked
2. Friendly Disposition
3. Has Vision

cdverrett94:
Closes #90

### DIFF
--- a/src/module/data/actor/character.mjs
+++ b/src/module/data/actor/character.mjs
@@ -136,6 +136,20 @@ export default class CharacterModel extends BaseActorModel {
     this.potency.strong += highestCharacteristic;
   }
 
+  async _preCreate(data, options, user) {
+    await super._preCreate(data, options, user);
+
+    this.parent.updateSource({
+      prototypeToken: {
+        actorLink: true,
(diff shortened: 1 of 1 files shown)


src/module/data/actor/character.mjs
```diff
@@ -136,6 +136,21 @@ export default class CharacterModel extends BaseActorModel {
     this.potency.strong += highestCharacteristic;
   }
 
+  /** @override */
+  async _preCreate(data, options, user) {
+    await super._preCreate(data, options, user);
```
JPMeehan, line 141: Although we don't currently ever return false in BaseActorModel, it's still good practice to include
```suggestion
    const allowed = await super._preCreate(data, options, user);
    if (allowed === false) return false;
```

JPMeehan (approved):

src/module/data/actor/character.mjs
cdverrett94: Fair enough. I'll keep that in mind for the future.

PR #113, opened 2024-12-28
Update keywords

### LINKED ISSUE #61: Update Keywords
Remove resistance, add strike

cdverrett94:
- Removed Attack and added Strike keywords.
- Added other keywords from class abilities
- I used a similar method to add the optgroups as the skills config. I just added sorting to it. Sorting is by no group first, then by group, then by key

With sorting
dropdown with sort

Without sorting
dropdown without sort

Closes #61

### DIFF
--- a/templates/item/partials/ability.hbs
+++ b/templates/item/partials/ability.hbs
@@ -9,7 +9,7 @@
   {{!-- TODO: Forced Movement --}}
 </fieldset>
 {{/inline}}
-{{formGroup systemFields.keywords value=system.keywords options=keywords}}
+{{formGroup systemFields.keywords value=system.keywords options=config.abilities.keywords.optgroups}}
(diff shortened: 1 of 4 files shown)


templates/item/partials/ability.hbs
```diff
@@ -9,7 +9,7 @@
   {{!-- TODO: Forced Movement --}}
 </fieldset>
 {{/inline}}
-{{formGroup systemFields.keywords value=system.keywords options=keywords}}
```
JPMeehan, line 12: Can also remove this part of preparePartContext in AbilityData

JPMeehan (approved):

templates/item/partials/ability.hbs
cdverrett94: Pushed. I assume that's what you meant.

PR #115, opened 2024-12-30
Add ability categories

### LINKED ISSUE #114: Add different ability categories such as Heroic, Signature, Villain Action
Currently no way to indicate that an ability is a heroic or signature ability. Villain actions could be included in this as well. Would denoting free strikes be here as well?

cdverrett94:
Added categories to ability config and add category to AbilityModel. I updated the type for AbilityModel to include this new field. I think I did this correctly as I'm not terribly familiar with TS, so please correct me on the best way.

Current categories are: 

- Heroic Ability
- Signature Ability
- Villain Action
- Free Strike

Closes #114

### DIFF
--- a/src/module/data/item/_types.d.ts
+++ b/src/module/data/item/_types.d.ts
@@ -50,6 +50,7 @@ declare module "./ability.mjs" {
     }
     keywords: Set<string>;
     type: keyof typeof ds["CONFIG"]["abilities"]["types"];
+    category: keyof typeof ds["CONFIG"]["abilities"]["categories"] | "";
     damageDisplay: "melee" | "ranged";
     distance: {
       type: keyof typeof ds["CONFIG"]["abilities"]["distances"];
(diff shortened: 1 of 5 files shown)


src/module/data/item/ability.mjs
```diff
@@ -29,6 +29,7 @@ export default class AbilityModel extends BaseItemModel {
 
     schema.keywords = new fields.SetField(new fields.StringField({required: true, blank: false}));
     schema.type = new fields.StringField({required: true, blank: false, initial: "action"});
+    schema.category = new fields.StringField({required: true, blank: true, initial: ""}),
```
JPMeehan, line 32:
```suggestion
    schema.category = new fields.StringField({required: true, nullable: false}),
```

Blank is true by default, and StringField has some smart initial handling if you throw in the nullable: false

src/module/data/item/_types.d.ts
```diff
@@ -50,6 +50,7 @@ declare module "./ability.mjs" {
     }
     keywords: Set<string>;
     type: keyof typeof ds["CONFIG"]["abilities"]["types"];
+    category: keyof typeof ds["CONFIG"]["abilities"]["categories"] | "";
```
JPMeehan, line 53: The way to check your work is try to access the property elsewhere, but this looks correct. Basically what these types file do is make sure our data models have the instance properties proscribed by the static-side schemas.

JPMeehan (approved):
Overall looks good, just one tweak

PR #116, opened 2024-12-31
Implement API to Use an ability

### LINKED ISSUE #76: Implement "Use Ability"
There should be a clean function sheets and macros can call to "Use" an ability. This may or may not involve making a Power Roll but should always create a chat message.

This may also be a point where we define a base chat message model to store any additional data.

### LINKED ISSUE #97: Add Characteristic to ability damage

JPMeehan:
- More work can be done both on the data model and the embedded HTML
- Also need to implement UI for using abilities from the character sheet
- File count is slightly bloated from making some other type fixes along the way

Closes #76 
Closes #97

### DIFF
--- a/draw-steel.mjs
+++ b/draw-steel.mjs
@@ -30,7 +30,7 @@ Hooks.once("init", function () {
     CONFIG[docCls.documentName].documentClass = docCls;
   }
 
-  const templates = [];
+  const templates = ["templates/item/embeds/ability.hbs"].map(t => DS_CONST.systemPath(t));
 
   // Assign data models & setup templates
   for (const [doc, models] of Object.entries(data)) {
@@ -96,7 +96,6 @@ Hooks.once("i18nInit", () => {
 Hooks.once("ready", async function () {
   // Wait to register hotbar drop hook on ready so that modules could register earlier if they want to
   // Hooks.on("hotbarDrop", (bar, data, slot) => helpers.macros.createDocMacro(data, slot));
-  await applications.helpers.registerDrawSteelPartials();
   Hooks.callAll("ds.ready");
   console.log(DS_CONST.ASCII);
 });
--- a/draw-steel.mjs
+++ b/draw-steel.mjs
@@ -93,7 +93,7 @@ Hooks.once("i18nInit", () => {
 /*  Ready Hook                                  */
 /* -------------------------------------------- */
 
-Hooks.once("ready", async function () {
+Hooks.once("ready", function () {
   // Wait to register hotbar drop hook on ready so that modules could register earlier if they want to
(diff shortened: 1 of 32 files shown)


draw-steel.mjs
```diff
@@ -93,7 +93,7 @@ Hooks.once("i18nInit", () => {
 /*  Ready Hook                                  */
 /* -------------------------------------------- */
 
-Hooks.once("ready", function () {
+Hooks.once("ready", async function () {
```
krbz999, line 96:
```suggestion
Hooks.once("ready", function () {
```

src/module/data/item/ability.mjs
```diff
@@ -182,4 +209,53 @@ export default class AbilityModel extends BaseItemModel {
...
+
+    DrawSteelChatMessage.applyRollMode(messageData, "roll");
+
+    if (this.powerRoll.enabled) {
+      const formula = this.powerRoll.formula ? `2d10 + ${this.powerRoll.formula}` : "2d10";
+      const rollData = this.parent.getRollData();
+      const rollOptions = {
+        type: "ability"
+      }; // TODO: Add in Banes & Edges
+      const powerRoll = new PowerRoll(formula, rollData, rollOptions);
```
krbz999, line 245: We should consider putting most of this API in the roll class where it can be given parameters and be separate from items and actors. That's for the future though.

src/module/documents/actor.mjs
```diff
@@ -2,7 +2,7 @@ export class DrawSteelActor extends Actor {
   /** @override */
   getRollData() {
     // Shallow copy
-    const rollData = {...this.system};
+    const rollData = {...this.system, flags: this.flags};
```
krbz999, line 5: I will suggest a dnd5e'ism here: Add `name` and `statuses` as well.

src/module/documents/item.mjs
```diff
@@ -4,8 +4,14 @@ export class DrawSteelItem extends Item {
   getRollData() {
     const rollData = this.actor?.getRollData() ?? {};
 
+    rollData.itemFlags = this.flags;
+
     // Shallow copy
-    rollData.item = {...this.system};
+    rollData.item = this.system;
```
krbz999, line 10:
```suggestion
    rollData.item = {...this.system, flags: this.flags, name: this.name};
```

src/module/documents/item.mjs
```diff
@@ -4,8 +4,14 @@ export class DrawSteelItem extends Item {
   getRollData() {
     const rollData = this.actor?.getRollData() ?? {};
 
+    rollData.itemFlags = this.flags;
```
krbz999, line 7:
```suggestion
```

krbz999 (approved):

PR #126, opened 2024-12-31
Add takeDamage method to BaseActorModel

### LINKED ISSUE #124: Add a takeDamage method to BaseActorModel
Should account for immunities/weaknesses as well.

I'm thinking the method should take the following parameters: 

- damage amount - required
- damage type - optional - if none then untyped damage
- ignore immunity - optional - should immunities apply. useful for the bleeding condition damage that can't be prevented in any way. some monsters look like they can ignore certain immunities as well

### LINKED ISSUE #125: Update immunities/weakness to no longer include keywords
In starting #124, I was reading the Damage Immunity section of the backer packet, and it no longer includes the parts that mention immunity to Weapon, Magic, or Psionic.

cdverrett94:
I also removed the keyword stuff from the damage weaknesses/immunities. I wasn't sure what the typing should be for `option.ignoredImmunities` to have it be either a damage type or all, so I just used string.

Closes #124
Closes #125

### DIFF
--- a/src/module/data/actor/base.mjs
+++ b/src/module/data/actor/base.mjs
@@ -187,7 +187,10 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
 
     damage = Math.max(0, damage + weaknessAmount - immunityAmount);
 
-    if(damage === 0) return this.parent;
+    if(damage === 0) {
+      ui.notifications.info("DRAW_STEEL.Actor.DamageNotification.ImmunityReducedToZero", {localize: true});
+      return this.parent;
+    }
 
     // If there's damage left after weakness/immunities, apply damage to temporary stamina then stamina value
     const staminaUpdates = {};
--- a/src/module/data/actor/base.mjs
+++ b/src/module/data/actor/base.mjs
@@ -188,6 +188,7 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
     damage = Math.max(0, damage + weaknessAmount - immunityAmount);
 
     if(damage === 0) {
+      // TODO: V13 allows the format option to be passed. Notification could be updated to include the damaged actor's name
       ui.notifications.info("DRAW_STEEL.Actor.DamageNotification.ImmunityReducedToZero", {localize: true});
       return this.parent;
     }
--- a/src/module/data/actor/base.mjs
+++ b/src/module/data/actor/base.mjs
@@ -42,8 +42,8 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
     });
 
     schema.damage = new fields.SchemaField({
-      immunities: damageTypes(requiredInteger, {all: true, keywords: true}),
-      weaknesses: damageTypes(requiredInteger, {all: true, keywords: true})
+      immunities: damageTypes(requiredInteger, {all: true}),
+      weaknesses: damageTypes(requiredInteger, {all: true})
     });
 
     return schema;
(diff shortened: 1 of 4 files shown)


src/module/data/actor/base.mjs
```diff
@@ -162,4 +162,41 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
...
+    // Reduce the immunities list to non-ignored immunities
+    const immunities = Object.entries(this.damage.immunities).reduce((acc, [type, amount]) => {
+      if (!options.ignoredImmunities.includes("all") && !options.ignoredImmunities.includes(type)) acc[type] = amount;
+      return acc;
+    }, {});
+    const immunityAmount = Math.max(immunities.all ?? 0, immunities[options.type] ?? 0); // Null check in case type is not in immunities
+
+    damage = Math.max(0, damage + weaknessAmount - immunityAmount);
+
+    if(damage === 0) return this.parent;
```
JPMeehan, line 190: Feels like we should generate some kind of message or something if it no-sells the damage? Thinking `ui.notifications.info("", {localize: true})` rather than a console thing.

src/module/data/actor/base.mjs
```diff
@@ -162,4 +162,41 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
...
+
+    // If there's damage left after weakness/immunities, apply damage to temporary stamina then stamina value
+    const staminaUpdates = {};
+    const damageToTempStamina = Math.min(damage, this.stamina.temporary);
+    staminaUpdates.temporary = Math.max(0, this.stamina.temporary - damageToTempStamina);
+
+    const remainingDamage = Math.max(0, damage - damageToTempStamina);
+    if(remainingDamage > 0) staminaUpdates.value = this.stamina.value - remainingDamage;
```
JPMeehan, line 198: Should we try to account for applying winded & dying here?

JPMeehan (changes requested):
Overall looks good - I think we also need to eventually add in hooks but I don't want to get ahead of ourselves in making public API commitments

src/module/data/actor/base.mjs
cdverrett94: Would that be better to go into _preUpdate or prepareDerivedData to accountsfor stamina changes in other places like just a direct actor update or changes from the actor sheet?

src/module/data/actor/base.mjs
cdverrett94: Will do.

src/module/data/actor/base.mjs
JPMeehan: Hmm, that's fair, and also the token bar is a frequent source of stamina adjustments. Might need to look into how dnd5e is handling bloodied.

src/module/data/actor/base.mjs
```diff
@@ -162,4 +162,41 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
...
+
+    return await this.parent.update({"system.stamina": staminaUpdates});
```
JPMeehan, line 200: Oh also don't need to throw in the await here, the promise will automatically be handled by any `await` usage by whomever calls this function

src/module/data/actor/base.mjs
cdverrett94: Looks like they have it placed in _onUpdate

src/module/data/actor/base.mjs
cdverrett94: I've added the notification. Let me know if the wording is good or the location in en.json is acceptable. Wasn't sure where to put it.

src/module/data/actor/base.mjs
cdverrett94: Removed the await

JPMeehan (approved):

PR #132, opened 2025-01-03
Automate resource gain on start combat

cdverrett94:
I didn't do the start of turn automation because I think that will require more work to determine when the actor's turn starts and adding a field for resource gain.


JPMeehan (approved):

JPMeehan:
We may end up refactoring to a combat subtype when we add negotiations but that can be a future problem

PR #133, opened 2025-01-03
Add winded/dying stamina based effects

### LINKED ISSUE #127: Implement Winded & Dying
Both should have similar functionality in terms of checking current against max HP. Might just involve some actor.statuses trickery in _onUpdate?

cdverrett94:
Once more of the rules journal is filled out, should probably add a rules entry to these similar to condition.

Closes #127

### DIFF
--- a/src/module/data/actor/base.mjs
+++ b/src/module/data/actor/base.mjs
@@ -154,7 +154,7 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
       let threshold = (Number.isNumeric(value.threshold)) ? value.threshold : foundry.utils.getProperty(this.parent, value.threshold);
       threshold = Number(threshold);
 
-      const active = (!Number.isNumeric(threshold) || (this.stamina.value > threshold)) ? false : true;
+      const active = Number.isNumeric(threshold) && (this.stamina.value <= threshold);
       this.parent.toggleStatusEffect(key, {active});
     });
   }
--- a/src/module/data/actor/base.mjs
+++ b/src/module/data/actor/base.mjs
@@ -104,12 +104,12 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
   get level() {
     return 1;
   }
-  
+
   /**
    * The actor's echelon based on their current level
    */
   get echelon() {
-    return Object.entries(ds.CONFIG.echelons).reduce((acc, [key,value]) => {
+    return Object.entries(ds.CONFIG.echelons).reduce((acc, [key, value]) => {
       return this.level >= value.threshold ? Number(key) : acc;
     }, 1);
   }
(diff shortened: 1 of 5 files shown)


src/module/data/actor/base.mjs
```diff
@@ -132,6 +132,33 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
...
+
+  /**
+   * Update the stamina effects based on updated stamina values
+   */
+  updateStaminaEffects() {
+    Object.entries(ds.CONST.staminaEffects).forEach(([key, value]) => {
+      let threshold = (Number.isNumeric(value.threshold)) ? value.threshold : foundry.utils.getProperty(this.parent, value.threshold);
+      threshold = Number(threshold);
+
+      const active = (!Number.isNumeric(threshold) || (this.stamina.value > threshold)) ? false : true;
```
JPMeehan, line 157: feels like writing the contrapositive is much more clear here
```suggestion
      const active = Number.isNumeric(threshold) && (this.stamina.value <= threshold);
```

src/module/data/actor/base.mjs
```diff
@@ -132,6 +132,33 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
...
+    super._onUpdate(changed, options, userId);
+
+    if (changed.system?.stamina) this.updateStaminaEffects();
+  }
+
+  /**
+   * Update the stamina effects based on updated stamina values
+   */
+  updateStaminaEffects() {
+    Object.entries(ds.CONST.staminaEffects).forEach(([key, value]) => {
```
JPMeehan, line 153: `for of` loops properly handle asynchronicity unlike forEach
```suggestion
    for (const [key, value] of Object.entries(ds.CONST.staminaEffects)) {
```

src/module/data/actor/base.mjs
```diff
@@ -132,6 +132,33 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
...
+      let threshold = (Number.isNumeric(value.threshold)) ? value.threshold : foundry.utils.getProperty(this.parent, value.threshold);
+      threshold = Number(threshold);
+
+      const active = (!Number.isNumeric(threshold) || (this.stamina.value > threshold)) ? false : true;
+      this.parent.toggleStatusEffect(key, {active});
```
JPMeehan, line 158: This function is asynchronous and with the change above will work properly with the await
```suggestion
      await this.parent.toggleStatusEffect(key, {active});
```

JPMeehan (changes requested):
Some adjustments

src/module/data/actor/base.mjs
cdverrett94: The _onUpdate method isn't asynchronous. So we can't await the return of updateStaminaEffects. That's the reason I didn't make updateStaminaEffects async and didn't add the await here or use a for loop.

cdverrett94: I've gone ahead and updated it to use async/await.

JPMeehan (approved):

src/module/data/actor/base.mjs
JPMeehan: Yeah offloading to a helper function helps here. Also makes it possible to subclass for Character & NPC if we decide to add Dead here later.

PR #134, opened 2025-01-03
Add turns to actor and combatant models

cdverrett94:
Added turns to the actor and combatant models. Actor turns is a static number for the combatant to pull from at combat start and next round.

Previously if you set an initiative of 5, on a new round, it would just update to 1. Now this allows it reset to the proper number. Useful for solo monsters.

Added a startCombat method to the actor models to run actor specific stuff like adding heroic resources for characters

If this isn't how you think this should be handled, let me know if you have changes or just want to decline it.

EDIT: I just remembered that players can't update system properties of Combatant. I remembered I saw your issue on the main foundry repo. I've updated the combat tracker to use initiative again instead of the system property. But I still think updating it to the actors turn amount on combat start/new round/reset is helpful. Once v13 hits, can update to using combatant.system.turns

### DIFF
--- a/src/module/documents/combat.mjs
+++ b/src/module/documents/combat.mjs
@@ -10,9 +10,8 @@ export class DrawSteelCombat extends Combat {
 
   /** @override */
   async startCombat() {
-    const characters = this.combatants.filter(combatant => combatant.actor.type === "character");
-    for (const character of characters) {
-      await character.actor.update({"system.hero.primary.value": character.actor.system.hero.victories});
+    for (const combatant of this.combatants) {
+      await combatant.actor.system.startCombat(combatant);
     }
 
     return super.startCombat();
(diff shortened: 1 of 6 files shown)


src/module/documents/combat.mjs
```diff
@@ -10,9 +10,8 @@ export class DrawSteelCombat extends Combat {
 
   /** @override */
   async startCombat() {
-    const characters = this.combatants.filter(combatant => combatant.actor.type === "character");
-    for (const character of characters) {
-      await character.actor.update({"system.hero.primary.value": character.actor.system.hero.victories});
+    for (const combatant of this.combatants) {
+      await combatant.actor.system.startCombat(combatant);
```
JPMeehan, line 14: Just to help with errors should throw in optional access on `actor` since it can be null
```suggestion
      await combatant.actor?.system.startCombat(combatant);
```

src/module/documents/combat.mjs
```diff
@@ -22,7 +21,7 @@ export class DrawSteelCombat extends Combat {
   async nextRound() {
     await super.nextRound();
     if (game.settings.get(systemID, "initiativeMode") !== "default") return;
-    const combatantUpdates = this.combatants.filter(c => !c.initiative).map(c => ({_id: c.id, initiative: 1}));
+    const combatantUpdates = this.combatants.map(c => ({_id: c.id, initiative: c.actor.system.combat.turns}));
```
JPMeehan, line 24: Same deal
```suggestion
    const combatantUpdates = this.combatants.map(c => ({_id: c.id, initiative: c.actor?.system.combat.turns ?? 1}));
```

JPMeehan (approved):
Overall I think I like this approach, good use of the document subtypes

PR #136, opened 2025-01-04
Add take respite method

### LINKED ISSUE #135: Add take respite method
Add a method to the CharacterModel for takeRespite. This will reset stamina/recoveries and convert victories to xp.

cdverrett94:
Resets recoveries and stamina values to their max.
Add victories to XP and set victories to 0.

I didn't add this to the character sheet because I'm not sure how best to handle type specific actions with the actor sheets handling all types.

Closes #135

### DIFF
--- a/src/module/data/actor/character.mjs
+++ b/src/module/data/actor/character.mjs
@@ -158,6 +158,27 @@ export default class CharacterModel extends BaseActorModel {
     await this.parent.update({"system.hero.primary.value": this.hero.victories});
   }
 
+  /** 
+   * Take a respite resetting the character's stamina/recoveries and convert victories to XP
+   * @returns {Promise<Actor>}
+   */
+  async takeRespite() {
+    return this.parent.update({
+      system: {
+        hero: {
+          recoveries: {
+            value: this.recoveries.max
+          },
+          victories: 0,
+          xp: this.hero.xp + this.hero.victories
+        },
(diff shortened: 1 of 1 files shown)


JPMeehan:
Yeah really not sure on the UI aspect to all of this, we will probably end up splitting the actor sheets

src/module/data/actor/character.mjs
```diff
@@ -158,6 +158,27 @@ export default class CharacterModel extends BaseActorModel {
     await this.parent.update({"system.hero.primary.value": this.hero.victories});
   }
 
+  /** 
+   * Take a respite resetting the character's stamina/recoveries and convert victories to XP
+   * @returns {Promise<Actor>}
```
JPMeehan, line 163: Not sure if this is imported at the top of the file but should return the configured actor class.

```suggestion
   * @returns {Promise<DrawSteelActor>}
```

JPMeehan (approved):

src/module/data/actor/character.mjs
cdverrett94: Pushed. Had to import it as well.

cdverrett94:
> Yeah really not sure on the UI aspect to all of this, we will probably end up splitting the actor sheets

That's what I was thinking would need to be done, but wasn't sure if you were 100% committed to a single sheet.

PR #140, opened 2025-01-05
Implemented helper function for filling in options for SetField instances

### LINKED ISSUE #123: Ensure all SetField(StringField) instances have appropriate options
By default, StringField uses
- required: false
- nullable: true
- blank: true

We should
A. Make a helper object that can be used to ensure this consistency.
B. Implement it wherever appropriate

This is *not* necessary for any fields that implement `choices`, which should be used if and only if it's CONST-based

JPMeehan:
Main fix is ensuring nullable: false, but also some cases where I mistakenly added blank: true.

Closes #123

### DIFF
--- a/src/module/data/helpers.mjs
+++ b/src/module/data/helpers.mjs
@@ -21,6 +21,15 @@ export const barAttribute = (initial, min) => new SchemaField({
 export const requiredInteger = ({initial = 0, label} = {}) => new NumberField({initial, label, required: true, nullable: false, integer: true, min: 0});
 
 /**
+ * Constructs a string field for use inside of a SetField
+ * @param {object} [options] Options to forward to the field
+ * @param {Record<string, string>} [options.choices] CONST-derived choices for the field
+ * @param {Function} [options.validate] A validator function for field values
+ * @returns A string field that is always truthy
+ */
+export const setOptions = ({choices, validate} = {}) => new StringField({required: true, nullable: false, blank: false, choices, validate});
+
+/**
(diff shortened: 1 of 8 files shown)


src/module/data/helpers.mjs
```diff
@@ -20,6 +20,15 @@ export const barAttribute = (initial, min) => new SchemaField({
...
 export const requiredInteger = ({initial = 0, label} = {}) => new NumberField({initial, label, required: true, nullable: false, integer: true, min: 0});
 
+/**
+ * Constructs a string field for use inside of a SetField
+ * @param {object} [options] Options to forward to the field
+ * @param {Record<string, string>} [options.choices] CONST-derived choices for the field
+ * @param {Function} [options.validate] A validator function for field values
+ * @returns A string field that is always truthy
+ */
+export const setOptions = ({choices, validate} = {}) => new StringField({required: true, nullable: false, blank: false, choices, validate});
```
krbz999, line 30: This feels a little overkill. But I can see the use case for an extension of `SetField` instead if we want to reduce LoCs this way.

krbz999 (approved):
I think something like `class KeysField extends SetField` is worthwhile here. I don't feel strongly about it though.

src/module/data/helpers.mjs
JPMeehan: I think the normal SetField handling is everything we want for both form input handling as well as the AE handling, and this is ergonomic enough (Could also have the SetField creation in here as well, but eh)

PR #141, opened 2025-01-06
Initialized Migration Version, renamed Gamemaster to Director

### LINKED ISSUE #103: Rename Gamemaster and Assistant Gamemaster to Director
This is probably more than just some i18n stuff. Obviously the underlying permission checks will use the term Gamemaster, but the default name should be Director. It's possible we need some kind of "world setup" hook that runs and updates the name and tracks if it's done via a non-config setting.

JPMeehan:
Double duty setup

- Establish early a setting to track "fresh worlds"
- Use that to also rename Gamemaster to Director

Uncertain if overriding the core role translations are a good idea or if this should *just* be the base GM name as its own i18n string. 

Closes #103


src/module/data/migrations.mjs
```diff
@@ -0,0 +1,10 @@
+import {systemID} from "../constants.mjs";
+
+export async function migrateWorld() {
```
krbz999, line 3: This runs for non-GMs.

src/module/data/migrations.mjs
```diff
@@ -0,0 +1,10 @@
...
+  const migrationVersion = game.settings.get(systemID, "migrationVersion");
+  if (!migrationVersion) {
+    // New world - initialize the migration version and rename Gamemaster to Director
+    await game.users.activeGM.update({name: game.i18n.localize("USER.RoleGamemaster")});
```
krbz999, line 7: We disallow GMs to have a proper name?

krbz999 (changes requested):

src/module/data/migrations.mjs
JPMeehan: So, theory here is this will run once to change Gamemaster to Director, which is also why I'm attaching it to initializing the migrationVersion

src/module/data/migrations.mjs
```diff
@@ -0,0 +1,19 @@
+import {systemID} from "../constants.mjs";
+
+/**
+ * Perform one-time migrations
+ * Run and awaited in the `ready` hook before `ds.ready` is called.
+ */
+export async function migrateWorld() {
+  // TODO: In v13 simplify to the new getter
+  if (!game.users.activeGM.isSelf) {
```
krbz999, line 9:
```suggestion
  if (!game.users.activeGM?.isSelf) {
```

src/module/data/migrations.mjs
krbz999: Not sure I see the point of mingling those two concepts together. Rather, extend `User` to set the name.

krbz999 (changes requested):

src/module/data/migrations.mjs
krbz999: I think we should at least divorce the two concepts (migrations and GM user name). However, I don't feel strongly enough about the user name anyhow.

krbz999 (approved):

PR #142, opened 2025-01-06
NPC Specific Properties

JPMeehan:
- Updated keys & CONFIG info for various NPC properties
- Split the Actor Sheet bio page into Character & NPC

Follow-up work

- Add the Character demographic info (height, weight, age) to the now-split Character Bio page
- #66  
- Improve UI for the helper dialog


src/module/apps/actor-sheet/npc.mjs
```diff
@@ -28,8 +36,141 @@ export default class DrawSteelNPCSheet extends DrawSteelActorSheet {
...
+        break;
+    }
+    return context;
+  }
+
+  /**
+   * Fetches the printable string for the monster's keywords
+   * @returns {string}
+   */
+  _getMonsterKeywords() {
```
krbz999, line 65: Feels like this should inherit from a `_getKeywords` method on a base sheet.

src/module/apps/actor-sheet/npc.mjs
```diff
@@ -28,8 +36,141 @@ export default class DrawSteelNPCSheet extends DrawSteelActorSheet {
...
+      currentMotivations: formatter.format(currentMotivations),
+      currentPitfalls: formatter.format(currentPitfalls)
+    };
+  }
+
+  /**************
+   *
+   *   ACTIONS
+   *
+   **************/
```
krbz999, line 128: The funky syntax here is probably something we should align on. I'm a fan of just:
```js
/*   Actions                                          */
```
(50 hyphens, three spaces before the label)

src/module/apps/actor-sheet/npc.mjs
```diff
@@ -28,8 +36,141 @@ export default class DrawSteelNPCSheet extends DrawSteelActorSheet {
...
+      options: Object.entries(monsterConfig.roles).map(([value, {label}]) => ({value, label}))
+    });
+    const evInput = schema.getField("monster.ev").toFormGroup({}, {value: monsterData.ev});
+
+    htmlContainer.append(keywordInput, levelInput, organizationInput, roleInput, evInput);
+
+    /** @type {FormDataExtended | null} */
+    const fd = await foundry.applications.api.DialogV2.prompt({
+      content: htmlContainer.outerHTML,
+      window: {title: "DRAW_STEEL.Actor.NPC.MonsterMetadata.DialogTitle"},
```
krbz999, line 162: `icon: spaghetti-monster`?

krbz999 (approved):

src/module/apps/actor-sheet/npc.mjs
JPMeehan: The two sets of keywords in the system are for Monsters and Abilities, so I don't think there's helpful overlap

PR #145, opened 2025-01-08
Improved Item Description UX

### LINKED ISSUE #144: Improve display of Item Description tab
1. Only show enriched description in play mode
2. In edit mode, make the edit icon always visible
3. Make sure the heights are nicely handled
4. Turn whole frame into editor while editor is active

JPMeehan:
Activating an editor in the Item Description hides the rest of the tab

image

image

Closes #144


krbz999 (commented):
This may be a case for us to adopt a dnd5e'ism. On the item sheets, they handle storing the active editor allowing them to render just that without the rest of the contents of the tab, plus it is not cancelled if the sheet is re-rendered due to other updates.

This has been asked about in the sys-dev channel as well, so may be a good way to showcase such a feature.

JPMeehan:
> This may be a case for us to adopt a dnd5e'ism. On the item sheets, they handle storing the active editor allowing them to render just that without the rest of the contents of the tab, plus it is not cancelled if the sheet is re-rendered due to other updates.
> 
> This has been asked about in the sys-dev channel as well, so may be a good way to showcase such a feature.

Seems reasonable; is this something you want to take point on?

krbz999:
> Seems reasonable; is this something you want to take point on?

It is not at this time.

JPMeehan:
Significant revision to the implementation, now relying on the sheet to do its own editor management.

image

image

krbz999 (approved):

PR #150, opened 2025-01-11
Implemented Damage Roll

### LINKED ISSUE #149: Implement Damage Roll
Add a DamageRoll roll subclass that has specific tracking for things like damage types, and leverage it in AbilityModel#use

JPMeehan:
Reorganized some of the files here, too many classes in one rolls file.

Closes #149


draw-steel.mjs
```diff
@@ -9,15 +10,17 @@ globalThis.ds = {
...
 };
 
 /** Special global access */
-globalThis.PowerRoll = helpers.rolls.PowerRoll;
-globalThis.ProjectRoll = helpers.rolls.ProjectRoll;
-globalThis.SavingThrowRoll = helpers.rolls.SavingThrowRoll;
+globalThis.DamageRoll = rolls.DamageRoll;
+globalThis.PowerRoll = rolls.PowerRoll;
+globalThis.ProjectRoll = rolls.ProjectRoll;
+globalThis.SavingThrowRoll = rolls.SavingThrowRoll;
```
krbz999, line 23: Generally not a fan of polluting `globalThis` more than with the system object. Is this necessary?

krbz999 (approved):

draw-steel.mjs
JPMeehan: My thought process is that this is better for macro users, but this does move things to just `ds.rolls` so that's probably sufficient

PR #154, opened 2025-01-12
Improved Actor Abilities Tab

JPMeehan:
- Expanded click handling (can now use abilities from the sheet)
- Implemented setting the sheet mode via render options for both actor and item sheets
- Very light styling touch ups

image

image

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -329,6 +329,9 @@
               "label": "Type"
             }
           },
+          "damageDisplay": {
+            "label": "Damage Display"
+          },
           "trigger": {
             "label": "Trigger"
           },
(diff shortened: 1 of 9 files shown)


krbz999 (approved):
Looks good overall. A lot of this will change with v13 as your comments also indicate, same with the context menu display.

Is it worth slowly moving this to v13 starting soon?

JPMeehan:
> Looks good overall. A lot of this will change with v13 as your comments also indicate, same with the context menu display.
> 
> Is it worth slowly moving this to v13 starting soon?

I think once v13 testing drops we can pull the trigger.

PR #159, opened 2025-01-14
Heroic Resource Automation

JPMeehan:
- Handling for spending resources
- Handling for gaining resources

May need to finesse things for stuff like the Talent's clarity, which can go negative

### DIFF
--- a/src/module/data/actor/base.mjs
+++ b/src/module/data/actor/base.mjs
@@ -168,6 +168,7 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
 
   /**
    * Updates performed at the start of combat
+   * @param {DrawSteelCombatant} combatant The combatant representation
    */
   async startCombat(combatant) {
     await combatant.update({initiative: this.combat.turns});
@@ -175,7 +176,7 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
 
   /**
    * Updates performed at the start of this actor's turn
-   * @param {DrawSteelCombatant} combatant
+   * @param {DrawSteelCombatant} combatant The combatant representation
    * @abstract
    */
   async _onStartTurn(combatant) {}
--- a/src/module/data/actor/base.mjs
+++ b/src/module/data/actor/base.mjs
@@ -177,7 +177,7 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
    * Updates performed at the start of this actor's turn
    * @param {DrawSteelCombatant} combatant
    */
-  async startTurn(combatant) {}
+  async _onStartTurn(combatant) {}
 
   /**
    * Prompt the user for what types
--- a/src/module/data/actor/base.mjs
+++ b/src/module/data/actor/base.mjs
@@ -1,3 +1,4 @@
+/** @import {DrawSteelCombatant} from "../../documents/combatant.mjs"; */
 import {damageTypes, requiredInteger, setOptions, SizeModel} from "../helpers.mjs";
 const fields = foundry.data.fields;
 
@@ -166,11 +167,17 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
   }
 
   /**
-   * Update combatant at the start of combat
+   * Updates performed at the start of combat
    */
   async startCombat(combatant) {
     await combatant.update({initiative: this.combat.turns});
   }
+
+  /**
+   * Updates performed at the start of this actor's turn
+   * @param {DrawSteelCombatant} combatant
+   */
+  async startTurn(combatant) {}
 
   /**
    * Prompt the user for what types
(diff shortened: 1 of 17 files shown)


src/module/data/actor/base.mjs
```diff
@@ -166,12 +167,18 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
...
    */
   async startCombat(combatant) {
     await combatant.update({initiative: this.combat.turns});
   }
 
+  /**
+   * Updates performed at the start of this actor's turn
+   * @param {DrawSteelCombatant} combatant
+   */
+  async _onStartTurn(combatant) {}
```
krbz999, line 180:
```suggestion
  /**
   * Updates performed at the start of this actor's turn
   * @param {DrawSteelCombatant} combatant
   * @abstract
   */
  async _onStartTurn(combatant) {}
```

src/module/data/actor/base.mjs
```diff
@@ -166,12 +167,18 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
   }
 
   /**
-   * Update combatant at the start of combat
+   * Updates performed at the start of combat
```
krbz999, line 170: Missing jsdoc for parameter(s).

src/module/data/item/ability.mjs
```diff
@@ -31,7 +33,8 @@ export default class AbilityModel extends BaseItemModel {
 
     schema.keywords = new fields.SetField(setOptions());
     schema.type = new fields.StringField({required: true, blank: false, initial: "action"});
-    schema.category = new fields.StringField({required: true, nullable: false}),
+    schema.category = new fields.StringField({required: true, nullable: false});
+    schema.resource = new fields.NumberField({initial: null, min: 1, integer: true});
```
krbz999, line 37: How does it make sense to be both nullable and min 1?

src/module/data/item/ability.mjs
```diff
@@ -72,7 +75,10 @@ export default class AbilityModel extends BaseItemModel {
       tier3: new fields.SchemaField(powerRollSchema())
     });
     schema.effect = new fields.StringField();
-    schema.spend = new fields.NumberField({integer: true});
+    schema.spend = new fields.SchemaField({
+      value: new fields.NumberField({integer: true}),
+      text: new fields.StringField()
```
krbz999, line 80:
```suggestion
      text: new fields.StringField({required: true})
```

src/module/data/item/ability.mjs
```diff
@@ -228,8 +252,61 @@ export default class AbilityModel extends BaseItemModel {
...
+
+      content += foundry.applications.fields.createFormGroup({
+        label: game.i18n.format("DRAW_STEEL.Item.Ability.ConfigureUse.SpendLabel", {
+          value: this.spend.value || "",
+          name: this.resourceName
+        }),
+        input: spendInput
+      }).outerHTML;
+
+      configuration = await foundry.applications.api.DialogV2.confirm({
```
krbz999, line 293: This is, I suppose, a dnd5e'ism to have a item usage configuration dialog. It is a good one, though.

We should consider moving it into its own class, one that accepts an item or some other configuration parameters.

src/module/data/item/ability.mjs
```diff
@@ -228,8 +252,61 @@ export default class AbilityModel extends BaseItemModel {
...
+        content,
+        window: {
+          title: "DRAW_STEEL.Item.Ability.ConfigureUse.Title",
+          icon: "fa-solid fa-gear"
+        },
+        yes: {
```
krbz999, line 299: I think `DialogV2.prompt` is more appropriate in the meantime.

src/module/data/item/ability.mjs
```diff
@@ -228,8 +252,61 @@ export default class AbilityModel extends BaseItemModel {
...
+          callback: (event, button, dialog) => {
+            return new FormDataExtended(button.form).object;
+          }
+        },
+        rejectClose: false
+      });
+
+      if (!configuration) throw new Error("Configuration required but not provided");
```
krbz999, line 307: If the dialog was cancelled, that should be taken to mean that the user didn't want to use the item after all, no? So it shouldn't throw an error, it should just abort.

krbz999 (approved):

src/module/data/item/class.mjs
```diff
@@ -36,7 +37,8 @@ export default class ClassModel extends AdvancementModel {
     });
 
     schema.primary = new fields.StringField({required: true});
-    // schema.secondary = new fields.StringField();
+
+    schema.recovery = new FormulaField();
```
cdverrett94, line 41: Should this be named something different so as to not be confused with stamina recovery?

src/module/data/item/ability.mjs
JPMeehan: Either this isn't an ability that costs a heroic resource (null) or it's got a positive base cost

PR #162, opened 2025-01-15
Automate weakened effect

### LINKED ISSUE #122: Implement Weakened condition
- Probably implement in places that actually call power rolls?
- Validate with `actor.statuses.has('weakened')

cdverrett94:
Closes #122 

I placed in the prompt instead of the places calling PowerRoll so that it wouldn't have to be accounted for in several different places.


JPMeehan (approved):
May end up refactoring down the road but this works for now while we figure other stuff out.

PR #168, opened 2025-01-16
Add targets to ability power roll usage

### LINKED ISSUE #165: Account for multiple targets in a power roll
An ability with multiple targets uses the same power roll but can have different banes and edges.

cdverrett94:
Closes #165 

- I added a PowerRollDialog AppV2 class since I didn't think the normal dialog wasn't going to be sufficient. I add a prompt method too it similar to DialogV2's
- In the ability, I added methods for get the actor's modifiers that apply to all ability rolls (i.e. prone) and for getting modifiers that apply to a specific target (i.e. frightened, taunted, grabbed)
- I tried to keep the return type of the ability use method similar. Changed from message to array of messages
- In the power roll class, I added a method for getting the actor's modifiers that apply to all power rolls (i.e. weakness)
- I only included applying weakness in this PR, since it was there from my last PR.

Let me know what you think. Totally cool if you think there's a better way to handle this.

Dialog with Targets:
dialog with targets

Dialog without Targets
dialog without targets

Test Roll Dialog
test roll


src/module/rolls/power.mjs
```diff
@@ -284,4 +279,11 @@ export class PowerRoll extends DSRoll {
 
     return context;
   }
+
+  /** Modify the options object based on conditions that apply to all Power Rolls
+   * @param {object} [options] Options for the dialog
+   */
+  static getActorModifiers(options) {
+    if (options.actor?.statuses.has("weakness")) options.modifiers.banes += 1;
```
JPMeehan, line 287: Need to annotate `actor` as a possible option, I'm also wondering if there's a better way to handle/highlight this part of things since the actor does seem so central.

JPMeehan (changes requested):
Still need to test stuff but highlighted various repo-maintenance things like additional comments and formatting

cdverrett94:
I think I've cleaned up the jsdoc issues.

src/module/rolls/power.mjs
cdverrett94: My thought would be check if `options.actor` is a `DrawSteelActor`. If not, potentially set `options.actor` to the result of `DrawSteelChatMessage.getSpeakerActor`. If there's still no actor, then just return early.

templates/rolls/power-roll-dialog.hbs
```diff
@@ -0,0 +1,51 @@
...
+  {{/each}}
+  {{#if skills}}
+  <label>
+    <span>{{localize "DRAW_STEEL.Actor.Character.FIELDS.hero.skills.label"}}</span>
+    <select name="skills">
+      {{selectOptions skills blank=(localize "None")}}
+    </select>
+  </label>
+  {{/if}}
+  <button type="submit">{{localize "DRAW_STEEL.Roll.Power.Prompt.Button"}}</button>
```
JPMeehan, line 50:
```suggestion
  <footer class="form-footer">
    <button type="submit">{{localize "DRAW_STEEL.Roll.Power.Prompt.Button"}}</button>
  </footer>
```

Matches the style from dialogs. Also, add an icon before the text; I assume you're already familiar with FontAwesome's free library?

templates/rolls/power-roll-dialog.hbs
```diff
@@ -0,0 +1,51 @@
+<div>
+  <div class="global group">
+    <div class="header">{{localize "DRAW_STEEL.Roll.Power.Prompt.GlobalModifiers"}}</div>
```
JPMeehan, line 3: Feels like this could use a bump up in font size, something like `font-size: var(--font-size-16);`

src/module/data/item/ability.mjs
```diff
@@ -335,26 +336,72 @@ export default class AbilityModel extends BaseItemModel {
...
+          const flavor = game.i18n.format("DRAW_STEEL.Item.Ability.DamageFlavor", {type: damageType});
+          const damageRoll = new DamageRoll(damageFormula, rollData, {flavor, type: damageType});
+          await damageRoll.evaluate();
+          dataCopy.rolls.push(damageRoll);
+        }
+
+        messages.push(DrawSteelChatMessage.create(dataCopy));
       }
+
+      return messages;
```
JPMeehan, line 378: This is returning an array of promises, we should be be using [Promise.allSettled](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) to only return a single promise.

src/module/data/item/ability.mjs
```diff
@@ -335,26 +336,72 @@ export default class AbilityModel extends BaseItemModel {
...
-      const damageFormula = tier.damage.value;
-      if (damageFormula) {
-        const damageType = ds.CONFIG.damageTypes[tier.damage.type]?.label ?? tier.damage.type;
-        const flavor = game.i18n.format("DRAW_STEEL.Item.Ability.DamageFlavor", {type: damageType});
-        const damageRoll = new DamageRoll(damageFormula, rollData, {flavor, type: damageType});
-        await damageRoll.evaluate();
-        messageData.rolls.push(damageRoll);
+
+      // Roll damage and create a message per powerRolls
+      const messages = [];
```
JPMeehan, line 361: This ends up embedding the text of the ability for each roll message, which seems like overkill. It's also not exceptionally clear from the messages which is for which target. I think there needs to be more work on the output display here, preferably grouping targets if they have identical results which will happen a lot. Furthermore, only the first message needs the bulk of the embed; the rest of them will only need the text for the individual tier it achieves.

JPMeehan (changes requested):
Had a chance to test - I think there's really good work here but it needs a little more cook time, even if that grows the PR in scope.

templates/rolls/power-roll-dialog.hbs
cdverrett94: The only applicable icon I could think would be a die, but the d10 icon is a pro only icon. Of the free options, [this one](https://fontawesome.com/icons/dice?f=classic&s=solid) would be my recommendation

cdverrett94:
So here's currently where I'm at. I've got the rolls grouped by tier. I'm not totally sure the best way to prevent the ability embed. Can potentially add to the ability use system schema for embedText and set it to true only for the first message.

updated-chat-entry

src/module/rolls/power.mjs
JPMeehan: I think what makes sense here is extracting `options` to an interface (like PowerRollPromptOptions) in the adjacent types file that is then used as the param type for both `prompt` and this helper function, `getActorModifiers`

templates/rolls/power-roll-dialog.hbs
JPMeehan: Works for me, it's unfortunate we don't have the d10s but I'm fine with using the pipped d6

JPMeehan:
> So here's currently where I'm at. I've got the rolls grouped by tier. I'm not totally sure the best way to prevent the ability embed. Can potentially add to the ability use system schema for embedText and set it to true only for the first message.

Yeah, a schema property I think is absolutely the right solution here so it can just be part of that message data. I think make it a boolean field with `initial: true`, and then just add a check on whether to make the embed.

cdverrett94:
Here's the updated chat card preview with the ability text on only the first message

updated-chat-entry

src/module/rolls/power.mjs
cdverrett94: I took it a little farther, so hopefully it's right. Sorry, still learning the TS/typing stuff so kind of having to figure it out as I go.

- I added a PowerRollModifiers interface since that was coming up multiple places.
- Added the PowerRollPromptOptions interface per your request
- Added an AbilityUseOptions as well that extends PowerRollPromptOptions. In the `use` method, The only option specific to this was the event.

With these interfaces defined, is it necessary to still list what is in the options object as `@params`?

Let me know if I messed this up.

JPMeehan: Yeah so once you define the interface that's all you need for that parameter, it's a really handy way to keep multiple functions that share argument formats in sync

src/module/data/item/ability.mjs
```diff
@@ -247,11 +249,10 @@ export default class AbilityModel extends BaseItemModel {
 
   /**
    * Use an ability, generating a chat message and potentially making a power roll
-   * @param {object} [options={}] Configuration
+   * @param {PartialAbilityUseOptions>} [options={}] Configuration
```
JPMeehan, line 252:
```suggestion
   * @param {Partial<AbilityUseOptions>} [options={}] Configuration
```

I think that's why this wasn't quite working for you, which also means you can leave off the rest of the param definition because the interface handles it.

src/module/data/item/ability.mjs
```diff
@@ -335,26 +336,84 @@ export default class AbilityModel extends BaseItemModel {
...
         formula,
         data: rollData,
-        evaluation: "evaluate",
-        banes: options.banes,
-        edges: options.banes
+        evaluation: "evaluate",      
+        actor: this.actor,
+        modifiers: {
+          banes: options.banes ?? 0,
+          edges: options.banes ?? 0
```
JPMeehan: Wouldn't these now be options.modifiers.banes?

src/module/apps/power-roll-dialog.mjs
```diff
@@ -0,0 +1,107 @@
...
+
+  /** 
+   * Set a final context for resolving the prompt, then close the dialog 
+   * @override
+   */
+  async _onSubmitForm(formConfig, event) {
+    const targets = this.options.context.targets;
+    if (!targets || (targets.length === 0)) this.finalContext = [this.options.context.modifiers];
+    else {
+      this.finalContext = targets.reduce((accumlator, target) => {
```
JPMeehan, line 79: spelling: f2 replace with `accumulator`

src/module/rolls/power.mjs
cdverrett94: Cool I've removed the unnecessary params. I also realized that AbilityUseOptions shouldn't extend PowerRollPromptOptions. so corrected that to just be the event and modifiers.

src/module/rolls/power.mjs
```diff
@@ -150,36 +145,35 @@ export class PowerRoll extends DSRoll {
...
+      roll.terms[0] = baseRoll.terms[0];
+      switch (evaluation) {
+        case "none":
+          rolls.push(roll);
+          break;
+        case "evaluate":
+          rolls.push(await roll.evaluate());
+          break;
+        case "message":
+          rolls.push(await roll.toMessage());
```
JPMeehan, line 172: Now that we have the `actor` as an option, I'm noticing the speaker isn't reliably assigned correctly for the plain characteristic roll; I think we should be passing the speaker data here. Actor should be the same for all roll contexts so can derive the speaker just before the loop.

JPMeehan (changes requested):
Overall a lot of progress. I'm also realizing that the roll targets is just stuffing an entire actor's data inside? That's going to bloat things quickly, I think some refactoring to pass a UUID and then retrieve that would be good. Can use `game.i18n.localize(Unknown)` if UUID retrieval fails.

src/module/data/item/ability.mjs
cdverrett94: Again.... *facepalm*

src/module/data/item/ability.mjs
cdverrett94: Yep, I had actually corrected that while you were reviewing I think

src/module/rolls/power.mjs
cdverrett94: I had already updated the rollCharacteristic to pass `actor: this.parent` to the prompt. I've also (while you were reviewing I think) added a nullish coalescing assignment of `options.actor` to `DrawSteelChatMessage.getSpeeakerActor`.

cdverrett94: I see what you meant now. Fixed.

JPMeehan: Most recent push fixes

JPMeehan (approved):
Looks good - I think the PowerRollDialogPrompt interface needs an update?

cdverrett94:
You're right. 

Thanks for helping me work through all the stuff on this one.

PR #176, opened 2025-01-17
Automate frightened condition

### LINKED ISSUE #166: Implement Frightened Condition
Can probably store the actor uuid of the imposing actor in `flags.frightened` or something. 

Once power rolls can accept targets, can check against this flag when attacking/being attacked

The UI for choosing who frightened you will probably be the most challenging bit

cdverrett94:
Closes #166 

Also fixed the formatting issues from your other PR and updated the return type of `getTargetModifiers` to `PowerRollModifiers`.


src/module/data/item/ability.mjs
```diff
@@ -399,19 +399,23 @@ export default class AbilityModel extends BaseItemModel {
...
 
-  /** 
+  /**
    * Get the modifiers based on conditions that apply to ability Power Rolls specific to a target
    * @param {DrawSteelActor} actor The actor using the ability
    * @param {DrawSteelActor} target A target of the Ability Roll
-   * @returns {object} 
+   * @returns {PowerRollModifiers}
    */
   getTargetModifiers(actor, target) {
```
JPMeehan, line 408: Looking at this again; shouldn't the actor using the ability just always be `this.actor`? If so, it probably isn't a good param to include.

src/module/data/item/ability.mjs
```diff
@@ -399,19 +399,23 @@ export default class AbilityModel extends BaseItemModel {
...
     const modifiers = {
       banes: 0,
       edges: 0
     };
 
-    //TODO: CONDITION CHECKS
+    //TODO: ALL CONDITION CHECKS
+    
+    // Frightened condition checks
+    if (actor.statuses.has("frightened") && (actor.flags[systemID]?.frightened === target.uuid)) modifiers.banes += 1; // Attacking the target frightening the actor
```
JPMeehan, line 417: use `getFlag` it's also synchronous access and reads better.

JPMeehan (changes requested):

JPMeehan:
I'm also realizing that one issue with the current implementation for Frightened that's more systemic is that you can be frightened of multiple sources

krbz999:
For ease of tracking, it could be tracked on the source rather than the target.

cdverrett94:
> I'm also realizing that one issue with the current implementation for Frightened that's more systemic is that you can be frightened of multiple sources; probably need to build an Array for it, and honestly consider just making this a proper Set property in BaseActor. Maybe some kind of `system.statuses` object that can hold additional data.

I'm not entirely sure what you mean. each actor should only have one source of the frightened condition.

JPMeehan:
> > I'm also realizing that one issue with the current implementation for Frightened that's more systemic is that you can be frightened of multiple sources; probably need to build an Array for it, and honestly consider just making this a proper Set property in BaseActor. Maybe some kind of `system.statuses` object that can hold additional data.
> 
> I'm not entirely sure what you mean. each actor should only have one source of the frightened condition.

Double checked, that is true. 
>  If you gain the frightened condition from one source while already frightened by a different source, the new condition replaces the old one.
Then the issue is figuring out a way to reliably resolve the order, especially in cases where something might be "Frightened + thing"

src/module/data/item/ability.mjs
```diff
@@ -399,19 +399,22 @@ export default class AbilityModel extends BaseItemModel {
...
       banes: 0,
       edges: 0
     };
 
-    //TODO: CONDITION CHECKS
+    //TODO: ALL CONDITION CHECKS
+    
+    // Frightened condition checks
+    if (this.actor.statuses.has("frightened") && (this.actor.getFlag(systemID, "frightened") === target.uuid)) modifiers.banes += 1; // Attacking the target frightening the actor
+    if (target.statuses.has("frightened") && (this.actor.uuid === target.getFlag(systemID, "frightened"))) modifiers.edges += 1; // Attacking the target the actor has frightened
```
krbz999, line 417: I think the `ActiveEffect` class is a better place for this. It can support a method such as `isSourceOf(actorA, actorB, status)` and return a boolean.

In combination with allowing for multiple sources of a status, this would reduce it to iterating over two actors' effects.

cdverrett94: I just want to clarify what you're recommending: 
- Replace `this.actor.statuses.has("frightened") && (this.actor.getFlag(systemID, "frightened") === target.uuid)` with `DrawSteelActiveEffect.isSourceOf(actorA, actorB, status)`
- Or that the entire concept of target specific modifier application should be in the `ActiveEffect` class.

cdverrett94:
> I'm also realizing that one issue with the current implementation for Frightened that's more systemic is that you can be frightened of multiple sources; probably need to build an Array for it, and honestly consider just making this a proper Set property in BaseActor. Maybe some kind of `system.statuses` object that can hold additional data.

While this won't be applicable for frightened or taunted sources, it would be applicable to the grabbed condition since nothing says it's limited to only 1 source.

So what I'm thinking is add `statuses` `SchemaField` to the `BaseActorModel`. It will have all of the conditions. Each condition will have it's own `SchemaField` with a `sources` property. The `sources` property will be an array of document uuids. This array will be a custom field to impose max sizes without erroring. It will slice off the last X elements. This will ensure frightened and taunted will always have a length of 1. I'm pretty sure active effects are already ordered by when they were added, so if there were 10 AE's that set the frightened source, then only the last one would apply.

src/module/data/item/ability.mjs
krbz999: Yes to the former. The latter I will leave to your discretion.

krbz999:
I don't think a new schema field is necessary, that would also quickly get bloated. A property can be instantiated during data preparation if needed, or it can be done by the AE itself.

cdverrett94:
I've made the following changes:
- Added `statuses` during actor model base data preparation. In derived data preparation, the status sources is sliced to the max length allowed.
- Added `DrawSteeActiveEffect.isStatusSource` for determining if a given actor is being given a status by another actor.
- Updated the frightened checks in `getTargetModifiers` to use the new `isStatusSource` method.

I wasn't sure if the `statuses` property should only include targeted conditions or all of them, so I just included them all.

Let me know your thoughts.

src/module/data/actor/base.mjs
```diff
@@ -80,13 +80,27 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
     this.potency = {
       bonuses: 0
     };
+
+    // Add status sources object for AEs to target
+    this.statuses = {};
+    for (const [statusId, config] of Object.entries(ds.CONFIG.conditions)) {
+      this.statuses[statusId] = {
+        sources: []
```
krbz999, line 88: This should be a `Set`, but that might require some additional work in the AE document class and isn't something that needs to hold up this PR.

What I *do* think would be good is moving this instantiation entirely to the methods that apply the changes in the AE class. That means we don't need to instantiate it for every status but only the ones that are present and when present.

src/module/data/actor/base.mjs
```diff
@@ -80,13 +80,27 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
...
   /** @override */
   prepareDerivedData() {
     super.prepareDerivedData();
 
     this.stamina.winded = Math.floor(this.stamina.max / 2);
+
+    // Limit the status sources array to the max length provided in the config
+    for (const statusId in this.statuses) {
+      const maxSources = ds.CONFIG.conditions[statusId].maxSources;
+      if (maxSources) this.statuses[statusId].sources = this.statuses[statusId].sources.slice(-maxSources);
```
krbz999, line 102: Likewise this logic can also be moved into the AE class, though I question whether we want to enforce the limit on the length of the array of sources, at least yet.

src/module/data/actor/base.mjs
cdverrett94: Already I've moved this to the Active Effect class and converted it to a Set. `ActiveEffect` doesn't handle `Set`s as far as I can tell. So I took a look at the `SetField` to see what it was doing. Not sure if converting to/from `Set`/`Array` is the best way to manage it, but it seems like that's what `SetField` is doing for it's casting methods.

src/module/data/actor/base.mjs
cdverrett94: I've moved this into the `ActiveEffect` class. If y'all decide we don't want to enforce the limit, I can easily remove it.

krbz999 (commented):
LGTM! Tossing back to Joseph.

JPMeehan (approved):
Yeah this is a lot cleaner and extensible

PR #188, opened 2025-01-18
Add potency information to chat card

### LINKED ISSUE #177: Add Potency checks to ability use chat card
Probably need to have the stat to check against as part of the model. Probably could also benefit from some condensing in the UI; it's really long for a lot of frankly short inputs

cdverrett94:
Closes #177

Summary of changes:
- Updated potency property to a schema that includes `enabled`, `characteristic`, and `value`. Value has an initial value of the relevant potency values roll data (i.e. @potency.weak).
- Added a `getPotencyData` method for retrieving the potency data for a given tier. This will evaluate the potency value and generate the embed text.
- Added the ability uuid to the roll data, so the ability info can be pulled during the roll's context preparation.
- The roll chat card now displays the potency embed text and whether the given target passes the potency value.
- I also updated the chat card to remove the duplicate formulas. Now the base power roll is displayed and for each target/roll only the total is shown. This will save space if there are many targets.
- Random bug fix for ability distance's not localizing the tertiary values.

Updated chat card
update-chat-card

### DIFF
--- a/src/module/data/item/ability.mjs
+++ b/src/module/data/item/ability.mjs
@@ -1,6 +1,6 @@
 import {systemPath} from "../../constants.mjs";
 import {DrawSteelActiveEffect, DrawSteelActor, DrawSteelChatMessage} from "../../documents/_module.mjs";
-import {DamageRoll, PowerRoll} from "../../rolls/_module.mjs";
+import {DamageRoll, DSRoll, PowerRoll} from "../../rolls/_module.mjs";
 import FormulaField from "../fields/formula-field.mjs";
 import {setOptions} from "../helpers.mjs";
 import BaseItemModel from "./base.mjs";
@@ -54,13 +54,17 @@ export default class AbilityModel extends BaseItemModel {
       value: new fields.NumberField({integer: true})
     });
 
-    const powerRollSchema = () => ({
+    const powerRollSchema = ({initialPotency} = {}) => ({
       damage: new fields.SchemaField({
         value: new FormulaField(),
         type: new fields.StringField({required: true})
       }),
       ae: new fields.SetField(setOptions({validate: foundry.data.validators.isValidId})),
-      potency: new FormulaField({deterministic: true}),
+      potency: new fields.SchemaField({
+        enabled: new fields.BooleanField(),
+        characteristic: new fields.StringField(),
+        value: new FormulaField({deterministic: true, initial: initialPotency, required: false})
+      }),
       forced: new fields.SchemaField({
         type: new fields.StringField({choices: config.forcedMovement, blank: false}),
         value: new fields.NumberField(),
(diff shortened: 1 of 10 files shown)


JPMeehan:
What's the value of checking enabled vs. seeing if characteristic has something set?

cdverrett94:
I think for enabled, cleaner display on the sheet to hide the fields and a more clear check than potency.characteristic !== "". But I suppose I'm not totally attached to the idea.

I'm okay with making the potency characteristic a top level property. In that case, I think enabled makes more sense as well because there are examples where only 1 tier has an effect that has a potency.

JPMeehan:
Good spot on only some tiers having potency, keep enabled and just have the top level characteristic choice. I think the potency stuff should be hidden on the tiers if there's no characteristic set

cdverrett94:
While I'm in this, I'm wondering. Should the potency stuff be a part of the AE stuff? Like have the `ae` be a `SchemaField` with `effects` and `potency`? Is there a situation where there would be a potency but not an AE as well?

JPMeehan:
So the tricky part is you can have
- AE without potency 
- AE strictly from potency
- Split AE that's always applied + more with potency

cdverrett94:
Good point. Took me a while of searching through the hero doc to find those examples, but I see them. I'll leave it as is.

I updated to have a top level `potencyCharacteristic`. In `getPotencyData` I updated it to only send the full data if the potency is enabled and there is a characteristic set.

JPMeehan:
Monster doc is also important and frequently does weirder things

src/module/data/item/ability.mjs
```diff
@@ -174,6 +178,31 @@ export default class AbilityModel extends BaseItemModel {
...
+   */
+  async getPotencyData(tierName) {
+    const potency = this.powerRoll[tierName].potency;
+    const potencyData = {
+      enabled: potency.enabled && !!this.powerRoll.potencyCharacteristic
+    };
+
+    if (!potencyData.enabled) return potencyData;
+
+    const potencyValue = (await new DSRoll(potency.value, this.parent.getRollData()).evaluate()).total;
```
JPMeehan, line 195: Can't we just do evaluateSync here, since potency.value is deterministic

JPMeehan (approved):
Some small details to fix - I'm also wondering if we can improve the styles on the potency display in the chat card, definitely want to add some <span> pieces to wrap things for targeting styles later.

cdverrett94:
> Some small details to fix - I'm also wondering if we can improve the styles on the potency display in the chat card, definitely want to add some pieces to wrap things for targeting styles later.

I've wrapped the individual components into their own spans so they can be targeted with CSS. Any specific ideas on styling it better?

JPMeehan:
> I've wrapped the individual components into their own spans so they can be targeted with CSS. Any specific ideas on styling it better?

Bolding `Potency:`, and I feel like some kind of distinction for success vs. failure is justified here?

cdverrett94:
potency-styling

Styling per suggestions.

JPMeehan:
Says there are merge conflicts

cdverrett94:
Resolved that!

PR #189, opened 2025-01-18
Added initial and placeholder values for ability roll bonus

JPMeehan:
Theory is simple - `@chr` will evaluate to 0 if no characteristics are set, but otherwise this is going to be the default for a *lot* of powers.

image

### DIFF
--- a/src/module/data/item/ability.mjs
+++ b/src/module/data/item/ability.mjs
@@ -71,7 +71,7 @@ export default class AbilityModel extends BaseItemModel {
 
     schema.powerRoll = new fields.SchemaField({
       enabled: new fields.BooleanField(),
-      formula: new FormulaField({initial: "@chr"}),
+      formula: new FormulaField({blank: false, initial: "@chr"}),
       characteristics: new fields.SetField(setOptions()),
       tier1: new fields.SchemaField(powerRollSchema()),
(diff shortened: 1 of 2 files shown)


src/module/data/item/ability.mjs
```diff
@@ -71,7 +71,7 @@ export default class AbilityModel extends BaseItemModel {
 
     schema.powerRoll = new fields.SchemaField({
       enabled: new fields.BooleanField(),
-      formula: new FormulaField(),
+      formula: new FormulaField({blank: false, initial: "@chr"}),
```
krbz999, line 74: I feel like it's cleaner to (1) allow blank, (2) treat a falsy value as `@chr`, (3) add a placeholder in all form inputs of `"@chr"`, and maybe even (4) set the property to `@chr` in derived data if blank (but using the source data when editing ofc)

That is, of course, entirely from a UX point of view. With a string field that does not allow blank values, users will still try removing the value to get the default but receive an error since this is not allowed.

JPMeehan: Not true actually, StringField#clean will turn the empty string into undefined, which then lets `initial` kick in without error.

krbz999 (approved):

PR #190, opened 2025-01-19
Reduce speeds when slowed

### LINKED ISSUE #121: Implement Slowed condition
- Reduce speeds to 2 in prepareDerivedData
  - Check against some flag first and then null coalesce to 2, that way stuff that adjusts your slowed speed (e.g. the human trait that increases it to 3) can work
- Validate with `actor.statuses.has('slowed')`

cdverrett94:
Closes #121 

Summary of Change:
- Added `defaultSpeed` to the slowed condition config.
- In `ActiveEffect.apply` if there's a speed change, set `system.statuses.slowed.speed` to `defaultSpeed`. This is necessary for the upgrade mode to apply. I made the AE key similar to the condition sources one.
- In `Actor.prepareDerivedData` set all movement speeds to either `system.statuses.slowed.speed` or the `defaultSpeed`. 
- In testing this, I realized that I messed up on the frightened PR. It was always converting the value to a Set instead of only when it was a change to a condition source.

There is currently an issue with `ActiveEffect._applyUpgrade` potentially making the value `undefined`. Looks like it's fixed in V13. https://github.com/foundryvtt/foundryvtt/issues/11527. Should I apply an override to that method to fix it until we swap to V13?

### DIFF
--- a/src/module/data/actor/base.mjs
+++ b/src/module/data/actor/base.mjs
@@ -80,6 +80,12 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
     this.potency = {
       bonuses: 0
     };
+
+    this.statuses = {
+      slowed: {
+        speed: ds.CONFIG.conditions.slowed.defaultSpeed
+      }
+    };
   }
 
   /** @override */
@@ -90,9 +96,8 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
 
     // Reduce all movement speeds when slowed
     if (this.parent.statuses.has("slowed")) {
-      const slowedSpeed = this.statuses?.slowed?.speed ?? ds.CONFIG.conditions.slowed.defaultSpeed;
       for (const movement in this.movement) {
-        if (this.movement[movement] > slowedSpeed) this.movement[movement] = slowedSpeed;
+        if (this.movement[movement] > this.statuses.slowed.speed) this.movement[movement] = this.statuses.slowed.speed;
       }
     }
   }
--- a/src/module/data/actor/base.mjs
+++ b/src/module/data/actor/base.mjs
@@ -87,6 +87,13 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
     super.prepareDerivedData();
 
     this.stamina.winded = Math.floor(this.stamina.max / 2);
+
+    // Reduce all movement speeds when slowed
+    if (this.parent.statuses.has("slowed")) {
+      for (const movement in this.movement) {
+        this.movement[movement] = this.statuses?.slowed?.speed ?? ds.CONFIG.conditions.slowed.defaultSpeed;
+      }
+    }
   }
 
   /**
--- a/src/module/data/actor/base.mjs
+++ b/src/module/data/actor/base.mjs
@@ -90,8 +90,8 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
 
     // Reduce all movement speeds when slowed
     if (this.parent.statuses.has("slowed")) {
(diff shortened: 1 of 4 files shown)


src/module/data/actor/base.mjs
```diff
@@ -87,6 +87,13 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
     super.prepareDerivedData();
 
     this.stamina.winded = Math.floor(this.stamina.max / 2);
+
+    // Reduce all movement speeds when slowed
+    if (this.parent.statuses.has("slowed")) {
+      for (const movement in this.movement) {
```
JPMeehan, line 93: What if this.movement is null for a type, e.g. no flying.

JPMeehan (changes requested):
I'm also good for adjusting overriding _applyUpgrade with a TODO: Remove in v13

src/module/data/actor/base.mjs
cdverrett94: Good catch, also forgot to apply only if their speed is higher.

src/module/data/actor/base.mjs
```diff
@@ -87,6 +87,14 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
...
+        const slowedSpeed = this.statuses?.slowed?.speed ?? ds.CONFIG.conditions.slowed.defaultSpeed;
```
JPMeehan, line 94: This can be outside of the loop

src/module/documents/active-effect.mjs
```diff
@@ -84,9 +84,18 @@ export class DrawSteelActiveEffect extends ActiveEffect {
     return this.system._isTemporary ?? super.isTemporary;
   }
 
+  /** @override */
+  apply(actor, change) {
+    // If there's a change to the slowed speed, and the property does not exist, set it to the default slowed speed. This allows for upgrade/downgrade to apply.
+    const slowedSpeedChange = change.key.match(/^system\.statuses\.slowed\.speed$/);
+    if (!!slowedSpeedChange && !actor.system.statuses?.slowed?.speed) foundry.utils.mergeObject(actor, {"system.statuses.slowed.speed": ds.CONFIG.conditions.slowed.defaultSpeed});
```
JPMeehan, line 91: feels like foundry.utils.hasProperty and foundry.utils.setProperty are more correct here. I don't know why `system.statuses.slowed.speed` would be 0, but you'd get the wrong return here.

JPMeehan (approved):
Some small fixes

src/module/documents/active-effect.mjs
cdverrett94: Didn't know about those methods.

That's what I get for copying.

JPMeehan: `!hasProperty` instead of `getProperty` in this case, we want to return true if the property is missing

JPMeehan:
I'm realizing we're all doing this wrong. Shouldn't we just set the slowed property in prepareBaseData, this both gets rid of the `apply` override *and* means you don't need to null coalesce when adjusting speed?

cdverrett94:
I had that idea at the beginning, but based on the comments on the frightened PR, it seemed like the desire was to have this stuff controlled by the AE class. I was going with the same thing. You're right, setting it in actor data prep would make this easier.

JPMeehan:
> I had that idea at the beginning, but based on the comments on the frightened PR, it seemed like the desire was to have this stuff controlled by the AE class and to instantiate it when necessary. I was going with the same thing. You're right, setting it in actor data prep would make this easier.

I think prepareBaseData is a much simpler spot than overriding apply - the situation with Frightened was more about tracking application sources with the effect doing the application.

cdverrett94:
I agree. I misunderstood the specifics on what should be in the AE class.

Thanks!

PR #191, opened 2025-01-22
Added power roll results to embeds

### LINKED ISSUE #178: Add power result to embed display
- Use the tier `description` to avoid worrying about parsing
- Ideally some special config options - if the tier is *not* provided, include all three, but if tier is provided only share one of them.

JPMeehan:
Further leverages the embed system to conditionally show descriptions for each power roll tier, adding clarity to the play mode sheet as well as the chat rolls

Closes #178

image

image

### DIFF
--- a/src/module/data/message/ability-use.mjs
+++ b/src/module/data/message/ability-use.mjs
@@ -35,8 +35,6 @@ export default class AbilityUseModel extends BaseMessageModel {
 
     const content = html.querySelector(".message-content");
 
-    console.log(item, tier);
-
     if (this.embedText) {
       /** @type {HTMLDivElement} */
       let embed;
--- a/src/module/data/message/ability-use.mjs
+++ b/src/module/data/message/ability-use.mjs
@@ -26,23 +26,38 @@ export default class AbilityUseModel extends BaseMessageModel {
   async alterMessageHTML(html) {
     await super.alterMessageHTML(html);
 
-    if (!this.embedText) return;
-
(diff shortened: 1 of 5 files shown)


src/module/data/message/ability-use.mjs
```diff
@@ -26,23 +26,36 @@ export default class AbilityUseModel extends BaseMessageModel {
...
+        embed.innerText = game.i18n.localize("DRAW_STEEL.Item.Ability.EmbedFail");
+      }
+
+      // If it's a roll, the roll rendering will replace the message's stored content. Otherwise we need to do it.
+      if (this.parent.isRoll) content.insertAdjacentElement("afterbegin", embed);
+      else content.innerHTML = embed.outerHTML;
+    } else if (item && tier) {
+      content.insertAdjacentHTML("afterbegin", `<p><strong>${
+        game.i18n.localize(`DRAW_STEEL.Roll.Power.Results.Tier${tier}`)
+      }: </strong>${item.system.powerRoll[`tier${tier}`].description}</p>`
```
krbz999, line 57: Can this be blank? If so, it should be left out.

krbz999 (approved):

src/module/data/message/ability-use.mjs
JPMeehan: Every result does *something* and I'm figuring this is just the raw text of the ability for this. Need to improve the input styling to accommodate abilities with longer descriptions.

PR #193, opened 2025-01-23
Update drag-drop to use .draggable

### LINKED ISSUE #47: Consider adjusting DragDrop implementation to match v13 API
With v13 adding dragDrop handling to ActorSheetV2, specifically through the use of the .draggable class, we should try to match core to keep things DRY

cdverrett94:
Closes #47 

Summary of Changes:
- Updated the `dragSelector` to `.draggable` added this class to the relevant items. Features and kits didn't have the `[data-drag]`, but I added the class to them anyways.
- Updated the sheet context variables for `features`, `kits`, and `abilities` to be sorted by their sort value, so that they can be rearranged on the sheet.
- The NPC sheet wasn't calling `super._onRender`, so dragDrop wasn't being registered.

### DIFF
--- a/src/module/apps/actor-sheet/base.mjs
+++ b/src/module/apps/actor-sheet/base.mjs
@@ -1,4 +1,3 @@
-import {systemPath} from "../../constants.mjs";
 import {DrawSteelItem} from "../../documents/item.mjs";
 import {DrawSteelItemSheet} from "../item-sheet.mjs";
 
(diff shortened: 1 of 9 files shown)


JPMeehan (approved):
I had noticed something was weird about the NPC sheet but hadn't dug in, and yeah part of why I wanted to do this is it was a way to do a second pass on making sure all the dragDrop spots were happy. Good work.

PR #196, opened 2025-01-23
Add ability restrictions with dazed as an example

### LINKED ISSUE #118: Implement Dazed condition
Actor sheet should gray out/otherwise visually indicate which abilities are Triggered Actions, Free Triggered Actions, or Free Maneuvers and thus cannot be used while dazed. This shouldn't be a hard restriction but it would be a visual reminder. Validate by `actor.statuses.has('dazed')`

Probably blocked by improving the actor sheet visuals.

cdverrett94:
Closes #118 

Summary of Changes:
- Added `restrictions` object to actor in data prep. This object has a `Set` for `type` and `dsid`. 
- In the ability, check if the ability's `type` or `_dsid` is in the `Set`s, if so return `true`, otherwise return `false`. This has a fallback for checking actor statuses in case just the status is added to an ability without the use of toggle status effect.
- Display a warning sign next to the ability name with a tooltip advising that the ability is restricted.

I wasn't sure if there was a better spot to add the `Set` application to the AE class. I didn't add a way to track which sources are restricting an ability, but if that's desired, could add a `source` property on `restrictions`. In the AE, could check if the key is a restriction change and then add the name of the AE to the `source` `Set`. 

Let me know your thoughts.

Screenshot 2025-01-22 233833

### DIFF
--- a/src/module/data/actor/base.mjs
+++ b/src/module/data/actor/base.mjs
@@ -86,6 +86,11 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
         speed: ds.CONFIG.conditions.slowed.defaultSpeed
       }
     };
+
+    this.restrictions = {
+      type: new Set(),
+      dsid: new Set()
+    };
   }
 
   /** @override */
--- a/src/module/data/item/ability.mjs
+++ b/src/module/data/item/ability.mjs
@@ -455,4 +455,22 @@ export default class AbilityModel extends BaseItemModel {
 
     return modifiers;
   }
+
+  /**
+   * Determine if an Active Effect or a status is restricting this ability.
+   * @returns {boolean}
+   */
+  get restricted() {
+    if (!this.actor) return false;
+
+    // Checking if active effects have restricted this type
+    const restrictions = this.actor.system.restrictions;
+    if (restrictions.type.has(this.type)) return true;
+    if (restrictions.dsid.has(this._dsid)) return true;
+
+    // Checking if statuses have restricted this type. In case the main condition isn't toggled, but the status is added to an effect
+    if (this.actor.statuses.has("dazed") && ["freeManeuver", "triggered", "freeTriggered"].includes(this.type)) return true;
+
+    return false;
+  }
 }
--- a/lang/en.json
+++ b/lang/en.json
@@ -592,7 +592,8 @@
           "Embed": "{characteristic}<{value}",
           "Success": "Success",
           "Failure": "Failure"
-        }
+        },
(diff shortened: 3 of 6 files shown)


src/module/data/actor/base.mjs
```diff
@@ -86,6 +86,11 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
         speed: ds.CONFIG.conditions.slowed.defaultSpeed
```
JPMeehan, line 86: I'm remembering a separate declaration I made - I think we should be using `CONFIG.statusEffects.find(e => e.id === statusId)` instead of referring to ds.CONFIG, since otherwise we have an issue with people needing to declare any edits twice - once to ds.CONFIG and then once to CONFIG.statusEffects. I'm open to changing that decision but if so we should adjust the comment on our conditions object.

src/module/data/item/ability.mjs
```diff
@@ -455,4 +455,22 @@ export default class AbilityModel extends BaseItemModel {
...
+  get restricted() {
+    if (!this.actor) return false;
+
+    // Checking if active effects have restricted this type
+    const restrictions = this.actor.system.restrictions;
+    if (restrictions.type.has(this.type)) return true;
+    if (restrictions.dsid.has(this._dsid)) return true;
+
+    // Checking if statuses have restricted this type. In case the main condition isn't toggled, but the status is added to an effect
+    if (this.actor.statuses.has("dazed") && ["freeManeuver", "triggered", "freeTriggered"].includes(this.type)) return true;
```
JPMeehan, line 472: While I'm second guessing if this should also be on the CONFIG object somewhere I think this works just fine. I will separately note that we may need to have some kind of API for un-restricting stuff in the future but that's not a now problem and I think the current design allows for that adjustment.

JPMeehan (approved):
Some adjustments to consider. We may also need to work on the UI for abilities with long names.

src/module/data/actor/base.mjs
cdverrett94: I've swapped the instances of ds.CONFIG.conditions to use CONFIG.statusEffect instead.

src/module/data/item/ability.mjs
cdverrett94: I've added a restrictions entry on the conditions CONFIG. The type was getting too long for one line, so broke it into multiple lines.

cdverrett94:
> Some adjustments to consider. We may also need to work on the UI for abilities with long names.

Are we thinking that with the ability embed part being done that the character sheet should display that information for each ability? Or are we sticking with the button with just the name?

PR #198, opened 2025-01-24
Automate grabbed and restrained conditions

### LINKED ISSUE #119: Implement Grabbed condition
- Set non-teleport speeds to 0 in Derived Data
- Have some kind of visual indicator for actions with _dsid matching `knockback` (Overlap with #118)
- Validate with `actor.statuses.has('grabbed')`

### LINKED ISSUE #120: Implement Restrained condition
- Set non-teleport speeds to 0
- Have some kind of visual indicator for actions with _dsid matching `stand-up` (Overlap with #118)
- Validate with `actor.statuses.has('restrained')`

cdverrett94:
Closes #120, Closes #119 

I updated the status check in the ability `restricted` to a for loop instead of a check for each condition individually. It means we only just have to update the CONFIG if there's ever changes to what a condition restricts. It also means modules could add new restrictions on the CONFIG.statusEffects object and those will work too.

### DIFF
--- a/src/module/data/actor/base.mjs
+++ b/src/module/data/actor/base.mjs
@@ -99,10 +99,15 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
 
     this.stamina.winded = Math.floor(this.stamina.max / 2);
 
-    // Reduce all movement speeds when slowed
-    if (this.parent.statuses.has("slowed")) {
+    // Set movement speeds when affected by grabbed, restrained, or slowed
+    if (this.parent.statuses.has("grabbed") || this.parent.statuses.has("restrained") || this.parent.statuses.has("slowed")) {
       for (const movement in this.movement) {
-        if (this.movement[movement] > this.statuses.slowed.speed) this.movement[movement] = this.statuses.slowed.speed;
+        // If slowed, set all speeds to slowed speed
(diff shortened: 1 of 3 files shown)


src/module/data/actor/base.mjs
```diff
@@ -99,10 +99,15 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
...
-    if (this.parent.statuses.has("slowed")) {
+    // Set movement speeds when affected by grabbed, restrained, or slowed
+    if (this.parent.statuses.has("grabbed") || this.parent.statuses.has("restrained") || this.parent.statuses.has("slowed")) {
       for (const movement in this.movement) {
-        if (this.movement[movement] > this.statuses.slowed.speed) this.movement[movement] = this.statuses.slowed.speed;
+        // If slowed, set all speeds to slowed speed
+        if (this.parent.statuses.has("slowed") && (this.movement[movement] > this.statuses.slowed.speed)) this.movement[movement] = this.statuses.slowed.speed;
+
+        // If grabbed or restrained, set non-teleport speeds to 0
+        const isGrabbedOrRestrained = this.parent.statuses.has("grabbed") || this.parent.statuses.has("restrained");
```
JPMeehan, line 109: Feels like these boolean checks should be first and then the `if` can be `isSlowed || isGrabbedOrRestrained`.

JPMeehan (approved):
I think the boolean checks on the BaseActorModel could use a cleanup but I like the move to a clean iterator in AbilityModel

PR #209, opened 2025-01-26
Add modifiers to grabbed and restrained

### LINKED ISSUE #207: Grabbed and restrained need modifiers applied
Follow up to #198 

I forgot to add this part in the other PR. I'll add those in today or tomorrow

cdverrett94:
Closes #207 

Applying the modifiers for grabbed and restrained. Added `characteristic` to the `options` on the `rollCharacteristic` method.

I added an `isNotStatusSource` to the AE class. I couldn't just use `!isStatusSource` because it would trigger even if the actor didn't have the status effect. I thought about adding an optional fourth parameter to reverse the status source check, but that seemed like the return would be confusing if `isStatusSource` returned true when they weren't the source. If you have a better idea on handling that, let me know.

### DIFF
--- a/src/module/_types.d.ts
+++ b/src/module/_types.d.ts
@@ -41,5 +41,6 @@ export interface PowerRollPromptOptions {
   data: Record<string, unknown>;
   skills: string[];
   targets: PowerRollTargets[],
-  ability?: string
+  ability?: string,
+  characteristic?: string
 }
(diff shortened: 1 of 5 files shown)


JPMeehan:
*usually* the answer for a three-part state is `false`, `null`, and `true` - it's one of the reasons that _preCreate relies on an explicit return of false. Would that help?

cdverrett94:
> _usually_ the answer for a three-part state is `false`, `null`, and `true` - it's one of the reasons that _preCreate relies on an explicit return of false. Would that help?

Yeah, that helps. Thanks for the advice!

PR #212, opened 2025-01-30
Add field for number of kits for a class

### LINKED ISSUE #186: Add a number field for the number of kits per class
As of packet 4, classes can have 0, 1, or 2 kits. This seems like it should be defined at the Class level.

cdverrett94:
Closes #186 

Summary of Changes:
- Added the # of kits to the class schema.
- In the kit's `_preCreate`, prompt a dialog for choosing which kits to replace the incoming kit with if they are at the max amount of kits.

Screenshot 2025-01-29 213905

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -634,6 +634,9 @@
           },
           "recoveries": {
             "label": "Recoveries"
+          },
+          "kits": {
+            "label": "Kits"
           }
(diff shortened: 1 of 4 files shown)


JPMeehan (approved):
Great work, clean code.

PR #216, opened 2025-01-31
Refactoring the ability tier effects

cdverrett94:
So I was going through abilities in the hero and monster pdf and noticed some gaps in how we're handling ability effects. Some abilities have multiple damage entries with different types. Some allow choosing a damage type. Some damage has a potency entry. 

Here's kind of the options I saw as I went through them:

- Damage – type, amount, ignore immunities, type choice, potency, display text
  - Some abilities have multiple damages, sometimes one with a potency
- Forced Movement – type, distance, vertical, potency, display text
  - Option of Forced Movement – i.e. slide or pull
- Condition/AE – multiselect, potency, AEs to apply on potency success and AEs to apply on potency failure
  - Some abilities can have different potency characteristics (ashen hoarder, bredbeddle, high elf deathtouch)
- Option of Condition/AE – i.e. grabbed or prone
- Bespoke Effect (no condition or AE, effect as described)
- Combinations of the above

Notes: 
- I've updated the power roll schema to use `ArrayField(TypedSchemaField)`. 
  - The current types are `damage`, `ae` `forced` and `other`. `other` is for bespoke effects with "the target gains two surges" or "an ally within 10 can make a signature strike against the target as a free triggered action" as examples from the PDF. This could be expanded to more types in the future.
  - This will allow multiple damage, AE/condition, forced movement, or other effects. 
  - Each type has a potency schema and a display text schema. 
    - The display field will be what's shown in the ability embed.
    - I've made it so that values like `{{potency}}` and `{{damage}} cold damage` will be updated to the proper values like `M < 2` and `5 cold damage` respectively. This allows for these values to be shown in the embeds at the proper value. With damage, it allows kit damage to be shown in the embed.
    - For the potency data to be available I had to add a method that runs once actor data prep is finished.
- Because this allows a single damage effect to have multiple damage type options, I've updated the power roll dialog to include a damage type choice. I've assumed there won't be multiple tier effects that allow differing damage type options.
- I'm attaching a video of the updated ability power roll UI and a screenshot of what the updated embeds look like

https://github.com/user-attachments/assets/feb8ee7b-4304-4509-b75e-f6109be96ab4

Screenshot 2025-01-31 225706


templates/item/partials/ability.hbs
```diff
@@ -1,16 +1,45 @@
...
-  {{formGroup field.fields.damage.fields.type value=value.damage.type options=damageTypes blank=""}}
-  {{formGroup field.fields.ae value=value.ae options=appliedEffects}}
-  {{#if @root.system.powerRoll.potencyCharacteristic}}
-  {{formGroup field.fields.potency.fields.enabled value=value.potency.enabled}}
-  {{#if value.potency.enabled}}
-  {{formGroup field.fields.potency.fields.value value=value.potency.value placeholder=field.fields.potency.fields.value.initial}}
-  {{/if}}
-  {{/if}}
-  {{formGroup field.fields.description value=value.description}}
+  <span class="add-tier-effect" data-tier="{{field.name}}">Add Effect</span>
```
JPMeehan, line 12: i18n

templates/item/partials/ability.hbs
```diff
@@ -1,16 +1,45 @@
...
+  {{#each value as |effect|}}
+  <div class="power-roll-effect">
+    <span class="delete-tier-effect" data-tier="{{../field.name}}" data-index="{{@index}}">Delete Effect</span>
```
JPMeehan, line 15: i18n

templates/item/partials/ability.hbs
```diff
@@ -1,16 +1,45 @@
...
+    {{!-- effectFields will be equal to the fields under that effect type --}}
+    {{#with (lookup (lookup ../field.element.types effect.type) "fields") as |effectFields|}}
```
JPMeehan, line 17: This is a lot cleaner than I was expecting tbh

src/module/data/item/ability.mjs
```diff
@@ -54,32 +54,53 @@ export default class AbilityModel extends BaseItemModel {
...
-      }),
-      description: new fields.StringField({required: true})
+    const potencySchema = (initialPotency) => ({
+      enabled: new fields.BooleanField({label: "DRAW_STEEL.Item.Ability.FIELDS.powerRoll.tier.potency.enabled.label"}),
+      characteristic: new fields.StringField({label: "DRAW_STEEL.Item.Ability.FIELDS.powerRoll.tier.potency.characteristic.label"}),
+      value: new FormulaField({deterministic: true, initial: initialPotency, label: "DRAW_STEEL.Item.Ability.FIELDS.powerRoll.tier.potency.value.label"})
     });
 
+    const powerRollSchema = ({initialPotency}) => {
+      return new fields.TypedSchemaField({
```
JPMeehan, line 64: I wonder if you could do some kind of programmatic loop here - do `const schema = new fields.TypedSchemaField()` and then loop through `schema.fields` to create the label. Not sure if that's actually simpler than what you've done here.

JPMeehan (commented):
- I like the work so far.
- One other consideration I have is trying to figure out - for the Effect option - is an easy way to just apply a core status. Stuff like prone where you don't want to make a whole effect on the item for it and really we should be calling `ActiveEffect.fromStatusEffect` y'know

src/module/data/item/ability.mjs
cdverrett94: Not sure if it's simpler necessarily, but it at least pre-localizes it instead of happening every render.

templates/item/partials/ability.hbs
cdverrett94: That's fair. Maybe it's just me. But with not indenting for the handlebars helpers, I get easily lost trying navigate through stuff. And the long name fields just made it worse for me.

I'm not sure how you feel about the revised way I did it. I broke each effect type into an inline partial just to easily separate them. Then dynamically reference them in the fieldset.

cdverrett94:
I've added the tabs per tier. I still think the Details tab is really long. Maybe that's okay, I wonder if adding a separate tab after Details for configuring the Power Roll would be good.

I used the ability model's `getSheetContext` to add the subtab data as I wasn't sure where best to put it in the item sheet in the current setup. 

I know v13 is making changes to how tab data is set up, so maybe something that can be better improved upon then?

cdverrett94:
> * I like the work so far.
> * One other consideration I have is trying to figure out - for the Effect option - is an easy way to just apply a core status. Stuff like prone where you don't want to make a whole effect on the item for it and really we should be calling `ActiveEffect.fromStatusEffect` y'know

I can agree on the core statuses. I'm not sure what the best UI for adding those would be though because each status can have different save types or no save needed (prone). Possibly could have default values for each condition/save combo. That would make the list huge though since you'd have 3 versions of each.

JPMeehan:
[v13 will fix the localization issues](https://github.com/foundryvtt/foundryvtt/issues/11804)

JPMeehan:
> I can agree on the core statuses. I'm not sure what the best UI for adding those would be though because each status can have different save types or no save needed (prone). Possibly could have default values for each condition/save combo. That would make the list huge though since you'd have 3 versions of each.

Seems like it needs a helper app/dialog

cdverrett94:
I've updated the damage and ability embed to use the new structure. On the embed, instead of display the potency for each target, I've included it in the embed. To get the potency info into the embed, I did have to add a method that runs at the end of actor derived data prep to derive the values on the abilities. This is because the potency values aren't updated yet in ability data prep.

Screenshot 2025-01-31 225706

cdverrett94:
Alright, I think I've got this ready for review. Pretty sure I hit everything that would be affected by this change and it all seems to be working properly from what I can tell.

I've updated the top comment notes and screengrabs to show the updated state.

JPMeehan (changes requested):
I overall like what you've done here both in terms of correctness (wide variety of options for what Might Be Possible) as well as style (keeping clean code, UI looks good). Some notes though on areas I'd like refactored

src/module/data/actor/character.mjs
```diff
@@ -136,6 +136,9 @@ export default class CharacterModel extends BaseActorModel {
     this.potency.weak += highestCharacteristic - 2;
     this.potency.average += highestCharacteristic - 1;
     this.potency.strong += highestCharacteristic;
+
+    // prepare ability potency embeds after potency data is available on the actor
+    this.parent.items.filter(item => item.type === "ability").forEach(ability => ability.system.preparePostActorPrepData());
```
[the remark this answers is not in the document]
cdverrett94: Can you clarify what you mean? 

What I think you're saying is basically have the `BaseItemModel` have a `preparePostActorPrepData` method. Then instead of only looping for abilities, loop through all items?

src/module/data/item/ability.mjs
```diff
@@ -106,6 +146,28 @@ export default class AbilityModel extends BaseItemModel {
...
+  /**
+   * Prepare derived data available post actor data prep
+   */
+  preparePostActorPrepData() {
+    for (const tier of PowerRoll.TIER_NAMES) {
+      const effects = this.powerRoll[tier];
+      for (const effect of effects) {
+
+        // Replace {{damage}} with derived damage formula. Allows for showing damage with kit damage included
+        if ((effect.type === "damage") && effect.display.includes("{{damage}}")) effect.display = effect.display.replaceAll("{{damage}}", effect.value);
```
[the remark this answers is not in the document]
cdverrett94: Umm, I guess technically no? I guess I included it so that it wouldn't run the replaceAll needlessly for performance. But I don't think it would have an impact really. I'll remove them.

src/module/data/actor/character.mjs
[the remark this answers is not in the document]
JPMeehan: Yeah, cut out the `filter` and just do the for loop. (Apologies for the late reply, it's been busy)

src/module/data/item/ability.mjs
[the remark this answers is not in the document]
JPMeehan: The `includes` check already runs through the string so there's no performance gain.

JPMeehan:
Some longstanding PRs I had finally got looked at so you'll need to do some merge fixes, feel free to ask questions since I filed those ages ago and the design goals may have changed

src/module/data/actor/character.mjs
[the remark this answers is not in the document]
cdverrett94: I've done that now! I also moved the for loop to the base actor's `prepareDerivedData` instead so that even NPCs will benefit from this. The thought is that NPCs can use `{{potency}}` as well so that if there's a potency bonus, that will get tracked as well.

cdverrett94:
> Some longstanding PRs I had finally got looked at so you'll need to do some merge fixes, feel free to ask questions since I filed those ages ago and the design goals may have changed

I think I've cleared the conflicts, but while testing the DSN stuff wasn't seeming to have the effect I was thinking it would? I was thinking only one copy of the power roll (base roll) was supposed to be showing up instead of the base roll and one per target. It seems like it was rolling multiple still. I tested on the base branch as well and it was doing it there too.

JPMeehan:
> > Some longstanding PRs I had finally got looked at so you'll need to do some merge fixes, feel free to ask questions since I filed those ages ago and the design goals may have changed
> 
> I think I've cleared the conflicts, but while testing the DSN stuff wasn't seeming to have the effect I was thinking it would? I was thinking only one copy of the power roll (base roll) was supposed to be showing up instead of the base roll and one per target. It seems like it was rolling multiple still. I tested on the base branch as well and it was doing it there too.

It looks like I need to refine the logic, right now it works if "Merge Same message Rolls" is unchecked

JPMeehan (approved):
OK, took the time to fully go through this and make sure everything was right. Really great work here, I think we can start looking at an alpha release.

PR #219, opened 2025-02-10
Cleaned up possible ways the editor could get in a bad state

### LINKED ISSUE #204: Track down issue with editors
The editor element can screw with general rendering sometimes, possibly a permission issue related to the Collaborate thing?

JPMeehan:
Improved editor handling

Closes #204


krbz999 (approved):
Race conditions?

JPMeehan:
> Race conditions?

Nah, just the rendering handling on the description tab assumes the editor is still in the DOM. You can close the sheet to remove the editor from the DOM but it wouldn't clean up the sheet state.

Issue #218, opened 2025-02-10
Adjust display of Monster Metadata dialog


JPMeehan:
The relatively wide labels are handled poorly

![Image](https://github.com/user-attachments/assets/bf4b59d4-3fe2-4737-95b3-d2efa8b77aed)

PR #220, opened 2025-02-11
Adjusted sheet header css

### LINKED ISSUE #182: Improve handling for longer item names
Long item names wrap badly.

Image

JPMeehan:
Updated Sheet Header

image

Updated Monster metadata form

image

Closes #182 and #218

### DIFF
--- a/src/scss/components/_forms.scss
+++ b/src/scss/components/_forms.scss
@@ -19,11 +19,11 @@
   }
 
   .header-center {
+    height: 100%;
     .document-name {
       height: 60px;
+      line-height: 60px;
       input, h2 {
-        height: 100%;
-        line-height: 60px;
         font-family: var(--font-h2);
         font-size: var(--font-h2-size);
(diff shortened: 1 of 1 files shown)


src/module/apps/item-sheet.mjs
```diff
@@ -60,14 +60,14 @@ export class DrawSteelItemSheet extends api.HandlebarsApplicationMixin(
    * Available sheet modes.
    * @enum {number}
    */
-  static MODES = {
+  static MODES = Object.freeze({
```
krbz999, line 63: Seems we have some shared functionality between item and actor sheets. Mixin time?

krbz999 (approved):

src/module/apps/item-sheet.mjs
JPMeehan: Yeah, I think the v13 refactor is going to involve adding a mixin

PR #222, opened 2025-02-14
Update ability tab to present more information and use the ability embed

### LINKED ISSUE #202: Improve type & category handling for Villain Actions
Villain Actions are both a type *and* category, and the UI should cleanly handle that.

cdverrett94:
I wanted to update the ability section to have more info. I took inspiration from the 5e sheet.

- Closes #202, I added the villain action as an ability type. This is so that the NPC sheets will show a section for that type.
- I factored out the label formatting for keywords, distance, and targets so that those could be used on the actor sheet without having to pull in the entire ability sheet context.
- The embeds are only generated when the item would be expanded. Figured it was wasteful to generate a bunch of HTML that wasn't being shown.
- Should probably change the section header color to stand out more against the alternating item background color
- If this approach is acceptable, once approved, I'll make the same changes to the features tab

### DIFF
--- a/src/module/apps/actor-sheet/base.mjs
+++ b/src/module/apps/actor-sheet/base.mjs
@@ -296,6 +296,7 @@ export default class DrawSteelActorSheet extends api.HandlebarsApplicationMixin(
       };
     }
 
+    // Adding here instead of the initial context declaration so that the "other" category appears last on the character sheet
     context["other"] = {
       label: game.i18n.localize("DRAW_STEEL.Sheet.Other"),
       abilities: [],
--- a/src/module/apps/actor-sheet/base.mjs
+++ b/src/module/apps/actor-sheet/base.mjs
@@ -282,7 +282,7 @@ export default class DrawSteelActorSheet extends api.HandlebarsApplicationMixin(
    */
   async _prepareAbilitiesContext() {
     const context = {};
-    const abilities = this.actor.items.filter(i => i.type === "ability").sort((a, b) => a.sort - b.sort);
+    const abilities = this.actor.itemTypes.ability.toSorted((a, b) => a.sort - b.sort);
 
     // Prepare ability categories for each ability type
     for (const [type, config] of Object.entries(ds.CONFIG.abilities.types)) {
@@ -296,23 +296,32 @@ export default class DrawSteelActorSheet extends api.HandlebarsApplicationMixin(
       };
     }
 
+    context["other"] = {
+      label: game.i18n.localize("DRAW_STEEL.Sheet.Other"),
+      abilities: [],
+      fields: abilities[0].system.schema.fields
+    };
+
     // Prepare the context for each individual ability
     for (const ability of abilities) {
-      if (!context[ability.system.type]) continue;
+      const type = context[ability.system.type] ? ability.system.type : "other";
 
       const abilityContext = {
         ability,
         expanded: this.#expanded.has(ability.id),
         formattedLabels: ability.system.formattedLabels
       };
+
       // only get the embed data when it's relevant and expanded
       if (this.#expanded.has(ability.id)) abilityContext.embed = await ability.toEmbed({});
 
       // add the order to the villain action based on the current # of villain actions in the context
-      const villianActionCount = context[ability.system.type].abilities.length;
-      if (ability.system.type === "villain") abilityContext.order = villianActionCount + 1;
+      if (type === "villain") {
+        const villainActionCount = context[type].abilities.length;
+        abilityContext.order = villainActionCount + 1;
+      }
 
-      context[ability.system.type].abilities.push(abilityContext);
+      context[type].abilities.push(abilityContext);
     }
 
     return context;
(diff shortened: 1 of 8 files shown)


src/module/apps/actor-sheet/base.mjs
```diff
@@ -457,16 +476,20 @@ export default class DrawSteelActorSheet extends api.HandlebarsApplicationMixin(
...
    */
   static async _viewDoc(event, target) {
-    const doc = this._getEmbeddedDocument(target);
-    doc.sheet.render(true);
+    const item = this._getEmbeddedDocument(target);
+    if (!item) {
+      console.error("Could not find item");
+      return;
+    }
+    await item.sheet.render({force: true, mode: this.#mode});
```
JPMeehan, line 492: the doc can be an effect too, right? Passing `mode` here isn't directly a problem but should at least rename the variable.

src/module/apps/actor-sheet/base.mjs
```diff
@@ -555,6 +578,25 @@ export default class DrawSteelActorSheet extends api.HandlebarsApplicationMixin(
...
+   */
+  static async _toggleItemEmbed(event, target) {
+    const {itemId} = target.closest(".item").dataset;
+
+    if (this.#expanded.has(itemId)) this.#expanded.delete(itemId);
+    else this.#expanded.add(itemId);
+
+    const part = target.closest("[data-application-part]").dataset.applicationPart;
+    this.render({parts: [part]});
+  };
```
JPMeehan, line 597: Lint the tabbing here, this seems off

JPMeehan (commented):
I like the changes, can you elaborate what a fully configured villain action should look like here?

Action type primacy for the display I think is perfectly good, we'll still eventually want to improve filters for things like current heroic resources.

cdverrett94:
> I like the changes, can you elaborate what a fully configured villain action should look like here?
> 
> Action type primacy for the display I think is perfectly good, we'll still eventually want to improve filters for things like current heroic resources.

Currently this is what it looks like. I hadn't done any special display for it, but it is its own section. I'm thinking for that section, could possibly remove the cost heading and have an `Order` heading that can just pull the `@index` from the loop so that the number is just based on the sorting and not an actual thing on the model.

va-display

I also need to probably do some grouping in the item sheet prepare part context because right now the HTML is looping through each item per ability type which is unnecessary. Wanted to just get a demo working first.

src/module/apps/actor-sheet/base.mjs
cdverrett94: That's true. The update to this function was partially a leftover bit because originally I was going to have view/edit/delete buttons in the right-most column and this was an easy way to just have one function for both view/edit. Then I remembered you had the context menu and removed the buttons in the right column.

src/module/apps/actor-sheet/base.mjs
cdverrett94: Not a linting issue. Just me accidentally not removing a curly bracket without realizing it.

src/module/apps/actor-sheet/base.mjs
```diff
@@ -269,6 +276,48 @@ export default class DrawSteelActorSheet extends api.HandlebarsApplicationMixin(
...
+      context[type] = {
+        label: config.label,
+        abilities: [],
+        fields: abilities[0].system.schema.fields
+      };
+    }
+
+    // Prepare the context for each individual ability
+    for (const ability of abilities) {
+      if (!context[ability.system.type]) continue;
```
JPMeehan, line 301: We probably want some kind of "other" category for everything that's a miss, which would cover villain actions on PCs

src/module/apps/actor-sheet/base.mjs
```diff
@@ -269,6 +276,48 @@ export default class DrawSteelActorSheet extends api.HandlebarsApplicationMixin(
...
+        ability,
+        expanded: this.#expanded.has(ability.id),
+        formattedLabels: ability.system.formattedLabels
+      };
+      // only get the embed data when it's relevant and expanded
+      if (this.#expanded.has(ability.id)) abilityContext.embed = await ability.toEmbed({});
+
+      // add the order to the villain action based on the current # of villain actions in the context
+      const villianActionCount = context[ability.system.type].abilities.length;
+      if (ability.system.type === "villain") abilityContext.order = villianActionCount + 1;
```
JPMeehan, line 313: spelling
```suggestion
      const villainActionCount = context[ability.system.type].abilities.length;
      if (ability.system.type === "villain") abilityContext.order = villainActionCount + 1;
```

JPMeehan (changes requested):
Fairly minor adjustments

cdverrett94:
Okay, I've made the changes per review. Added the "other" category for missed abilities. The sheet will only show the "other" category if there are abilities in that category. That category also doesn't show the add ability icon either.

JPMeehan (changes requested):
Need to also test light mode - last two fixes

image

One fix to consider is see if there's any helpful css variables

cdverrett94:
light-mode

Here's an adjusted light mode display. None of the existing CSS variables seemed suitable, so I added 2 new ones. One for the header color and one for the alternating item bg color.

JPMeehan (approved):

PR #234, opened 2025-02-21
Fix characteristics not rolling

cdverrett94:
I swear I had fixed this last week. I thought I included in another PR, but I guess not. My b.


JPMeehan (approved):
Yeah not sure what happened here, maybe merge error?

JPMeehan:
Can you also update the versions in system.json to 0.6.2?

JPMeehan:
If we need to do more hotfixes we can branch off of main.

PR #237, opened 2025-02-21
Various V13 changes

cdverrett94:
These are some relatively minor, non-controversial changes.

- Update a couple instances of notications to use the new format option.
- Update a couple uses of `game.users.activeGM.isSelf` to `game.user.isActiveGM`
- Remove the `editImage` method from actor/item sheets
- Remove setting the active effect legacy transferral flag to false
- Remove the active effects `_applyUpgrade` override


JPMeehan (approved):
Will merge after we're otherwise in the clear

PR #244, opened 2025-02-23
Adjustments to the kit section of the features tab

cdverrett94:
- Show how many kits are allowed on the kit header so the player knows how many they can have without digging into the class info
- Only show the kit section if the class allows them or if there are currently kits on the actor to prevent them being inaccessible. 
- Update the kit swap dialog to be a little more clear on why you are having to replace a kit. 

updated kit section

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@
 
 ### Changed
 - Updated the view of the features tab to match the style of the abilities tab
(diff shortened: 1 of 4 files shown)


JPMeehan (approved):
Add changelog entry

cdverrett94:
Ahh shoot. I had already done but forgot to push the changes.

Issue #249, opened 2025-02-26
Add projects tab to character sheet


cdverrett94:
Add a projects tab to the character sheet that has roll buttons to easily roll and assign points to individual projects. Would show current progress as well.

PR #257, opened 2025-03-03
Add a project item

### LINKED ISSUE #248: Add new Project item type
Would allow tracking projects on the character sheet and automatically adding the items from crafting projects to the character

The necessary fields would be:
- item prereqs
- project source
- roll characteristic
- goal
- current project points - used for tracking on character sheets
- yields - used for automatically adding equipment once completed - schema field
   - item - item uuid
   - amount - formula field

cdverrett94:
Closes #248

- Adds a project item and its embed
   - If the initial creation data includes the item UUID, all of the item's project info is transferred to the project item and the name will be updated to "Craft {item name}"
   - The interactivity with characters will happen when it gets added to the character sheet as part of #249 
- Moved the equipment's item prerequisite to the project schema since the prereq is project related. If you need me to migrate this, I might need some help because I'm not sure how to do migrations.

project-item

### DIFF
--- a/draw-steel.mjs
+++ b/draw-steel.mjs
@@ -28,7 +28,7 @@ Hooks.once("init", function () {
   }
 
   helpers.registerHandlebars();
-  const templates = ["templates/item/embeds/ability.hbs", "templates/item/embeds/kit.hbs"].map(t => DS_CONST.systemPath(t));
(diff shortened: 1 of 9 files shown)


JPMeehan:
We're early enough I'm not terribly concerned about providing comprehensive migrations.

JPMeehan (approved):

cdverrett94:
Alright, power is finally back on, so was able to make the changes per request. Also updated the parameters of the goal/progress fields and added a progress input to the project sheet when its owned.

PR #271, opened 2025-03-06
Add DrawSteelActiveEffectConfig

### LINKED ISSUE #226: Update renderActiveEffectConfig
Currently, our custom active effect config is purely handled by a hook in `src/module/apps/hooks/activeEffectConfig.mjs`. A simple option is to update the hook, but a custom subclass would be more powerful going forward and is worth it now that Foundry has migrated the class.

cdverrett94:
Closes #226 

I left the core fields in for now even if they aren't useful, but if we want to remove them now, I can do that.


src/module/apps/_module.mjs
```diff
@@ -1,3 +1,4 @@
+export {DrawSteelActiveEffectConfig} from "./active-effect-config.mjs";
```
JPMeehan, line 1: This should go in the `/core` folder

src/module/apps/active-effect-config.mjs
```diff
@@ -0,0 +1,42 @@
+import {systemPath} from "../constants.mjs";
+
+const {sheets} = foundry.applications;
+
+/**
+ * The Application responsible for configuring a single ActiveEffect document within a parent Actor or Item.
+ */
+export class DrawSteelActiveEffectConfig extends sheets.ActiveEffectConfig {
+
+  static DEFAULT_OPTIONS = {
```
JPMeehan, line 10:
```suggestion
  /** @override */
  static DEFAULT_OPTIONS = {
```

src/module/apps/active-effect-config.mjs
```diff
@@ -0,0 +1,42 @@
...
+    classes: ["draw-steel"]
+  };
+
+  static PARTS = {
```
JPMeehan, line 14:
```suggestion
  /** @override */
  static PARTS = {
```

src/module/apps/active-effect-config.mjs
```diff
@@ -0,0 +1,42 @@
...
+    },
+    changes: {
+      template: "templates/sheets/active-effect-config/changes.hbs"
+    },
+    footer: {
+      template: "templates/generic/form-footer.hbs"
+    }
+  };
+
+  async _prepareContext(options) {
```
JPMeehan, line 35:
```suggestion
  /** @override */
  async _prepareContext(options) {
```

src/module/apps/active-effect-config.mjs
```diff
@@ -0,0 +1,42 @@
+import {systemPath} from "../constants.mjs";
+
+const {sheets} = foundry.applications;
```
JPMeehan, line 3: Would prefer to drop the dereferencing since we're only extending one class instead of class + mixin

JPMeehan (changes requested):
I also had to double check that core is actually just using that inline `#if tab.active`, messy but would rather not convolute the context prep.

src/module/apps/_module.mjs
cdverrett94: Didn't realize that was there.

src/module/apps/active-effect-config.mjs
cdverrett94: ooof

src/module/apps/active-effect-config.mjs
cdverrett94: Fair. I mostly did it because the line was pretty long already.

cdverrett94:
> I also had to double check that core is actually just using that inline `#if tab.active`, messy but would rather not convolute the context prep.

I changed this to just use `{{tab.cssClass}}` since its in the context already. Initially I just copy/pasted the core duration template and just added our fields without looking at that piece.

JPMeehan (approved):

PR #272, opened 2025-03-06
Update ActorSheet ContextMenu to use v13 standards

cdverrett94:
Also added the star icon to the preferred kit button.

The ability melee/ranged options should probably use the pro sword/bow icons, but that's something you would have to do.

### DIFF
--- a/src/module/apps/actor-sheet/base.mjs
+++ b/src/module/apps/actor-sheet/base.mjs
@@ -411,8 +411,7 @@ export default class DrawSteelActorSheet extends api.HandlebarsApplicationMixin(
    */
   async _onFirstRender(context, options) {
     await super._onFirstRender(context, options);
-    // TODO: Change to ContextMenu.create in v13 with jQuery: false
-    new ContextMenu(this.element, "[data-document-class]", this._getItemButtonContextOptions());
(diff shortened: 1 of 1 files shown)


src/module/apps/actor-sheet/base.mjs
```diff
@@ -411,8 +411,7 @@ export default class DrawSteelActorSheet extends api.HandlebarsApplicationMixin(
    */
   async _onFirstRender(context, options) {
     await super._onFirstRender(context, options);
-    // TODO: Change to ContextMenu.create in v13 with jQuery: false
-    new ContextMenu(this.element, "[data-document-class]", this._getItemButtonContextOptions());
+    foundry.applications.ui.ContextMenu.create(this, this.element, "[data-document-class]", {hookName: "ItemButtonContext", jQuery: false, fixed: true});
```
JPMeehan, line 414: Kinda ugly and surprised this didn't get broken up for length but neither of those are your fault, just an artifact of how this is constructed now with the triply nested class + static method + lengthy list of options.

JPMeehan (approved):
I'm pleased that we barely had to adjust the helper function

PR #281, opened 2025-03-08
Add skills to power roll dialog and include +2 bonus when skill selected

### LINKED ISSUE #277: Include skill selection in test roll `PowerRollDialog`s
Relies on #276

cdverrett94:
Closes #277 

- Updated the `rollCharacteristic` method to return a `DrawSteelChatMessage`. The previous implementation was using "message" as the evaluation method which meant that the create message didn't have the base roll in it. Updating to evaluate and creating the message in `rollCharacteristic` means that the base roll is showing in the test chat messages too
- v13 made expanding the roll result a `data-action` so added that to our template to restore that functionality.

test-roll

### DIFF
--- a/src/module/apps/power-roll-dialog.mjs
+++ b/src/module/apps/power-roll-dialog.mjs
@@ -50,7 +50,7 @@ export class PowerRollDialog extends HandlebarsApplicationMixin(ApplicationV2) {
 
     if (context.targets) await this._prepareTargets(context);
 
-    if (context.skills && (context.skills.size > 0)) this._prepareSkillOptions(context);
+    if (context.skills?.size > 0) this._prepareSkillOptions(context);
 
     return context;
   }
--- a/src/module/apps/power-roll-dialog.mjs
+++ b/src/module/apps/power-roll-dialog.mjs
@@ -50,6 +50,8 @@ export class PowerRollDialog extends HandlebarsApplicationMixin(ApplicationV2) {
 
     if (context.targets) await this._prepareTargets(context);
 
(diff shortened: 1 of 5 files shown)


src/module/apps/power-roll-dialog.mjs
```diff
@@ -50,6 +50,8 @@ export class PowerRollDialog extends HandlebarsApplicationMixin(ApplicationV2) {
 
     if (context.targets) await this._prepareTargets(context);
 
+    if (context.skills && (context.skills.size > 0)) this._prepareSkillOptions(context);
```
JPMeehan, line 53: can just use optional chaining since undefined > 0 is false
```suggestion
    if (context.skills?.size > 0) this._prepareSkillOptions(context);
```

src/module/data/actor/base.mjs
```diff
@@ -240,10 +243,23 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
...
+    const modifiers = {
+      edges: options.edges ?? 0,
+      banes: options.banes ?? 0,
+      bonuses: options.bonuses ?? 0
+    };
+
+    const rolls = await PowerRoll.prompt({type, evaluation, formula, data, flavor, modifiers, actor: this.parent, characteristic, skills});
+    return DrawSteelChatMessage.create({
+      speaker: DrawSteelChatMessage.getSpeaker({actor: this.parent}),
+      rolls
```
JPMeehan, line 261: I believe we also need to pass `sound: CONFIG.Dice.sound` here or something like that

JPMeehan (approved):
Looks good thanks for tackling this

PR #282, opened 2025-03-08
Token Ruler labels and Free Strikes

krbz999:
Adds `TokenDocument#getHostileTokensFromPoints` which returns token documents (different from the token) that are within 1 grid space of any of the given points.

Adjusts the label when actively measuring to display the number of free strikes that might happen.

Adds system subclasses of `TokenRuler`, `Token`, and `TokenDocument`.

TODO:
- [ ] Take shifting into account.
- [ ] Adjust ruler labels when the token is actively moving (user is done measuring).
- [ ] Adjust ruler labels for the movement history.
- [ ] Add a prompt (?) and/or apply damage to the moving token's actor if there are any free strikes.
- [x] The changes to the label should not apply to "passed" segments.
- [x] Increase strike count at each waypoint.

I'm not sold on the above way of handling free strikes; might also segmentize movement for each point where the token would move into range of an enemy, pause it, resolve the damage, and continue.


JPMeehan:
_does_ data/canvas make sense as a folder structure? Feels like it should just directly be `canvas`.

krbz999:
> _does_ data/canvas make sense as a folder structure? Feels like it should just directly be `canvas`.

Not really, but most of everything else was there already. I've personally started mimicking Core's structure in all module exports and replicating it nearly 1-to-1 with the folder structure.

JPMeehan:
Part of the reason I'm asking is I would not be surprised if we end up also doing custom measured templates. I agree on mirroring core's structure in the exports, we're fortunately unburdened by migrations.

src/module/documents/_module.mjs
```diff
@@ -6,4 +6,5 @@ export {DrawSteelChatMessage} from "./chat-message.mjs";
 export {DrawSteelCombat} from "./combat.mjs";
 export {DrawSteelCombatant} from "./combatant.mjs";
 export {DrawSteelItem} from "./item.mjs";
+export {default as DrawSteelTokenDocument} from "./token.mjs";
```
JPMeehan, line 9: Should we flip all the Documents to `export default`?

krbz999: I prefer it, but not in this PR.

krbz999:
I think this works well enough to get reviewed. The remaining points can be follow-ups.

src/module/documents/token.mjs
```diff
@@ -0,0 +1,60 @@
...
+  /* -------------------------------------------------- */
+
+  /**
+   * Get hostile tokens within range of movement.
+   * @param {Point[]} [points]                An array of points describing a segment of movement.
+   * @param {object} [options]
+   * @param {boolean} [options.ignoreFirst]   Ignore any tokens that were in range of the first point.
+   * @returns {DrawSteelTokenDocument[]}      Hostile tokens.
+   */
+  getHostileTokensFromPoints(points = [], {ignoreFirst = false} = {}) {
```
JPMeehan, line 19: I think some of the logic here is flipped - I'm getting a strike count for *approaching* a target but not for directly retreating.

JPMeehan (changes requested):
Really nice work, would like to adjust the retreat/engage logic
image

image

draw-steel.mjs
```diff
@@ -39,6 +41,10 @@ Hooks.once("init", function () {
     }
   }
 
+  // Assign canvas-related classes
+  CONFIG.Token.objectClass = canvas.placeables.DrawSteelToken;
+  CONFIG.Token.rulerClass = canvas.placeables.tokens.DrawSteelTokenRuler;
```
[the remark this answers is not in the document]
krbz999: I don't think there's much to be gained from trying to configure it in a way that `Object.assign` would be possible.

src/module/documents/token.mjs
krbz999: The logic should perhaps be tokens you pass by (as it is now) but whom you are not within 1 square of at the path's end point.

krbz999: What do the rules state happens if you move in and out of range of an enemy repeatedly on a single turn?

JPMeehan: Looks like you could take multiple free strikes if you move away, back in, then away again.

krbz999:
By golly I've done it.

image

JPMeehan (approved):

PR #284, opened 2025-03-09
Add roll mode buttons to power roll dialog

### LINKED ISSUE #143: Add roll mode configuration to power roll prompt
As a user, I should be able to choose the roll mode for a roll without adjusting the setting in the chat box

cdverrett94:
Closes #143 

- The dialog defaults to using the users' core roll mode setting
roll-mode-buttons

### DIFF
--- a/src/module/apps/power-roll-dialog.mjs
+++ b/src/module/apps/power-roll-dialog.mjs
@@ -171,7 +171,7 @@ export class PowerRollDialog extends HandlebarsApplicationMixin(ApplicationV2) {
    */
   static setRollMode(event, target) {
     this.options.context.rollMode = target.dataset.rollMode;
-    this.render(true);
+    this.render();
   }
 
   /* -------------------------------------------- */
--- a/src/module/apps/power-roll-dialog.mjs
+++ b/src/module/apps/power-roll-dialog.mjs
@@ -17,6 +17,9 @@ export class PowerRollDialog extends HandlebarsApplicationMixin(ApplicationV2) {
     position: {
       width: 400
     },
+    actions: {
+      setRollMode: this.setRollMode
+    },
     tag: "form",
     form: {
       closeOnSubmit: true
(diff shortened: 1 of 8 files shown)


src/module/apps/power-roll-dialog.mjs
```diff
@@ -148,9 +159,21 @@ export class PowerRollDialog extends HandlebarsApplicationMixin(ApplicationV2) {
...
   }
 
+  /**
+  * @this DrawSteelItemSheet
+   * @param {PointerEvent} event   The originating click event
+   * @param {HTMLElement} target   The capturing HTML element which defined a [data-action]
+   */
+  static setRollMode(event, target) {
+    this.options.context.rollMode = target.dataset.rollMode;
+    this.render(true);
```
JPMeehan, line 174: This should just be a normal re-render?
```suggestion
    this.render();
```

src/module/rolls/power.mjs
```diff
@@ -129,7 +129,7 @@ export class PowerRoll extends DSRoll {
   /**
    * Prompt the user with a roll configuration dialog
    * @param {Partial<PowerRollPromptOptions>} [options] Options for the dialog
-   * @return {Promise<Array<PowerRoll | DrawSteelChatMessage | object>>} Based on evaluation made can either return an array of power rolls or chat messages
+   * @return {Promise<{rollMode: keyof typeof CONFIG["Dice"]["rollModes"], powerRolls: Array<PowerRoll | DrawSteelChatMessage | object>}>} Based on evaluation made can either return an array of power rolls or chat messages
```
JPMeehan, line 132: Should this be factored out to an interface?

JPMeehan (approved):
Nits; I really like the re-use of the core buttons.

src/module/apps/power-roll-dialog.mjs
cdverrett94: Fair. This made me think about something. It might be a good idea to separate the different sections into their own parts (global mods, targets' mods, skill selection, footer). This would allow for specific part re-rendering and part specific context prep. Not for this PR though, but this made me think about it.

JPMeehan: If it grows any more in complexity probably?

cdverrett94: The growing complexity is what made me think about it. If we don't feel its necessary now, that's fine.

Issue #233, opened 2025-02-21
Update Actor & Item Sheets to take advantage of v13 updates


JPMeehan:
- Remove _onEditImage
- Refactor tabs handling
- Remove dragdrop handling from ActorSheet
- Adjust context prep to call `super` since we can inherit the base fields and document.

cdverrett94:
Commenting for assignment

cdverrett94:
Was there anything on this one that needed to be done?

JPMeehan:
Yeah I think we've got everything here done.

PR #285, opened 2025-03-09
Update actor sheet tabs to v13 method

cdverrett94:
Related to #233 

Also updated the actor sheet to call `super._prepareContext` and removed the duplicate entries for the actor sheet.

### DIFF
--- a/src/module/apps/actor-sheet/base.mjs
+++ b/src/module/apps/actor-sheet/base.mjs
@@ -39,6 +39,22 @@ export default class DrawSteelActorSheet extends api.HandlebarsApplicationMixin(
     }
   };
 
+  /** @override */
+  static TABS = {
+    primary: {
+      tabs: [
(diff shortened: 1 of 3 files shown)


src/module/apps/actor-sheet/base.mjs
```diff
@@ -90,10 +106,9 @@ export default class DrawSteelActorSheet extends api.HandlebarsApplicationMixin(
 
   /** @override */
   async _prepareContext(options) {
-    const context = {
+    const context = foundry.utils.mergeObject(await super._prepareContext(options), {
```
JPMeehan, line 109: I think `Object.assign` is safer/better here due to the way we're accessing static properties, we don't need deep merging.
```suggestion
    const context = Object.assign(await super._prepareContext(options), {
```

JPMeehan (approved):

PR #288, opened 2025-03-11
Update token config hooks

### LINKED ISSUE #232: Update renderTokenConfig
I think this works fine as a hook, there's not really a clean way to replicate the current functionality with subclassing the app 
- we'd have to subclass *both* PrototypeTokenConfig *and* TokenConfig (probably with a mixin for efficiency)
- Adjusting the prepared context doesn't seem to be a clear win in efficiency compared to the current hook-based setup

Given this, I think the simple option is to just adjust our hook call to the new setup (which is why this is only a Task instead of a Feature)

cdverrett94:
Closes #232

- I renamed the hook file to match the new hook name.
- I think I did the typing stuff correctly. I took the example from the combatant config hook

### DIFF
--- a/src/module/apps/hooks/tokenApplication.mjs
+++ b/src/module/apps/hooks/tokenApplication.mjs
@@ -1,7 +1,7 @@
 /** @import {DrawSteelActor} from "../../documents/actor.mjs" */
 
 /** @typedef {InstanceType<foundry["applications"]["sheets"]["PrototypeTokenConfig"]>} PrototypeTokenConfig */
-/** @typedef {InstanceType<foundry["applications"]["sheets"]["TokenConfig"]>} TokenConfig*/
+/** @typedef {InstanceType<foundry["applications"]["sheets"]["TokenConfig"]>} TokenConfig */
 
 /**
  * A hook event that fires when the TokenConfig application is rendered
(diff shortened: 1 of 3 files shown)


src/module/apps/hooks/tokenApplication.mjs
```diff
@@ -1,19 +1,23 @@
...
+ * @param {PrototypeTokenConfig | TokenConfig} app             The Application instance being rendered
+ * @param {HTMLElement>} html  The inner HTML of the document that will be displayed and may be modified
  * @param {Record<string, any>} context The object of data used when rendering the application
+ * @param {ApplicationRenderOptions} options
  */
-export function renderTokenConfig(app, [html], context) {
+export function renderTokenApplication(app, html, context, options) {
   /** @type {DrawSteelActor} */
-  const actor = app.document.actor;
+  const actor = app.actor;
```
JPMeehan, line 15: Does this throw an error if you open the Default Token Config? Just want to make sure it quietly exits.

JPMeehan (approved):

src/module/apps/hooks/tokenApplication.mjs
cdverrett94: Looks like the Default Token Config was replaced Prototype Overrides and doesn't use the same hook

PR #289, opened 2025-03-11
Update Actor Sheet dragDrop handling for v13

cdverrett94:
Related to #233. I _think_ this is the last piece from that issue.

- Core didn't have a sort active effect method or an implementation of folder dropping, so i left ours in there.
- The methods in core have a return of `Promise<void>`, so I updated the remaining of ours to match that.

### DIFF
--- a/src/module/apps/actor-sheet/base.mjs
+++ b/src/module/apps/actor-sheet/base.mjs
@@ -354,7 +354,6 @@ export default class DrawSteelActorSheet extends api.HandlebarsApplicationMixin(
 
   /**
    * Actions performed after a first render of the Application.
-   * Post-render steps are not awaited by the render process.
    * @param {ApplicationRenderContext} context      Prepared context data
    * @param {RenderOptions} options                 Provided render options
    * @protected
@@ -482,7 +481,6 @@ export default class DrawSteelActorSheet extends api.HandlebarsApplicationMixin(
 
   /**
    * Actions performed after any render of the Application.
-   * Post-render steps are not awaited by the render process.
    * @param {ApplicationRenderContext} context      Prepared context data
    * @param {RenderOptions} options                 Provided render options
    * @protected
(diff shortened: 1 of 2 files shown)


src/module/apps/actor-sheet/base.mjs
```diff
@@ -490,8 +488,8 @@ export default class DrawSteelActorSheet extends api.HandlebarsApplicationMixin(
    * @protected
    * @override
    */
-  _onRender(context, options) {
-    this.#dragDrop.forEach((d) => d.bind(this.element));
+  async _onRender(context, options) {
```
JPMeehan, line 491: I think the comments might need updates as well since they now await several of these functions

src/module/apps/actor-sheet/base.mjs
```diff
@@ -662,97 +660,21 @@ export default class DrawSteelActorSheet extends api.HandlebarsApplicationMixin(
...
-    }
-  }
-
-  /**
-   * Handle the dropping of ActiveEffect data onto an Actor Sheet
-   * @param {DragEvent} event                  The concluding DragEvent which contains drop data
-   * @param {object} data                      The data transfer extracted from the event
-   * @returns {Promise<ActiveEffect|boolean>}  The created ActiveEffect object or false if it couldn't be created.
-   * @protected
+   * Core doesn't sort active effects, so if target is this actor, sort the effects
```
JPMeehan, line 663: I believe this comment overrides the inherited comment in intellisence, so we should either tag this with `@remarks` or copy the core method annotation.

JPMeehan (approved):
Looks good, nice that we need a lot less Boilerplate here.

src/module/apps/actor-sheet/base.mjs
cdverrett94: I updated a few of the comments to remove the bit about not being awaited.

src/module/apps/actor-sheet/base.mjs
cdverrett94: I'll just copy the core annotation. My comment wasn't really required to understand what was happening.

PR #293, opened 2025-03-11
Add immunities/weaknesses to actor stats tab

### LINKED ISSUE #184: Add sheet section for immunities and weaknesses

cdverrett94:
Closes #184 

Edit Mode
edit-iw

Play Mode
play-iw

### DIFF
--- a/src/module/apps/actor-sheet/base.mjs
+++ b/src/module/apps/actor-sheet/base.mjs
@@ -218,7 +218,7 @@ export default class DrawSteelActorSheet extends api.HandlebarsApplicationMixin(
   }
 
   /**
-   * Constructs an object with the actor's languages as well as all options available from CONFIG.DRAW_STEEL.languages
+   * Constructs an object with the formatted immunities and weaknesses with a list of damage labels
    * @returns {{immunities: string, weaknesses: string, labels: Record<string, string>}}
    */
   _getImmunitiesWeaknesses() {
--- a/src/module/apps/actor-sheet/base.mjs
+++ b/src/module/apps/actor-sheet/base.mjs
@@ -133,7 +133,7 @@ export default class DrawSteelActorSheet extends api.HandlebarsApplicationMixin(
       case "stats":
         context.characteristics = this._getCharacteristics();
         context.movement = this._getMovement();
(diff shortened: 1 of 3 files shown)


src/module/apps/actor-sheet/base.mjs
```diff
@@ -133,6 +133,7 @@ export default class DrawSteelActorSheet extends api.HandlebarsApplicationMixin(
       case "stats":
         context.characteristics = this._getCharacteristics();
         context.movement = this._getMovement();
+        context.damage = this._getImmunitiesWeaknesses("immunities");
```
JPMeehan, line 136: Should be a more specific property name and it doesn't look like you still take an argument

templates/actor/character/stats.hbs
```diff
@@ -155,6 +155,45 @@
     {{formGroup systemFields.movement.fields.teleport value=systemSource.movement.teleport rootId=actor.uuid dataset=datasets.isSource}}
   </fieldset>
   {{/if}}
+  {{#if isPlay}}
+  <div class="immunities-weaknesses flexrow">
+    <fieldset>
+      <legend>{{systemFields.damage.fields.immunities.label}}</legend>
+      <span class="play">
+        {{#each damage.immunities as |immunity|}}
+        <span class="immunity">{{lookup @root.damage.labels @key}} {{immunity}}</span>{{#unless @last}}, {{/unless}}
```
JPMeehan, line 164: We should be using i18n formatters for this, which means moving the whole string to context prep

JPMeehan (changes requested):

src/module/apps/actor-sheet/base.mjs
cdverrett94: Oops

templates/actor/character/stats.hbs
cdverrett94: Fair, I had it that way before without wrapping each item in a `span`, but noticed that damage type and value could be on separate lines which didn't look great. I didn't really think to just include the `span`s in the context prep.

JPMeehan (approved):

Issue #93, opened 2024-12-26
Evaluate if Actor Sheet or Item Sheet should be using the `standard-form` class


JPMeehan:
Foundry's `standard-form` class provides a lot of built-in styling that may or may not be desirable on the Actor and/or Item Sheet

This should be evaluated as part of an overall "make the sheets look nice" pass

JPMeehan:
Item sheet example

With standard-form
![image](https://github.com/user-attachments/assets/a5fad81e-5117-4a08-bde2-743b251d6aa5)

Without standard-form
![image](https://github.com/user-attachments/assets/4109ce06-bd0a-48d2-9384-d688f54adca3)

cdverrett94:
Do we want to make the changes to add the standard form to item sheets? 

In V13 without standard form:
![Image](https://github.com/user-attachments/assets/ca94c11d-9a49-4fce-8c1a-b2ed09493a1c)

In V13 with standard form:
![Image](https://github.com/user-attachments/assets/9c916c0d-dfd7-43be-96cd-32c2f15ea9fc)

JPMeehan:
I think it's worth it for now, we should probably add `stacked` to any string inputs we expect to be long

PR #300, opened 2025-03-13
Add a project tab with project rolls and item creation

### LINKED ISSUE #249: Add projects tab to character sheet

cdverrett94:
Closes #249 

- Add project tab to the character sheet
- When an equipment item either singularly or in a folder is dropped onto the project tab, create the item as a project instead of as an equipment
   - Added a helper method to the `EquipmentModel` to create as a project on the actor
- Add a roll method to the `ProjectModel` to roll the project roll and add the roll to the progress.
   - If there is a yielded item and the progress beats the goal, create a number of them based on the amount formula
- Update various types
- Update some actor sheet stuff that I noticed while making the project changes
   - Update actor sheet part config to v13 method
   - Update actor sheet tabs to only include biography in limited mode and to remove the equipment/project tabs for non-characters
 
project-tab


src/module/data/item/project.mjs
```diff
@@ -98,4 +117,65 @@ export default class ProjectModel extends BaseItemModel {
...
+    const {rollMode, projectRoll} = promptValue;
+
+    const total = projectRoll.total;
+    const updatedProgress = this.progress + total;
+    await this.parent.update({"system.progress": updatedProgress});
+
+    // If the project has been completed and there is a yield item, notify the user.
+    // If there is a yielded item, roll the amount formula and add that many of the item.
+    if (updatedProgress >= this.goal) {
+      ui.notifications.info("DRAW_STEEL.Item.Project.CompletedNotification", {format: {
```
JPMeehan, line 151: Use the new `success` notification type
```suggestion
      ui.notifications.success("DRAW_STEEL.Item.Project.CompletedNotification", {format: {
```

src/module/data/item/project.mjs
```diff
@@ -98,4 +117,65 @@ export default class ProjectModel extends BaseItemModel {
...
+      }});
+
+      if (this.yield.item) {
+        const item = await fromUuid(this.yield.item);
+        const yieldRoll = await new DSRoll(this.yield.amount).evaluate();
+        const amount = yieldRoll.total;
+        const itemArray = Array(amount).fill(item.toObject());
+
+        await this.actor.createEmbeddedDocuments("Item", itemArray);
+        ui.notifications.info("DRAW_STEEL.Item.Project.Craft.CompletedNotification", {format: {
```
JPMeehan, line 163: Same here
```suggestion
        ui.notifications.success("DRAW_STEEL.Item.Project.Craft.CompletedNotification", {format: {
```

JPMeehan (commented):
Mobile review will give more thorough later, but looks good at first glance

cdverrett94:
> Mobile review will give more thorough later, but looks good at first glance

Didn't realize the success notification was a thing now.

Something a little semantic. Should update the schema from `progress` to `points`? I just realized when re-reading the project section, that the PDF uses the `project points` as the wording.

JPMeehan:
> > Mobile review will give more thorough later, but looks good at first glance
> 
> Didn't realize the success notification was a thing now.
> 
> Something a little semantic. Should update the schema from `progress` to `points`? I just realized when re-reading the project section, that the PDF uses the `project points` as the wording.

I think that's reasonable

templates/item/partials/project.hbs
```diff
@@ -3,12 +3,13 @@
 {{else}}
 <fieldset>
```
JPMeehan, line 4: I'm not sure what the fieldset is really accomplishing in this case, and the whole display is kinda cramped. I think the stacked fields works for some but not all of these (e.g. the number fields don't need the full width), and definitely need some spacing between the inputs and the next label.
image

(I realize that this isn't *new* to this PR but I overlooked this previously and we're addressing these items now)

JPMeehan (changes requested):
OK, more in depth review this time

templates/item/partials/project.hbs
cdverrett94: Using a fieldset for the whole thing, was just a misunderstanding/misreading of some of the other templates. I also thought you needed a singular outer element for it to render. I swear that used to be the case.

As for the styling of the elements and the elements being stacked, that's actually nothing I've done. That's a core CSS change. You can look at all of the item sheets details. I think if we did #93 and added the standard-form it would display properly. 

standard-form

JPMeehan: The requirement for a single outer element is for the top-level PARTS in HandlebarsApplicationMixin; our details partials are inside them.

Sounds good on the styling, we can tackle that in a future PR.

JPMeehan (approved):

templates/item/partials/project.hbs
cdverrett94: Ahhh, right. Thanks for the clarification!

PR #318, opened 2025-03-17
Update hotbar item dropping

### LINKED ISSUE #313: Change ability hotbar drop to using the ability
As a user, when I drop an ability on the hotbar, I would like the macro to use that ability rather than open its item sheet.

cdverrett94:
Closes #313 

- For projects/abilities, creates a roll macro, but for other item types it creates the default hotbar's toggle sheet macro.

### DIFF
--- a/src/module/helpers/macros.mjs
+++ b/src/module/helpers/macros.mjs
@@ -8,7 +8,7 @@
  */
 export async function createDocMacro(data, slot) {
   // First, determine if this is a valid owned item.
-  if (!data.uuid.includes("Actor.") && !data.uuid.includes("Token.")) {
+  if (!data.uuid.includes("Actor.")) {
     return ui.notifications.warn("DRAW_STEEL.Macro.Warnings.Create.NotOwnedItem", { localize: true });
   }
   // If it is, retrieve it based on the uuid.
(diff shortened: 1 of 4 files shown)


src/module/helpers/macros.mjs
```diff
@@ -8,49 +8,56 @@
  */
 export async function createDocMacro(data, slot) {
   // First, determine if this is a valid owned item.
-  if (data.type !== "Item") return;
   if (!data.uuid.includes("Actor.") && !data.uuid.includes("Token.")) {
```
JPMeehan, line 11: I'm pretty sure items even in a token delta still list the Actor in their uuid?

JPMeehan (approved):
Some considerations

src/module/helpers/macros.mjs
cdverrett94: Yeah, looks like that's right. I just had left that check as is. I'll update to just check `Actor.`

src/module/helpers/macros.mjs
```diff
@@ -8,49 +8,56 @@
...
+  const item = await fromUuid(itemUuid);
+  if (!item) return ui.notifications.warn("DRAW_STEEL.Macro.Warnings.Roll.NoItem", { localize: true });
+  if (!item.parent) return ui.notifications.warn("DRAW_STEEL.Macro.Warnings.Roll.NotOwnedItem", { format: { item: item.name } });
 
-    // Trigger the item roll
-    item.roll();
-  });
+  switch (item.type) {
+    case "ability":
+      item.system.use();
```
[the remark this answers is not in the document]
JPMeehan: I think the best option is to use `{@see AbilityModel#use}` and just expect people to look for the original rather than the alias.

krbz999:
Generally the concept of "rolling" an item I believe to be an outdated 5e-ism that also spilled over into other simpler systems back in the day.

Are we opposed to using `use` or `execute`?

JPMeehan:
> Generally the concept of "rolling" an item I believe to be an outdated 5e-ism that also spilled over into other simpler systems back in the day.
> 
> Are we opposed to using `use` or `execute`?

We have "use" for Abilities as our default, for Projects the "roll" is because it's the downtime activity to roll progress

JPMeehan:
> I've updated to use the `AbilityModel#roll` alias, but let me know if there's a different way you want to align those methods.

Yeah, I think this works for me now.

PR #319, opened 2025-03-18
Fixed inheritDoc, context icons

JPMeehan:
- Unified `inheritdoc` over `inheritDoc`
- Re-added <i> tags to context menu option constructors

### DIFF
--- a/src/module/applications/sheets/actor-sheet.mjs
+++ b/src/module/applications/sheets/actor-sheet.mjs
@@ -426,7 +426,7 @@ export default class DrawSteelActorSheet extends api.HandlebarsApplicationMixin(
       //Ability specific options
       {
         name: "DRAW_STEEL.Item.Ability.SwapUsage.ToMelee",
-        icon: "fa-solid fa-sword",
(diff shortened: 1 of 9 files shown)


src/module/applications/sheets/actor-sheet.mjs
```diff
@@ -426,7 +426,7 @@ export default class DrawSteelActorSheet extends api.HandlebarsApplicationMixin(
       //Ability specific options
       {
         name: "DRAW_STEEL.Item.Ability.SwapUsage.ToMelee",
-        icon: "fa-solid fa-sword",
+        icon: "<i class=\"fa-solid fa-sword\"></i>",
```
krbz999, line 429: Should add the `fa-fw` class here too.

krbz999 (approved):

src/module/applications/sheets/actor-sheet.mjs
krbz999: preferably a second one unless its inconvenient

PR #338, opened 2025-03-21
Refactor actor sheets to use partials

### LINKED ISSUE #336: Add immunities and weaknesses to NPC sheet
Follow up to #293 & #184

cdverrett94:
Closes #336

### DIFF
--- a/src/module/applications/sheets/character.mjs
+++ b/src/module/applications/sheets/character.mjs
@@ -51,6 +51,7 @@ export default class DrawSteelCharacterSheet extends DrawSteelActorSheet {
     },
     biography: {
       template: systemPath("templates/actor/character/biography.hbs"),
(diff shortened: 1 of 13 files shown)


JPMeehan:
> At some point, should the common sections between NPC and character become individual partials so that the HTML is duplicated?

Yes, if you want to expand this PR to do that I think that would be reasonable

cdverrett94:
Yeah, I can do that. Should these go under the shared folder or should I add a partials folder to shared?

JPMeehan:
shared/partials I think

cdverrett94:
Okay, moved all of the common stats sections to use partials.

I used shared/partials/stats as the folder because I can see the biography tab also needed to do this since has a few common sections as well.

Do you want me to do the biography stuff to partials as part of this as well?

JPMeehan (approved):

cdverrett94:
Alright, biography has been updated!

PR #340, opened 2025-03-21
Update from scss to css files

### LINKED ISSUE #253: Consider switching from scss to pure css
With [Foundry upgrading to Electron 34.2](https://github.com/foundryvtt/foundryvtt/issues/12236) there are increasingly few reasons to continue to use scss, as now the native app cleanly supports element nesting. This has the benefit of reducing the number of install steps for contributors or anyone else who directly loads up the repo.

cdverrett94:
Closes #253 

- I've left the SCSS files in here for now. I also am including the old compiled `draw-steel.css` file as `old-draw-steel.css` just so its there.
   - I can remove both of these before merging once we're satisfied.
- As far as the folder structure, I've made the top level folders match the layer they are on.
   - For the applications folder, I tried to match the `module/applications` structure.
- For the mixin file, I didn't see the two mixins used anywhere, so I didn't include anything for them in the updated CSS
- Going through, everything looks fine. The only issue I noticed due to the layers structure was that the item description/gm notes wasn't hiding the non-active editor. This was due to the application layer have lower priority than the compatibility layer. I fixed it by using `!important`.


package.json
```diff
@@ -7,7 +7,7 @@
     "pullJSONtoLDB": "node ./tools/pullJSONtoLDB.mjs",
     "pushLDBtoJSON": "node ./tools/pushLDBtoJSON.mjs",
     "unpack": "npm run pushLDBtoJSON",
-    "build": "npm run build:packs && npm run build:css && npm run build:code",
+    "build": "npm run build:packs && npm run build:code",
     "build:packs": "npm run pullJSONtoLDB",
     "build:css": "sass src/scss/draw-steel.scss css/draw-steel.css --style=expanded --no-source-map",
     "build:css:watch": "sass src/scss/draw-steel.scss css/draw-steel.css --style=expanded --source-map --watch",
```
JPMeehan, line 13: we can just remove these too & also remove sass as a dev dependency (which ofc means package-lock changes)

css/applications/components/forms.css
```diff
@@ -0,0 +1,90 @@
+.draw-steel {
+  .item-form {
```
JPMeehan, line 2: Can this get moved? Where is this used?

css/applications/sheets/item-sheet.css
```diff
@@ -0,0 +1,85 @@
...
+        font-size: var(--font-size-18);
+      }
+    }
+  }
+
+  section[data-tab="description"] {
+    &.editorActive {
+      .prosemirror.editor {
+        &:not(.active) {
+          display: none !important;
```
JPMeehan, line 19: Is this something we could move up to a separate file so it is higher layer-wise instead of using `!important`, like some kind of `patches.css` file?

css/draw-steel.css
```diff
@@ -0,0 +1,6 @@
+@import url("https://fonts.googleapis.com/css2?family=Roboto:wght@400;700&display=swap");
+
+/* Light theme fixes */
+.theme-light .header-center .document-name h2 {
```
JPMeehan, line 4: we should check if we still need this one

system.json
```diff
@@ -77,7 +77,12 @@
     "verified": "13.338"
   },
   "esmodules": ["draw-steel.mjs"],
-  "styles": ["css/draw-steel.css"],
+  "styles": [
+    "css/draw-steel.css",
+    { "src": "css/variables/_variables.css", "layer": "variables" },
+    { "src": "css/elements/_elements.css", "layer": "elements" },
+    { "src": "css/applications/_applications.css", "layer": "applications" }
```
JPMeehan, line 84: Is this layer supposed to be for core applications? What's the advantage of going here instead of the default for the majority of our CSS?

system.json
```diff
@@ -77,7 +77,12 @@
     "verified": "13.338"
   },
   "esmodules": ["draw-steel.mjs"],
-  "styles": ["css/draw-steel.css"],
+  "styles": [
+    "css/draw-steel.css",
```
JPMeehan, line 81: High level question; does it make sense at all for us to try and rollup the actual css files when we ship them? I don't know the relative performance benefits.

JPMeehan (changes requested):
I think we can also go ahead and remove the old compiled file

css/applications/sheets/item-sheet.css
JPMeehan: Double checking the layers, is it possible that if we just moved the majority of this to the `system` layer we would be fine?

system.json
krbz999: A reduction in the number of shipped files is almost always a good thing.

JPMeehan: If so I think we should have our css folder as just the entry point files and then we stick the rest in `src/styles`

css/applications/components/forms.css
cdverrett94: searching `item-form` doesn't find anything. So I guess nowhere? It was in the SCSS files and I largely just copy/pasted without seeing what was/wasn't used. Want me to remove it?

system.json
cdverrett94: I guess there's probably not? Those layers just seemed applicable. I can have most of this moved to the system layer instead. Especially if we're wanting to just rollup to one main CSS file.

system.json
cdverrett94: Are we just wanting to rollup to a single css file? I'm not familiar and have never used rollup, so I'll see if I can figure out how to set that up in the config file myself. If I can't, I'll ping for help in discord.

css/draw-steel.css
cdverrett94: I don't think it is needed either. Without it, seems fine.

css/applications/components/forms.css
JPMeehan: Yes, I started the scss files from Boilerplate but then refactored kinda everything.

system.json
JPMeehan: I think the variables & elements are good inclusions, and I think for our `system` layer it's worth having that be explicit rather than the implicit "string means system layer"

system.json
JPMeehan: Well, I'm thinking one css file per layer we're using. So `elements` and `variables` are still valuable.

JPMeehan:
Other note: with 13.338 [fixing hotReload](https://github.com/foundryvtt/foundryvtt/issues/11939), this PR should also make sure it adjusts our hot reload flags so everything works as expected. This may take experimentation to figure out how the glob patterns work and if you need all the imported css from `src/styles`

system.json
```diff
@@ -77,7 +77,11 @@
     "verified": "13.338"
   },
   "esmodules": ["draw-steel.mjs"],
-  "styles": ["css/draw-steel.css"],
+  "styles": [
+    { "src": "src/styles/draw-steel.css", "layer": "system" },
+    { "src": "src/styles/variables/_variables.css", "layer": "variables" },
+    { "src": "src/styles/elements/_elements.css", "layer": "elements" }
```
JPMeehan, line 83: So, my thought here is that our entry points would still be in a `css/` folder we ship, so we don't have to ship `src` *or* modify `system.json` during the build process. This does mean that `elements` and `variables` may not import from `src/styles` at all, or at least handle things a little differently; kinda depends on how much stuff we expect those two layers to be using.

cdverrett94: Just want to make sure I understand the ask. Per our discord convo, we can do the roll-up stuff later. So should all of the css be in the `css` folder for now? Then when the roll-up stuff is done, it can be moved back to `src/styles`?

cdverrett94: I figured out how to rollup the CSS. It required a separate plugin `rollup-plugin-import-css`. I set up a separate rollup config for it so it could be built separately from the code.

Let me know how that looks and if I messed it up. Seemed to work fine for me. The biggest issue was that it creates an empty `.mjs` file in the `css` folder for each of the three built css files and i'm not sure how to stop that.

JPMeehan (changes requested):
I apologize if I wasn't clear enough - what I want is for the actual roots to be committed, similar to how `draw-steel.mjs` is. Then, during the build process, we roll everything up and replace the existing files with expanded versions that are basically "pre-imported". The main benefit is this actually removes the need to fully initialize the repository for it to run locally as a developer because the expected css files will be present.

cdverrett94:
Thanks for the clarification. That helped me figure out what you meant!

Let me know if this is the way you wanted it.

CONTRIBUTING.md
```diff
@@ -6,11 +6,11 @@ Please ensure there is an open issue about whatever contribution you are submitt
 
 ## Developer Tooling
 
-To start, clone this repository and either place it in or symlink it to your `Data/systems/draw-steel` user data directory. The CSS file referenced by `system.json` is derived from the SCSS files committed to the repository, so you must perform the following instructions to be able to load the repository in Foundry from the committed source.
+To start, clone this repository and either place it in or symlink it to your `Data/systems/draw-steel` user data directory. The CSS file referenced by `system.json` is derived from the CSS files in `src/styles` committed to the repository, so you must perform the following instructions to be able to load the repository in Foundry from the committed source.
```
JPMeehan, line 9:
```suggestion
To start, clone this repository and either place it in or symlink it to your `Data/systems/draw-steel` user data directory.
```

It should just work now which means we don't need the additional commentary, people can figure out that we use imports themselves.

JPMeehan (approved):
Looks great, just one small adjustment to the wording on the CONTRIBUTING instructions

CONTRIBUTING.md
cdverrett94: Ahh, oops. Forgot to update that part. Thanks for catching that.

PR #351, opened 2025-03-25
Follow up to CSS refactor

cdverrett94:
Forgot to remove an unused file from an earlier version of the CSS PR.
Exclude mjs files from css folder
Add the Google font import to the elements layer

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -99,6 +99,9 @@ jobs:
         npm run prepareRelease
         mv --force public/draw-steel.mjs draw-steel.mjs
         mv --force public/draw-steel.mjs.map draw-steel.mjs.map
+        mv --force public/css/draw-steel-system.css css/draw-steel-system.css
+        mv --force public/css/draw-steel-elements.css css/draw-steel-elements.css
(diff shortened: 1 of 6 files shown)


JPMeehan (approved):
Could you elaborate on the mjs part? The compiled files are .mjs that we rename?

cdverrett94:
When the CSS gets rolled up, it adds empty .mjs files for each of the built CSS files. Probably not necessary since the only time it should be run is on release and they should get filtered out by the github action.

JPMeehan:
Yeah basically we should git ignore any of the build result files. There's some utility in being able to run and see the output without that being a committable file.

JPMeehan:
Actually, can we ensure that the resulting .css file from the build has a separate name, similar to how we have the public/draw-steel.mjs that then gets moved? It ensures the development file and the build file stay separate even after a build.

cdverrett94:
Should it get built to `public/css` or just built to `css` with like a name of `draw-steel-layer-built.css`? 

If it's built to the public directory does `mv --force public/css/draw-steel-layer.css css/draw-steel-layer.css` for each layer file?

cdverrett94:
Let me know if that's good. Makes a lot more sense this way

Issue #137, opened 2025-01-05
Implement API functions in actor sheet


JPMeehan:
Recent PRs have implemented a number of API functions which need implementation on the actor sheet

- [x] Use Ability
- [x] Take Respite
- [ ] Spend Recovery
- [ ] Spend Hero token 
  - [x] Gain Surges
  - [ ] Regain Stamina

PR #353, opened 2025-03-27
Update actor sheet header layout and styling

### LINKED ISSUE #263: Make Character sheets resizeable
Currently the character sheets area set size, which should be fine, however I just ran into an unintended consequence:

The Troubadour class name is too big and as such blocks out the victories edit field in edit mode:

Image

If there would be the option to manually readjust the length of a sheet (bottom right corner drag button for example), this should be possible to amend manually, and if the list of abilities gets too long, that also can be amended that way.

### LINKED ISSUE #283: V13: Sheet mode toggle disappears in edit mode
Looks like the `div.header-right` is being wrapped to the next line and then being hidden by `overflow:hidden`. Not sure why it wraps though.

Image

Image

### LINKED ISSUE #305: Resize NPC character sheet header
This request probably will go hand in hand with request [#263](https://github.com/MetaMorphic-Digital/draw-steel/issues/263), but I wanted to highlight this particular input issue separately:

The NPC sheet header is too small as it currently omits some information:

The human storm mage is a level 3 platoon controller, but the controller part is being omitted although included in the meta data.
Image

Image

### LINKED ISSUE #352: No field to write down accumulated XP
We would need either a text field to be able to manually write accumulated XP in, after we convert our victories to experience during a respite, or an idea would also be, create a 'Respite' button, that converts accumulated victories to XP, sets victories and heroic resource to zero, and refiells recoveries to max?

cdverrett94:
This is the updated version of the stuff I had posted in the discord. Let me know your thoughts and what changes I should make to improve it.

The new functionality would be the XP input and the take respite button.

Actor Sheet
updated-character-header-play
updated-character-header-edit

Monster Sheet
updated-monster-header-play
updated-monster-header-play-malice
updated-monster-header-edit

Closes #352
Closes #305 
Closes #283 
Closes #263
Related to #137

### DIFF
--- a/src/module/applications/sheets/actor-sheet.mjs
+++ b/src/module/applications/sheets/actor-sheet.mjs
@@ -17,8 +17,11 @@ export default class DrawSteelActorSheet extends api.HandlebarsApplicationMixin(
   static DEFAULT_OPTIONS = {
     classes: ["draw-steel", "actor"],
     position: {
-      width: 600,
+      width: 850,
       height: 600,
+    },
+    window: {
+      resizable: true,
     },
     actions: {
       toggleMode: this._toggleMode,
@@ -29,6 +32,7 @@ export default class DrawSteelActorSheet extends api.HandlebarsApplicationMixin(
       roll: this._onRoll,
       useAbility: this._useAbility,
       toggleItemEmbed: this._toggleItemEmbed,
+      takeRespite: this._takeRespite,
     },
     form: {
       submitOnChange: true,
@@ -683,6 +687,10 @@ export default class DrawSteelActorSheet extends api.HandlebarsApplicationMixin(
 
     const part = target.closest("[data-application-part]").dataset.applicationPart;
     this.render({ parts: [part] });
+  }
+
+  static async _takeRespite(event, target) {
+    await this.actor.system.takeRespite();
   }
 
   /** Helper Functions */
--- a/src/module/applications/sheets/actor-sheet.mjs
+++ b/src/module/applications/sheets/actor-sheet.mjs
@@ -32,7 +32,6 @@ export default class DrawSteelActorSheet extends api.HandlebarsApplicationMixin(
       roll: this._onRoll,
       useAbility: this._useAbility,
       toggleItemEmbed: this._toggleItemEmbed,
-      takeRespite: this._takeRespite,
     },
     form: {
       submitOnChange: true,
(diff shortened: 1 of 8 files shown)


src/module/applications/sheets/actor-sheet.mjs
```diff
@@ -685,6 +689,10 @@ export default class DrawSteelActorSheet extends api.HandlebarsApplicationMixin(
     this.render({ parts: [part] });
   }
 
+  static async _takeRespite(event, target) {
```
JPMeehan, line 692: Method annotations

src/module/data/actor/character.mjs
```diff
@@ -258,12 +258,19 @@ export default class CharacterModel extends BaseActorModel {
     return this.parent.items.filter(i => i.type === "kit");
   }
 
+  /**
+   * Returns the total xp required for the next level
+   */
+  get nextLevelXP() {
+    if (!this.class) return 0;
+    return ds.CONFIG.hero.xp_track[this.level];
```
JPMeehan, line 266: Realizing we should probably have handling for if something goes "off track"

templates/actor/npc/header.hbs
```diff
@@ -5,66 +5,76 @@
...
+      {{#if isPlay}}
+      {{system.source.label}}
+      {{else}}
+      <a data-action="updateSource">
+        <i class="fa-solid fa-book"></i>
+        {{actor.system.source.label}}
+      </a>
+      {{/if}}
+    </div>
+    <div class="actor-data monster-data">
```
JPMeehan, line 59: We should consider if this would be better served by using dl, dt, and dd elements

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl

JPMeehan (changes requested):
Overall looks good, some adjustments

src/module/applications/sheets/actor-sheet.mjs
cdverrett94: Also realized I accidentally put it in the actor-sheet instead of the character sheet

templates/actor/npc/header.hbs
cdverrett94: Instead of the label/value setup right now?

src/module/data/actor/character.mjs
cdverrett94: Good call, the check for the class also isn't needed anymore since the level getting already returns 0 if no class.

templates/actor/npc/header.hbs
JPMeehan: Yeah, basically changing the semantics.

JPMeehan: I also like the default bolding more than using colons.

cdverrett94: Alright, I've got the npc sheet moved to using those elements. Had to reset some of the styles core sets.

Do you want me to move the character sheet play mode ones to using these as well?

PR #355, opened 2025-03-28
Combat group permissions

JPMeehan:
- Updated system minimum to 13.339
- Intellisense adjustments
- Updated changelog
- Small adjustment to the display of the combatant group turn button

Issue #358, opened 2025-03-28
Keyboard movement causes errors in ruler strike calculations


JPMeehan:
![Image](https://github.com/user-attachments/assets/2b0494fc-b86b-40d0-8da3-90b37396cd6f)

PR #359, opened 2025-03-28
Token movement improvements

### LINKED ISSUE #273: Implement movement actions for V13
In order for V13's new "Modify Movement Cost" region behavior to work properly with various movement types, a few steps must be taken:

**Register Movement Actions:**
Each "movement action" (movement type) must be defined in `CONFIG.Token.movement.actions`, e.g.
```js
CONFIG.Token.movement.actions = {
  walk: {
    label: "DRAW_STEEL.Actor.base.FIELDS.movement.walk.label",
    icon: "path/to/some/optional/icon"
  },
  burrow: {
    label: "DRAW_STEEL.Actor.base.FIELDS.movement.burrow.label",
    icon: "path/to/some/optional/icon"
  },
  // etc
};
```
The `icon` field in question is actually perhaps not currently used. Each action can also have a `deriveDifficulty` property, which would exclude them from the Region Behavior config, and would be useful if there are any movement types that are _explicitly_ something like "half your Walk speed."

**Override Token Placeable Methods:**
There are three relevant methods in `Token` that must be overridden (which, of course, will necessitate extending `Token`):
`_getDragWaypointProperties`, which is used when dragging a token to determine the `action` and whether it should count as a teleport,
`_getKeyboardMovementAction`, which is used when moving via keyboard to determine the movement `action`, and
`_getHUDMovementAction`, which is used when adjusting elevation via the Token HUD to determine the movement `action`.
I personally think it makes sense to have each refer to the same function on the Token Document (which, again, would need to be extended for this), something like:
```js
/** @override */
_getDragWaypointProperties() {
  return {
    action: this.document.getMovementAction() // optionally set teleport here as well, if allowing teleport as a movement action
  };
}

/** @override */
_getKeyboardMovementAction() {
  return this.document.getMovementAction();
}

/** @override */
_getHUDMovementAction() {
  return this.document.getMovementAction();
}
```
Then, naturally, the `TokenDocument` extension would need to implement a `getMovementAction` function, bringing us to:

**Enable Specifying a "Current" Movement Type:**
The specifics of this are up for discussion, but in essence, there must be a way of specifying what movement action/type is currently "in use." Whether this means modifying the `movement` schema field to have a "current," or storing that info somewhere else on the Actor, it should be either selectable or in some other way easily modified (perhaps with mutually-exclusive conditions for each movement type, or something). The "current" movement action would be what gets returned by the extended `TokenDocument` class's `getMovementAction`, which would result in the correct cost modifications taking place when moving through a region with the "Modify Movement Cost" behavior configured.

### LINKED ISSUE #328: Simplify actor to base speed + movement type tags
To my knowledge there is no creature that has two different values for different movement types. All are simply displayed as Speed 8 (burrow) for example.

This could be simplified to a base speed value input, and key word tags

JPMeehan:
- Refactors movement attributes to just be value/types/hover, since it does not appear that Draw Steel supports a creature with something a speed like "Walk 4, Swim 8".
- Adds token HUD controls to select active movement type + toggle shifting
  -  Does not implement the shifting logic for free strikes
  - Does not implement specific "movement cost" calculation for non-walk movements (e.g. a creature without a climb or swim costs 2x movement)
  - Adds "reference" property typing but does not add relevant journal pages

image

Will have merge conflicts with #355 over the intellisense fixes, will resolve those after this is merged + add the changelog entries.

Closes #328 

Closes #273


src/module/applications/hooks/token-hud.mjs
```diff
@@ -0,0 +1,87 @@
...
+/** @import DrawSteelTokenDocument from "../../documents/token.mjs"; */
+
+/**
+ * A hook event that fires when the TokenHUD application is rendered
+ * @param {TokenHUD} app
+ * @param {HTMLElement} html
+ * @param {object} context
+ * @param {ApplicationRenderOptions} options
+ */
+export async function renderTokenHUD(app, html, context, options) {
```
krbz999, line 14: I think it's worth extending the token HUD rather than using a hook.

src/module/config.mjs
```diff
@@ -84,6 +85,63 @@ DRAW_STEEL.sizes = {
...
 preLocalize("sizes", { key: "label" });
 
+/**
+ * @typedef DrawSteelTokenMovementAction
+ * @property {string} label    The label of the movement action
+ * @property {string} icon     The icon of the movement action
+ * @property {boolean} [speedOption] Whether or not to include this as an option on the actor sheet
+ * @property {string} [reference] UUID link to a journal page describing the movement type
+ * @property {(nonDerivedDifficulties: {[action: string]: number}) => number} [deriveDifficulty]
+ *   If set, this function is used to derive the terrain difficulty from from nonderived difficulties,
```
krbz999, line 95:
```suggestion
 *   If set, this function is used to derive the terrain difficulty from nonderived difficulties,
```

src/module/documents/token.mjs
```diff
@@ -7,7 +9,17 @@ export default class DrawSteelTokenDocument extends foundry.documents.TokenDocum
...
+  }
+
+  /* -------------------------------------------------- */
+
+  /**
+   * The token's current movement action.
+   * @type {string}
+   */
+  get movementType() {
+    return this.getFlag(systemID, "movementType") ?? CONFIG.Token.movement.defaultAction;
```
krbz999, line 22: Perhaps we should set this in `prepareDerivedData` instead, in a private property. To get validation that the flag property is in `CONFIG.Token.movement` if truthy.

krbz999 (commented):
Just some passing thoughts before I dive into a review tomorrow.

src/module/documents/token.mjs
JPMeehan: I don't see how that's too different from a getter? Can do the same extra validation there.

src/module/applications/hooks/token-hud.mjs
JPMeehan: The reason I was avoiding it was because it means copying all the rest of the template over rather than just inserting our stuff; but I'm open to the idea we should just go for it anyways.

krbz999: We won't need to copy the template, we can use `_onRender`.

src/module/documents/token.mjs
```diff
@@ -7,7 +9,17 @@ export default class DrawSteelTokenDocument extends foundry.documents.TokenDocum
    * @type {boolean}
    */
   get isShifting() {
-    return false; // TODO
+    return !!this.getFlag(systemID, "shifting");
```
krbz999, line 12: I was not seeing an effect from shifting in my testing.

krbz999 (commented):
Functionality seems good. Ruler labels work with the movements as well, and teleporting works.

Shifting however still gained Free Strikes when moving past enemies, but maybe the issue lies elsewhere.

JPMeehan:
> Functionality seems good. Ruler labels work with the movements as well, and teleporting works.
> 
> Shifting however still gained Free Strikes when moving past enemies, but maybe the issue lies elsewhere.

Shifting I had put to another PR since I figured between #358 and #364 there was enough other reasons to revisit those calculations.

src/module/applications/hooks/token-hud.mjs
JPMeehan: Will refactor, good point about _onRender

JPMeehan: Done

src/module/documents/token.mjs
JPMeehan: Adjusted to validate that the flag's value is `in` the actions record

src/module/applications/hud/token-hud.mjs
```diff
@@ -0,0 +1,113 @@
...
+   * Open the movement tray on left click, reset to walk on right
+   * @this DrawSteelTokenHUD
+   * @param {PointerEvent} event   The originating click event
+   * @param {HTMLElement} target   The capturing HTML element which defined a [data-action]
+   */
+  static async #toggleMovementTray(event, target) {
+    if (event.button === 0) {
+      let active = !this.#movementTrayActive;
+      this.#movementTrayActive = active;
+      const button = this.element.querySelector(".control-icon[data-action=toggleMovementTray]");
```
krbz999, line 91: This is `target`.

src/module/applications/hud/token-hud.mjs
```diff
@@ -0,0 +1,113 @@
...
+   * @param {HTMLElement} target   The capturing HTML element which defined a [data-action]
+   */
+  static async #shifting(event, target) {
+    /** @type {DrawSteelTokenDocument} */
+    const tokenDocument = this.document;
+    const isShifting = !tokenDocument.isShifting;
+    tokenDocument.setFlag(systemID, "shifting", isShifting);
+  }
+
+}
```
krbz999, line 113:
```suggestion
  }
}
```

src/module/applications/hud/token-hud.mjs
```diff
@@ -0,0 +1,113 @@
...
+
+  /**
+   * Open the movement tray on left click, reset to walk on right
+   * @this DrawSteelTokenHUD
+   * @param {PointerEvent} event   The originating click event
+   * @param {HTMLElement} target   The capturing HTML element which defined a [data-action]
+   */
+  static async #toggleMovementTray(event, target) {
+    if (event.button === 0) {
+      let active = !this.#movementTrayActive;
```
krbz999, line 89:
```suggestion
      const active = !this.#movementTrayActive;
```

krbz999 (approved):

PR #365, opened 2025-03-31
Minion Token HUD improvements

### LINKED ISSUE #357: Improve token HUD stamina bar and input to support minion squad stamina.

cdverrett94:
Closes #357 

- I think I accounted for all of the times the tokens' bars would need to be updated.
   - Combatant group change, combatant deletion, group stamina change, group deletion

### DIFF
--- a/src/module/data/actor/base.mjs
+++ b/src/module/data/actor/base.mjs
@@ -182,7 +182,7 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
   get combatGroup() {
     const groups = this.combatGroups;
 
-    return groups.first();
+    return (group.size === 1) ? groups.first() : undefined;
   }
 
   /**
--- a/src/module/data/actor/base.mjs
+++ b/src/module/data/actor/base.mjs
@@ -168,21 +168,15 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
    * @returns {Set<DrawSteelCombatantGroup>}
    */
   get combatGroups() {
-    const combatants = game.combat?.getCombatantsByActor(this.parent) ?? [];
-    const groups = new Set();
-    combatants.forEach(c => groups.add(c.group));
-
-    return groups;
+    return new Set(game.combat?.getCombatantsByActor(this.parent).map(c => c.group) ?? []);
   }
 
   /**
-   * Returns the combatant group if the actor is only a part of one. If they are in multiple, return null;
+   * Returns the first combatant group in all of the actor's combatant groups.
    * @returns {DrawSteelCombatantGroup | null}
    */
   get combatGroup() {
-    const groups = this.combatGroups;
-
-    return (group.size === 1) ? groups.first() : null;
+    return this.combatGroups.first() || null;
   }
 
   /**
@@ -333,7 +327,7 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
     if (this.isMinion) {
       const combatGroups = this.combatGroups;
       if (combatGroups.size === 1) {
-        return combatGroups.first().update({ "system.staminaValue": group.system.staminaValue - damage });
+        return this.combatGroup.update({ "system.staminaValue": group.system.staminaValue - damage });
       }
       else if (combatGroups.size === 0) {
         ui.notifications.warn("DRAW_STEEL.CombatantGroup.Error.MinionNoSquad", { localize: true });
--- a/src/module/data/actor/base.mjs
+++ b/src/module/data/actor/base.mjs
@@ -177,12 +177,12 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
 
   /**
    * Returns the combatant group if the actor is only a part of one. If they are in multiple, return null;
-   * @returns {DrawSteelCombatantGroup | undefined}
+   * @returns {DrawSteelCombatantGroup | null}
    */
   get combatGroup() {
     const groups = this.combatGroups;
 
-    return (group.size === 1) ? groups.first() : undefined;
+    return (group.size === 1) ? groups.first() : null;
   }
 
   /**
--- a/src/module/data/actor/base.mjs
+++ b/src/module/data/actor/base.mjs
@@ -177,13 +177,12 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
 
   /**
    * Returns the combatant group if the actor is only a part of one. If they are in multiple, return null;
-   * @returns {DrawSteelCombatantGroup | null}
+   * @returns {DrawSteelCombatantGroup | undefined}
    */
   get combatGroup() {
     const groups = this.combatGroups;
-    const [group] = groups;
 
-    return (groups.size === 1) ? group : null;
+    return groups.first();
   }
 
   /**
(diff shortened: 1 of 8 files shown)


src/module/data/actor/base.mjs
```diff
@@ -163,6 +163,29 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
...
+
+  /**
+   * Returns the combatant group if the actor is only a part of one. If they are in multiple, return null;
+   * @returns {DrawSteelCombatantGroup | null}
+   */
+  get combatGroup() {
+    const groups = this.combatGroups;
+    const [group] = groups;
+
+    return (groups.size === 1) ? group : null;
```
JPMeehan, line 186: Foundry adds a `first()` property that we can use

src/module/data/actor/base.mjs
```diff
@@ -309,16 +332,12 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
...
-      const combatants = game.combat?.getCombatantsByActor(this.parent) ?? [];
-      const sameGroup = combatants.every((c) => c.group === combatants[0].group);
-      if ((combatants.length > 0) && sameGroup) {
-        /** @type {DrawSteelCombatantGroup} */
-        const group = combatants[0].group;
-        if (group) return group.update({ "system.staminaValue": group.system.staminaValue - damage });
-        else ui.notifications.warn("DRAW_STEEL.CombatantGroup.Error.MinionNoSquad", { localize: true });
+      const combatGroups = this.combatGroups;
+      if (combatGroups.size === 1) {
+        const [group] = combatGroups;
```
JPMeehan, line 337: same change here with using `first()`

src/module/data/combatant-group/squad.mjs
```diff
@@ -46,4 +46,9 @@ export default class SquadModel extends BaseCombatantGroupModel {
       return maxStam;
     }, 0);
   }
+
+  /** @inheritdoc */
+  _onDelete(options, userId) {
+    this.refreshTokens();
```
JPMeehan, line 52: Add super call

src/module/documents/combatant.mjs
```diff
@@ -28,6 +28,24 @@ export default class DrawSteelCombatant extends foundry.documents.Combatant {
     if (("initiative" in changes) && (changes.initiative < this.initiative)) await this.actor?.system._onStartTurn(this);
   }
 
+  _onUpdate(changed, userId) {
```
JPMeehan, line 31: 1. inheritdoc
2. missing `options` from signature

src/module/documents/combatant-group.mjs
```diff
@@ -135,4 +135,8 @@ export default class DrawSteelCombatantGroup extends foundry.documents.Combatant
     // Provide a default image
     if (!data.img) this.updateSource(this.constructor.getDefaultArtwork(data));
   }
+
+  _onUpdate(changed, options, userId) {
```
JPMeehan, line 139: 1. inheritdoc
2. Need to add the `super` call so you capture stuff like ClientDocumentMixin refreshing sheets.

JPMeehan (approved):
Structurally looks good but needs some specific adjustments

src/module/data/actor/base.mjs
cdverrett94: I thought I remembered there being a `first()` option, but when I looked up `Sets` I didn't find anything, so assumed I saw it somewhere else. lol

JPMeehan: Foundry's primitive additions are sometimes kinda arbitrary you just have to memorize them or always refer to the docs

src/module/data/actor/base.mjs
```diff
@@ -163,6 +163,28 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
...
+  }
+
+  /**
+   * Returns the combatant group if the actor is only a part of one. If they are in multiple, return null;
+   * @returns {DrawSteelCombatantGroup | undefined}
+   */
+  get combatGroup() {
+    const groups = this.combatGroups;
+
+    return (group.size === 1) ? groups.first() : undefined;
```
JPMeehan, line 185: Wait why switch to undefined?

src/module/data/combatant-group/base.mjs
```diff
@@ -15,4 +15,15 @@ export default class BaseCombatantGroupModel extends foundry.abstract.TypeDataMo
...
+
+  /**
+   * Refresh the token bars and HUD for each token within the group.
+   * Is triggered on stamina updates and each combatants updates and deletions.
+   */
+  refreshTokens() {
+    for (const combatant of this.parent.members) {
+      combatant.token?.object?.renderFlags.set({ refreshBars: true });
+    }
+    this.parent.parent._refreshTokenHUD(this.parent.members);
```
JPMeehan, line 27: We may as well add `get combat()` that returns `this.parent.parent`, plus also annotating that it could be null (if for some reason you have a temporary CombatantGroup created with `new`)

src/module/data/actor/base.mjs
```diff
@@ -163,6 +163,28 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
...
+  /**
+   * Returns a Set of all combatant groups this actor is a part of
+   * @returns {Set<DrawSteelCombatantGroup>}
+   */
+  get combatGroups() {
+    const combatants = game.combat?.getCombatantsByActor(this.parent) ?? [];
+    const groups = new Set();
+    combatants.forEach(c => groups.add(c.group));
+
+    return groups;
```
krbz999, line 175:
```suggestion
    return new Set(game.combat?.getCombatantsByActor(this.parent) ?? []);
```

src/module/data/actor/base.mjs
cdverrett94: Honestly, I had a brain fart when swapping to the `first()` bit since it returns undefined if there's no set, but the size check prevents that.

src/module/data/combatant-group/base.mjs
cdverrett94: Done

src/module/data/actor/base.mjs
cdverrett94: That changes the return from a Set of combatant groups to a Set of combatants.

src/module/data/actor/base.mjs
```diff
@@ -163,6 +163,28 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
...
+
+  /**
+   * Returns the combatant group if the actor is only a part of one. If they are in multiple, return null;
+   * @returns {DrawSteelCombatantGroup | null}
+   */
+  get combatGroup() {
+    const groups = this.combatGroups;
+
+    return (group.size === 1) ? groups.first() : null;
+  }
```
krbz999, line 186: In core, for tokens with multiple combatants, something like `token.combatant` returns the first found. Is there a specific reason we want `null` with multiple groups here and be inconsistent?

JPMeehan: Honestly just didn't know what to do with the multiple scenario. Always returning the first makes as much sense as anything

cdverrett94: This is largely used for minions where if they have multiple groups, then should update the minion actor instead of any one group. 

Suppose it could return the first and I'd just update the minion checks to be `isMinion && combatGroups.size === 1` instead of just `combatGroup`

cdverrett94: Okay, updated it to always return the first. Updated the places I was just checking `combatGroup` to instead check by size.

PR #366, opened 2025-03-31
Add power roll section to item sheet embed.

cdverrett94:
Addresses the [issue mentioned in discord](https://discordapp.com/channels/332362513368875008/1342298358664138805/1356322133042659519) where the power roll tiers weren't being added to the item sheet embed.

I also added the `Power Roll + Characteristics` to the embed as well with the characteristic that would be used emphasized.

ability-embed-updates


JPMeehan (approved):
Looks good

PR #372, opened 2025-04-02
Use description list element in actor header and item embeds

### LINKED ISSUE #356: Switch to using dl/dt/dd elements
We have a number of places that represent key: value data. We can improve our semantic HTML and perhaps get some styling improvements by switching to these elements.

cdverrett94:
Closes #356 

- Updated actor headers and item embeds to use the dl elements.
- Added an await to the project embed `getSheetContext` call because it was prevent the item link from showing up on the embed on the actor sheet.
- Also updated the default macro to use the namespaced call to the `Hotbar`. Felt weird to make a separate PR for just that.

project-embed-updates
ability-embed-updates
kit-embed-updates

### DIFF
--- a/src/module/data/item/project.mjs
+++ b/src/module/data/item/project.mjs
@@ -96,7 +96,7 @@ export default class ProjectModel extends BaseItemModel {
       systemFields: this.schema.fields,
       config: ds.CONFIG,
     };
-    this.getSheetContext(context);
+    await this.getSheetContext(context);
 
     const embed = document.createElement("div");
     embed.classList.add("project");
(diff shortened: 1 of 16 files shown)


JPMeehan:
Wow that looks a lot better

src/module/data/item/project.mjs
```diff
@@ -96,7 +96,7 @@ export default class ProjectModel extends BaseItemModel {
       systemFields: this.schema.fields,
       config: ds.CONFIG,
     };
-    this.getSheetContext(context);
+    await this.getSheetContext(context);
```
JPMeehan, line 99: Do we have any other spots we aren't awaiting this properly

JPMeehan (approved):

src/module/data/item/project.mjs
JPMeehan: Looks like we're inconsistent. Can we mark *all* of these functions as async and then await the calls in ability, kit, and project?

cdverrett94: Just marked them as async

PR #375, opened 2025-04-02
Add draw steel class to content embeds

cdverrett94:
Notice after the DL got merged, that the CSS wasn't affecting the embeds in the chat log because there was no parent with class "draw-steel". Easiest way seemed to be added the class to our embeds so they are targeted everywhere.


JPMeehan (approved):

JPMeehan:
Don't we also have a base embed? Should probably be consistent across all embeds this way

PR #380, opened 2025-04-03
Apply minion temporary stamina before applying to combat group

### LINKED ISSUE #379: Update `BaseActorModel#takeDamage` to take temporary stamina into account for minions
Currently the `BaseActorModel#takeDamage` returns before temp hp can be affected.

Example of minion temp stamina:
Image

cdverrett94:
Closes #379


krbz999:
Would be good to split this into two methods: `Actor#determineDamage` and `Actor#takeDamage`.

PR #391, opened 2025-04-14
Award victories post combat

### LINKED ISSUE #388: Award Victories at the End of Combat
As a Director, I would like to be prompted to give victories to all heroes at the end of combat. Ideally, it would prompt with an input to specify the number of victories to award.

cdverrett94:
Closes #388 

Award Victories

### DIFF
--- a/src/module/documents/combat.mjs
+++ b/src/module/documents/combat.mjs
@@ -71,6 +71,8 @@ export default class DrawSteelCombat extends foundry.documents.Combat {
       /** @type {MaliceModel} */
       const malice = game.actors.malice;
       await malice.endCombat();
+
+      await this.awardVictories();
     }
 
     return deletedCombat;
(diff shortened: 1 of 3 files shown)


src/module/documents/combat.mjs
```diff
@@ -166,4 +168,39 @@ export default class DrawSteelCombat extends foundry.documents.Combat {
...
+        title: "DRAW_STEEL.Combat.AwardVictories.Title",
+      },
+      ok: {
+        label: "DRAW_STEEL.Combat.AwardVictories.Button",
+      },
+      rejectClose: false,
+    });
+
+    if (fd) {
+      for (const combatant of this.combatants.values()) {
```
JPMeehan, line 198: Collections should be directly iterable without transforming to an array?

JPMeehan (approved):

src/module/documents/combat.mjs
cdverrett94: Oh yeah, thanks.

I also updated the changelog to include this.

PR #394, opened 2025-04-17
Improved system documentation workflow

### LINKED ISSUE #382: Integrate Wiki and Journal packs
The wiki pages should be mirrored to a "system documentation" journal.

Actual implementation should probably be some kind of `src/docs` folder that is full of markdown files. When updated, the changes are pushed to the wiki. Similarly, the compendium packs should build a journal from these files. 

Stretch goals
- [ ] Leverage "page level" for deeper organization 
- [ ] Leverage journal entry categories

JPMeehan:
Hopefully straightforward integration of the system wiki & in-foundry documentation that also enables community wiki contributions.

Closes #382


src/docs/Enrichers.md
```diff
@@ -0,0 +1,5 @@
+"Enrichers" refer to specialized text syntax that can be used in any rich text editor (such as an actor description, journal entry, or chat message). Enriched content displays simply in the editor but can represent large and complex displays in the rendered text.
+
+## Item Embeds
+
+All items support the @Embed<s></s>[uuid] syntax, which by default share the description of the item. For abilities, kits, and projects, their embeds feature additional information.
```
krbz999, line 5: Markup here for the embed seems faulty.

krbz999 (approved):

src/docs/Enrichers.md
JPMeehan: The issue is that if it's an un-interrupted `@Embed[uuid]` foundry will try and fail to parse, and if I keep the code block ticks then anything I stick in the middle to break the regex will get escaped and rendered. So I removed the code block ticks and stuck in a tiny HTML fragment to prevent foundry from trying to parse the embed reference.

PR #395, opened 2025-04-17
Add spending recovery to actor sheet

### LINKED ISSUE #137: Implement API functions in actor sheet

cdverrett94:
Closes #137

- Added `CharacterModel#spendRecovery` that reduces the recovery value by and adds stamina.
- Added `CharacterModel#spendStaminaHeroToken` that prompts the user to spend the tokens and adds stamina.
- Clicking "Recoveries" on the character sheet will spend a recovery
- Clicking "Stamina" on the character sheet will prompt for spending hero tokens

### DIFF
--- a/src/module/data/actor/base.mjs
+++ b/src/module/data/actor/base.mjs
@@ -306,7 +306,7 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
       return this.parent;
     }
 
-    return this.parent.update({ "system.stamina.value": this.stamina.value + stamina });
+    return this.parent.modifyTokenAttribute("stamina", stamina, true);
   }
 
   /**
--- a/src/module/data/actor/base.mjs
+++ b/src/module/data/actor/base.mjs
@@ -296,6 +296,20 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
   }
 
   /**
+   * Regain stamina equal to the provided value
+   * @param {number} stamina
+   * @returns {DrawSteelActor}
+   */
+  async regainStamina(stamina) {
+    if (!Number.isNumeric(stamina)) {
+      console.error("stamina is required to be a number.");
+      return this.parent;
+    }
+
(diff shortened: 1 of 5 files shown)


src/module/data/actor/base.mjs
```diff
@@ -295,6 +295,20 @@ export default class BaseActorModel extends foundry.abstract.TypeDataModel {
...
+   * @param {number} stamina
+   * @returns {DrawSteelActor}
+   */
+  async regainStamina(stamina) {
+    if (!Number.isNumeric(stamina)) {
+      console.error("stamina is required to be a number.");
+      return this.parent;
+    }
+
+    return this.parent.update({ "system.stamina.value": this.stamina.value + stamina });
```
JPMeehan, line 309: This is a good opportunity to add max clamping, although honestly there's a question if we should just be using `Actor#modifyTokenAttribute`?

JPMeehan (commented):
Overall like the approach and agree this can finally close out 137

src/module/data/actor/base.mjs
cdverrett94: Yeah, using `Actor#modifyTokenAttribute` will work. Our implementation already has the min/max clamping in place.

At some point, we also need to clamp the value when updating via the input on the actor sheet.

JPMeehan: I think the sheet not clamping is fine in case there's some kind of overheal situation or whatever.

cdverrett94: Fair enough

cdverrett94: I guess not. It was mostly for ensuring only a number would be passed.

JPMeehan (approved):

PR #403, opened 2025-04-20
Creature Role Token icons

ItsDrafty:
Added 12 token icons with credits to the creator for creature role tokens.


krbz999:
github seems to not want to show them visually.

image

JPMeehan (approved):
Great icons, not sure if we have any specific code we want to integrate these with or just keep them as a resource for now.

PR #409, opened 2025-04-24
Added combatant group ownership controls

JPMeehan:
Leveraged the new `ownership` property on combatant groups.


krbz999 (approved):
Rendering threw an error due to 13.340 changes and there was a deprecation due to `renderTemplate`. Looks good otherwise.

PR #425, opened 2025-04-26
[#419] Document Sheet Mixin

### LINKED ISSUE #419: Applications API - Dialog, async Application, and sheet mixin
- A sheet mixin would be a mixin used to form the base of our document sheets. We currently have two sheets (actors and items) making use of a lot of duplicated logic in eg edit mode. This mixin would form the "base" actor sheet (eg `DrawSteelActorSheet extends DrawSteelDocumentSheetMixin(DocumentSheet)` and similar for items) to enable sharing all this logic that is likely never going to make it into the base core foundry `DocumentSheet`.

- A system-specific implementation of `applications.api.Dialog` would be a simple subclass of the core foundry `api.Dialog`, adding in some base css classes. This should be used only for the *simplest* of prompts - likely nothing but `confirm` dialogs, anything that does not require any semi-advanced logic.

- The system-specific `Application` would hold boilerplate logic for more complicated prompts, with an async factory method, re-rendering logic, and using the core `HandlebarsTemplateMixin`. It would also have the same css and things from the system `Dialog`. This would be used for anything that requires user input, can change visually depending on user input, and should somehow be tied to a document like an actor or item.

These things would be good to implement sooner rather than later before we implement a slew of dialogs for, e.g., respites, or overhaul the roll configuration dialog.

To mimic the core structure, these would all be found in `ds.applications.api`.

krbz999:
Closes #419.

Adds a single mixin intended for use with `ActorSheet` and `ItemSheet` currently. I was conservative in moving things from the existing sheets as I ran out of time and it is possible we may want to use this for other document sheets, such as for effects.

A follow-up PR could likely move the drag-drop handlers into the mixin.

### DIFF
--- a/src/module/applications/api/document-sheet-mixin.mjs
+++ b/src/module/applications/api/document-sheet-mixin.mjs
@@ -74,9 +74,9 @@ export default base => {
         document: this.document,
         system: this.document.system,
         systemSource: this.document.system._source,
+        systemFields: this.document.system.schema.fields,
         flags: this.document.flags,
         config: ds.CONFIG,
-        systemFields: this.document.system.schema.fields,
       });
       return context;
     }
--- a/src/module/applications/api/document-sheet-mixin.mjs
+++ b/src/module/applications/api/document-sheet-mixin.mjs
@@ -1,0 +1,84 @@
+const { HandlebarsApplicationMixin } = foundry.applications.api;
+
+export default base => {
+  return class DSDocumentSheet extends HandlebarsApplicationMixin(base) {
+    /** @inheritdoc */
+    static DEFAULT_OPTIONS = {
+      classes: ["draw-steel"],
+      form: {
+        submitOnChange: true,
+        closeOnSubmit: false,
+      },
+      window: {
+        resizable: true,
+      },
+    };
+
+    /* -------------------------------------------------- */
+
+    /**
+     * Available sheet modes.
+     * @enum {number}
+     */
+    static MODES = Object.freeze({
+      PLAY: 1,
+      EDIT: 2,
+    });
+
+    /* -------------------------------------------------- */
+
+    /**
+     * The mode the sheet is currently in.
+     * @type {DSDocumentSheet.MODES}
+     */
+    _mode;
+
+    /* -------------------------------------------------- */
+
+    /**
+     * Is this sheet in Play Mode?
+     * @returns {boolean}
+     */
+    get isPlayMode() {
+      return this._mode === DSDocumentSheet.MODES.PLAY;
+    }
+
+    /* -------------------------------------------------- */
+
+    /**
+     * Is this sheet in Edit Mode?
+     * @returns {boolean}
+     */
+    get isEditMode() {
+      return this._mode === DSDocumentSheet.MODES.EDIT;
+    }
+
+    /* -------------------------------------------------- */
+
+    /** @inheritdoc */
+    _configureRenderOptions(options) {
+      super._configureRenderOptions(options);
+      if (options.mode && this.isEditable) this._mode = options.mode;
+    }
+
+    /* -------------------------------------------------- */
+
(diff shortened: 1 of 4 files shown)


src/module/applications/api/document-sheet-mixin.mjs
```diff
@@ -0,0 +1,84 @@
...
+    async _prepareContext(options) {
+      const context = await super._prepareContext(options);
+      Object.assign(context, {
+        isPlay: this.isPlayMode,
+        owner: this.document.isOwner,
+        limited: this.document.limited,
+        gm: game.user.isGM,
+        document: this.document,
+        system: this.document.system,
+        systemSource: this.document.system._source,
```
JPMeehan, line 76: Possibility of erroring if we reuse this for any document without a `system` property, not sure how likely that is though.

src/module/applications/api/document-sheet-mixin.mjs
```diff
@@ -0,0 +1,84 @@
...
+        flags: this.document.flags,
+        config: ds.CONFIG,
+        systemFields: this.document.system.schema.fields,
```
JPMeehan, line 79:
```suggestion
        systemFields: this.document.system.schema.fields,
        flags: this.document.flags,
        config: ds.CONFIG,
```
minor sorting change

src/module/applications/sheets/actor-sheet.mjs
```diff
@@ -104,38 +77,20 @@ export default class DrawSteelActorSheet extends api.HandlebarsApplicationMixin(
...
-      systemSource: this.actor.system._source,
-      flags: this.actor.flags,
-      // Adding a pointer to ds.CONFIG
-      config: ds.CONFIG,
-      // Necessary for formInput and formFields helpers
-      systemFields: this.document.system.schema.fields,
+    const context = await super._prepareContext(options);
+    Object.assign(context, {
+      actor: context.document,
       datasets: this._getDatasets(),
```
JPMeehan, line 87: maybe out of scope for this PR but I'm wondering if we should just drop this and use the `object` helper instead.

src/module/applications/sheets/item-sheet.mjs
```diff
@@ -132,40 +103,22 @@ export default class DrawSteelItemSheet extends api.HandlebarsApplicationMixin(
...
-      system: this.isPlayMode ? this.item.system : this.item.system._source,
-      systemSource: this.item.system._source,
-      flags: this.item.flags,
-      // Adding a pointer to ds.CONFIG
-      config: ds.CONFIG,
-      // You can factor out context construction to helper functions
+    const context = await super._prepareContext(options);
+    Object.assign(context, {
+      item: context.document,
+      system: context.isPlay ? context.system : context.systemSource,
```
JPMeehan, line 113: yeah clearly could/should make necessary adjustments to harmonize the use of systemSource vs. having this switch.

JPMeehan (approved):
Overall looks good, definitely room for follow up

src/module/applications/api/document-sheet-mixin.mjs
krbz999: I find it quite unlikely. However, I will make a follow-up PR that adds a convenience method for preparing field context that should take care of this.

src/module/applications/sheets/actor-sheet.mjs
krbz999: We only use it once, so the method isn't really worth having, I feel.

krbz999: I'm not finding anywhere we actually use this? If you could point it out.

src/module/applications/sheets/item-sheet.mjs
krbz999: Will resolve in a follow-up.

PR #427, opened 2025-04-26
Remove uses of "root.actor" and "root.item" in actor/item sheets

krbz999:
Bit of clean up of actor and item sheets before Ability sheets take over.

Related: #426.

### DIFF
--- a/src/module/applications/sheets/item-sheet.mjs
+++ b/src/module/applications/sheets/item-sheet.mjs
@@ -6,9 +6,7 @@ const { sheets, ux } = foundry.applications;
 /**
  * AppV2-based sheet for all item classes
  */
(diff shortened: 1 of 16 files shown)


src/module/applications/sheets/item-sheet.mjs
```diff
@@ -108,8 +108,10 @@ export default class DrawSteelItemSheet extends DSDocumentSheetMixin(
   /** @inheritdoc */
   async _prepareContext(options) {
     const context = await super._prepareContext(options);
+
+    const enrichment = { relativeTo: this.document, rollData: this.document.getRollData() };
```
JPMeehan, line 112: Unused?

JPMeehan (approved):
OK, makes sense to me

PR #428, opened 2025-04-27
Create utils directory.

krbz999:
- Renames the current `helpers/utils` to `helpers/localization` and moves the button html helper out into a new directory for actual utility methods.
- Adds a new method for getting unique set of actors from selected tokens.
- Uses said method in the damage application.

### DIFF
--- a/src/module/data/message/ability-use.mjs
+++ b/src/module/data/message/ability-use.mjs
@@ -1,4 +1,3 @@
-import { constructHTMLButton } from "../../helpers/utils.mjs";
 import { DamageRoll } from "../../rolls/damage.mjs";
 import BaseMessageModel from "./base.mjs";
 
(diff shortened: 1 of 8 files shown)


src/module/data/message/ability-use.mjs
```diff
@@ -80,7 +79,7 @@ export default class AbilityUseModel extends BaseMessageModel {
     const damageRolls = this.parent.rolls.filter(roll => roll instanceof DamageRoll);
     for (const roll of damageRolls) {
       const typeLabel = ds.CONFIG.damageTypes[roll.options.type]?.label ?? "";
-      const button = constructHTMLButton({
+      const button = ds.utils.constructHTMLButton({
```
JPMeehan, line 82: Does this mean there's an unused import at the top?

JPMeehan (approved):

src/module/data/message/ability-use.mjs
krbz999: Already removed. :)

Issue #416, opened 2025-04-26
Power rolls do not display bonus from an edge


krbz999:
A single edge adds a bonus of +2. The displayed roll's formula remains "2d10 + x" where it would be expected to display "2d10 + x + 2".

Additionally, we should consider displaying "Double Edge" instead of "2 Edges" when applying 2 edges.

JPMeehan:
I'm just gonna close this as part of 0.7, if we end up feeling strongly on double edge that's fine

PR #432, opened 2025-04-30
Update power roll messages to show formula per target

cdverrett94:
Related to #416

- Per that issue, banes/edges should show in the formula, which requires showing the formula per target. I set it up so that the formula and roll total are on the same line to conserve space when there are many targets.
- I didn't change the `2 Edges` to `Double Edge` as noted in the issue.
- In the process, I noticed banes weren't being applied due to a misplaced closing parenthesis.
- Flavor was also being incorrectly set so I fixed that as well

power-roll-formula

### DIFF
--- a/src/module/rolls/power.mjs
+++ b/src/module/rolls/power.mjs
@@ -265,6 +265,16 @@ export class PowerRoll extends DSRoll {
     return (this.dice[0].total >= this.options.criticalThreshold);
   }
 
+  /**
+   * Return a version of the formula that doesn't have the flavor text.
+   * @returns {string}
+   */
+  get flavorlessFormula() {
+    const flavorlessRoll = new this.constructor(this.formula);
+    for (const term of flavorlessRoll.terms) term.options.flavor = "";
+    return flavorlessRoll.formula;
+  }
+
   async _prepareContext({ flavor, isPrivate }) {
     const context = await super._prepareContext({ flavor, isPrivate });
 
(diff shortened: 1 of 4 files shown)


krbz999:
I suggest dropping the flavor on the individual modifiers and implementing a tooltip for the roll for a more detailed overview.

I.e., show "2d10 + 2 + 2 + 2" but on hover show, "2d10 + 2 + 2[Edge] + 2[Bonus]"

krbz999:
> That's a good suggestion! Is there an easy way to strip the flavor from the formula?

```js
for (const term of roll.terms) term.options.flavor = "";
```
Though note that this mutates the evaluated roll, so you may wish to save the original formula for the tooltip.

cdverrett94:
Thanks for that snippet. I've updated per the request and updated the screenshot in the initial post to showcase.

Instead of doing it and mutating the original formula, I added a getter to generate a new roll and stripped the flavor from that roll instead.

src/module/rolls/power.mjs
```diff
@@ -261,6 +265,16 @@ export class PowerRoll extends DSRoll {
     return (this.dice[0].total >= this.options.criticalThreshold);
   }
 
+  /**
+   * Return a version of the formula that doesn't have the flavor text.
+   * @returns {string}
+   */
+  get flavorlessFormula() {
+    const flavorlessRoll = new this.constructor(this.formula);
```
JPMeehan, line 273: Might be worth checking on the difference between this function and using `this.clone()`, which uses `new this.constructor(this._formula, this.data, this.options)`

JPMeehan (approved):
Generally looks good, and yeah the tooltip was a good adjustment

src/module/rolls/power.mjs
JPMeehan: I think this is fine, your reasoning makes sense to me. Could you add a comment?

cdverrett94: Done!

PR #434, opened 2025-05-01
Pseudo-documents: Document and subtype mixins

krbz999:
Adds two mixins, one for documents one for subtypes.

The mixin for subtypes lets us define where pseudo-document collections live, mirroring the structure of `Document.metadata.embedded`. This will be a record of document names (such as "PowerRoll") and the path relative to the root document (such as "system.power.rolls").

The mixin for documents lets us define shared functionality for all document classes, currently implementing just `getEmbeddedDocument` - this will allow pseudo-documents to have functional uuids that work with `fromUuid`.

The `metadata` properties have had to be turned into getters - I plan to add the `Record` of pseudo-document subtypes here as well as the application class that they use to render. Having them as static properties would throw an error.

### DIFF
--- a/src/module/data/item/_types.d.ts
+++ b/src/module/data/item/_types.d.ts
@@ -2,7 +2,6 @@ import { PowerRollModifiers } from "../../_types.js";
 import DrawSteelItem from "../../documents/item.mjs";
 import SourceModel from "../models/source.mjs";
 
-// TODO: this should extend SubtypeMetadata
 export type ItemMetaData = Readonly<{
   /** The expected `type` value */
   type: string;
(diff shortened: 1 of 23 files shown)


src/module/data/item/_types.d.ts
```diff
@@ -2,6 +2,7 @@ import { PowerRollModifiers } from "../../_types.js";
 import DrawSteelItem from "../../documents/item.mjs";
 import SourceModel from "../models/source.mjs";
 
+// TODO: this should extend SubtypeMetadata
```
JPMeehan, line 5: Simplest option is to do an & inside the Readonly

JPMeehan (approved):
Simple & good. This does affirm we won't ever be fully switching to TS, but I think I've made my peace with that and the symlinked types are good enough.

PR #435, opened 2025-05-01
Pseudo-documents: Pseudo-document data model

krbz999:
Adds the base pseudo-document datamodels from which all other pseudo-documents will inherit.

Pseudo-documents will support having pseudo-document collections themselves, and can either
- be an untyped PseudoDocument, which is a subtype that inherits from `PseudoDocument`.
- have multiple subtypes (think akin to dnd5e activities) which would benefit from having one "base" model the subtypes inherit from - these should inherit from `TypedPseudoDocument`.

### DIFF
--- a/src/module/data/pseudo-documents/pseudo-document.mjs
+++ b/src/module/data/pseudo-documents/pseudo-document.mjs
@@ -1,4 +1,4 @@
-const { DocumentIdField, StringField } = foundry.data.fields;
+const { DocumentIdField } = foundry.data.fields;
 
 /** @import { PseudoDocumentMetadata } from "../../_types" */
 
@@ -96,6 +96,8 @@ export default class PseudoDocument extends foundry.abstract.DataModel {
 
   /**
    * Reference to the sheet of this pseudo-document, registered in a static map.
+   * A pseudo-document is temporary, unlike regular documents, so the relation here
+   * is not one-to-one.
    * @type {PseudoDocumentSheet|null}
    */
   get sheet() {
--- a/src/module/data/pseudo-documents/pseudo-document.mjs
+++ b/src/module/data/pseudo-documents/pseudo-document.mjs
@@ -1,0 +1,272 @@
+const { DocumentIdField, StringField } = foundry.data.fields;
+
+/** @import { PseudoDocumentMetadata } from "../../_types" */
+
+export default class PseudoDocument extends foundry.abstract.DataModel {
+  /**
+   * Pseudo-document metadata.
+   * @type {PseudoDocumentMetadata}
+   */
+  static get metadata() {
+    return {
+      documentName: null,
+      embedded: {},
+    };
+  }
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Registered sheets.
+   * @type {Map<string, PseudoDocumentSheet>}
+   */
+  static #sheets = new Map();
+
+  /* -------------------------------------------------- */
+
+  /** @inheritdoc */
+  static defineSchema() {
+    return {
+      _id: new DocumentIdField({ initial: () => foundry.utils.randomID() }),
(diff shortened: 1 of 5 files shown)


src/module/data/pseudo-documents/pseudo-document.mjs
```diff
@@ -0,0 +1,272 @@
...
+    return path;
+  }
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Reference to the sheet of this pseudo-document, registered in a static map.
+   * @type {PseudoDocumentSheet|null}
+   */
+  get sheet() {
```
krbz999, line 101: I'm likely gonna make a follow-up PR once the base sheet class is merged to put this logic entirely in the sheet class - rather than a method to "destroy" a sheet, the pseudo-document should I feel rather have a method to grab/create a new sheet, which is logic that should reside in the sheet class.

Then the sheet class can be responsible for destroying itself when closing as well.

src/module/data/pseudo-documents/pseudo-document.mjs
```diff
@@ -0,0 +1,272 @@
+const { DocumentIdField, StringField } = foundry.data.fields;
```
JPMeehan, line 1: StringField is unused

src/module/data/pseudo-documents/pseudo-document.mjs
```diff
@@ -0,0 +1,272 @@
...
+   * @type {Map<string, PseudoDocumentSheet>}
+   */
+  static #sheets = new Map();
+
+  /* -------------------------------------------------- */
+
+  /** @inheritdoc */
+  static defineSchema() {
+    return {
+      _id: new DocumentIdField({ initial: () => foundry.utils.randomID() }),
```
JPMeehan, line 30: slightly concerned that it might be possible for an undefined `_id` to get initialized as different values on different clients, but I think this should generally work. Need to be careful though.

src/module/data/pseudo-documents/pseudo-document.mjs
```diff
@@ -0,0 +1,272 @@
...
+      path = [this.parent.fieldPath, this.parent.id, path].join(".");
+    }
+
+    return path;
+  }
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Reference to the sheet of this pseudo-document, registered in a static map.
```
JPMeehan, line 98: Maybe could elaborate a bit more that we expect this model to get destroyed/created regularly so it can't hold its own sheet reference?

src/module/data/pseudo-documents/pseudo-document.mjs
```diff
@@ -0,0 +1,272 @@
...
+    super._initialize(...args);
+    this.prepareData();
+  }
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Prepare data.
+   */
+  prepareData() {
```
JPMeehan, line 135: This is making me realize, where do these actually sit in data prep? I don't believe these will naturally get prepped by prepareEmbeddedDocuments, are we going to add that to the mixin?

JPMeehan (approved):

src/module/data/pseudo-documents/pseudo-document.mjs
krbz999: They will, I overrode `_initialize`.

src/module/data/pseudo-documents/pseudo-document.mjs
krbz999: I believe this to be the case even in dnd5e - however the `DocumentIdField` is required and can't be blank, you have to provide a value, so only abuse of the API will allow for the key (in `TypedObjectField`) to be different from the `_id` property.

PR #436, opened 2025-05-01
Pseudo-documents: Pseudo-document sheet

krbz999:
Adds the base API for a pseudo-document sheet whose purpose is to configure a single pseudo-document standalone. The sheet registers as an app for the base document, not the pseudo-document, and re-renders on any updates.

If the root document, a parent pseudo-document, or the pseudo-document itself is deleted, the sheet will close with no errors as intended.

A lot of the behavior is mimicking `DocumentSheet`, but saving a reference to the pseudo-document's uuid - this allows for dynamically fetching the current iteration of the pseudo-document, as unlike documents a pseudo-document is not itself updated but reinstantiated.

While the sheet could use `fromUuidSync`, this of course results in an error if the root document is in a compendium.

### DIFF
--- a/src/module/applications/api/pseudo-document-sheet.mjs
+++ b/src/module/applications/api/pseudo-document-sheet.mjs
@@ -1,0 +1,243 @@
+const { HandlebarsApplicationMixin, Application } = foundry.applications.api;
+
+export default class PseudoDocumentSheet extends HandlebarsApplicationMixin(Application) {
+  constructor(options) {
+    super(options);
+    this.#pseudoUuid = options.document.uuid;
+    this.#document = options.document.document;
+  }
+
+  /* -------------------------------------------------- */
+
+  /** @inheritdoc */
+  static DEFAULT_OPTIONS = {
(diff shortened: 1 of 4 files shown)


src/module/applications/api/pseudo-document-sheet.mjs
```diff
@@ -0,0 +1,243 @@
...
+    tag: "form",
+    window: {
+      contentClasses: ["standard-form"],
+    },
+  };
+
+  /* -------------------------------------------------- */
+
+  /** @inheritdoc */
+  static PARTS = {
```
JPMeehan, line 39: is this even helpful since you're always just gonna override it?

JPMeehan (approved):

src/module/applications/api/pseudo-document-sheet.mjs
krbz999: You can use the spread operator in a subclass though I don't feel strongly at all.

PR #437, opened 2025-05-01
Pseudo-documents: Collection Field

krbz999:
Adds a new `CollectionField`, a subclass of `TypedObjectField` (my beloved) which initializes as a `Collection` with some added API for filtering, storing invalid models, etc.

To support modules adding subtypes of pseudo-documents, the `CollectionField` will store invalid models in a different set where they cannot be reached.

Example use:
```js
static defineSchema() {
  return {
    powers: new ds.data.fields.CollectionField(ds.data.pseudoDocuments.TypedPseudoDocumentSubclass),
  };
}
```

### DIFF
--- a/src/module/data/fields/collection-field.mjs
+++ b/src/module/data/fields/collection-field.mjs
@@ -1,0 +1,142 @@
+const { EmbeddedDataField, TypedObjectField, TypedSchemaField } = foundry.data.fields;
+
+/**
+ * A collection that houses pseudo-documents.
+ */
+export default class CollectionField extends TypedObjectField {
+  constructor(model, options = {}, context = {}) {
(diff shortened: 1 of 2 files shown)


src/module/data/fields/collection-field.mjs
```diff
@@ -0,0 +1,142 @@
...
+    return this.reduce((pass, v, i) => pass && predicate(v, i, this), true);
+  }
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Convert the ModelCollection to an array of simple objects.
+   * @returns {object[]}    The extracted array of primitive objects.
+   */
+  toObject() {
```
JPMeehan, line 139: lmao not even giving the option of source false, go off king.

JPMeehan (approved):
Looks good to me

PR #438, opened 2025-05-01
Implement scaffolding of power roll effects in Ability items

krbz999:
Showing an example use of the new `CollectionField` and a `PseudoDocument`, this implements `PowerRollEffect`, a typed pseudo-document, residing in a collection at `system.power.effects` in `ability`-items.

No UI yet, but the sheet can render. To test:

```js
const item = /* my ability item */
await ds.data.pseudoDocuments.powerRollEffects.BasePowerRollEffect.create({ type: "damage" }, { parent: item });
item.system.power.effects.contents[0].sheet.render({ force: true });
```

### DIFF
--- a/src/module/data/item/ability.mjs
+++ b/src/module/data/item/ability.mjs
@@ -67,7 +67,6 @@ export default class AbilityModel extends BaseItemModel {
       effects: new ds.data.fields.CollectionField(ds.data.pseudoDocuments.powerRollEffects.BasePowerRollEffect),
     });
 
-    // TODO: move these into `power`?
     schema.effect = new fields.StringField({ required: true });
     schema.spend = new fields.SchemaField({
       value: new fields.NumberField({ integer: true }),
--- a/src/module/data/item/ability.mjs
+++ b/src/module/data/item/ability.mjs
@@ -19,6 +19,9 @@ export default class AbilityModel extends BaseItemModel {
     return foundry.utils.mergeObject(super.metadata, {
       type: "ability",
       detailsPartial: [systemPath("templates/item/partials/ability.hbs")],
+      embedded: {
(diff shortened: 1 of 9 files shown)


krbz999:
If untyped, may just want to rename from `BasePowerRoll` to `PowerRoll` and move it up one level.

JPMeehan:
OK, high level: A given ability is generally only going to have one actual power roll, the appeal of pseudodocuments here would be for trying to tackle all of the effects a given result might have.

krbz999:
> OK, high level: A given ability is generally only going to have one actual power roll, the appeal of pseudodocuments here would be for trying to tackle all of the effects a given result might have.

Were there not some edge case abilities we were tracking down?

src/module/data/item/ability.mjs
```diff
@@ -56,78 +58,16 @@ export default class AbilityModel extends BaseItemModel {
...
+    schema.power = new fields.SchemaField({
+      roll: new fields.SchemaField({
+        flat: new fields.BooleanField(),
+        formula: new FormulaField({ blank: true, initial: "@chr" }),
+        characteristics: new fields.SetField(setOptions()),
+      }),
+      effects: new ds.data.fields.CollectionField(ds.data.pseudoDocuments.powerRollEffects.BasePowerRollEffect),
     });
+
+    // TODO: move these into `power`?
```
JPMeehan, line 70: I think no, there's plenty of abilities that don't make a power roll but can have an effect

JPMeehan (approved):

krbz999:
If we're merging this, we should be committed to finish the ability data model and sheet, otherwise it'd be nonfunctional.
Could merge into a feature branch though.

PR #439, opened 2025-05-01
Remove unfilled origin button interactivity in play mode

### LINKED ISSUE #412: "Add Ancestry / Culture / Career / Class" buttons should be disabled in "play mode"
- [ ] Disable these buttons if not in Edit mode.
- [ ] Don't apply context menu if there is no relevant item.

cdverrett94:
Closes #412 

- In play mode, updated unfilled origin items to be a span displaying "No Class", "No Ancestry", etc. Having them saying "+ Add Class" without having interactivity seemed counterintuitive.
- Updated the context menu css selector to only create if it has `data-document-class` and either `data-item-id` or `data-effect-id`. This also stops it from creating on the other document create buttons (project, equipment, ability, etc) on the sheet.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -9,6 +9,7 @@
         "CombatantGroup": "Draw Steel Combatant Group Config"
       },
       "Add": "+ Add {itemName}",
+      "NoDocument": "No {documentName}",
       "Play": "Play",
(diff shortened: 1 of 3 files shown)


JPMeehan (approved):
LGTM

Issue #442, opened 2025-05-03
Pseudo-document sheets should be unregistered if the root document is deleted


krbz999:
This otherwise leads to a bug.

Repro:
- Create an item in the sidebar, create a pseudo-document on it.
- Add the item to an actor.
- Call the `sheet` getter of the pseudo-document.
- Delete the parent item off the actor.
- Add the item to the actor again from the sidebar (it has the same id).
- Attempt to render the pseudo document sheet.

PR #443, opened 2025-05-03
Psseudo-document sheet fix and tier rendering context.

krbz999:
Adds a `watch` parameter to the sheet method - this means the `PseudoDocument#sheet` getter can be accessed without registering a unique sheet; registering a *unique* sheet happens only if it is actually ever rendered. While that does mean a sheet gets instantiated twice in most cases, it prevents the issue in #442.

Also moved rendering context into the PowerRollEffect subtype.

### DIFF
--- a/src/module/data/pseudo-documents/power-roll-effects/damage-effect.mjs
+++ b/src/module/data/pseudo-documents/power-roll-effects/damage-effect.mjs
@@ -37,7 +37,7 @@ export default class DamagePowerRollEffect extends BasePowerRollEffect {
       this.damage[`tier${n}`].value ??= this.#defaultDamageValue(n);
     }
 
-    this.text ||= "{{damage}}";
+    // this.text ||= "{{damage}}";
   }
 
   /* -------------------------------------------------- */
--- a/src/module/data/pseudo-documents/power-roll-effects/damage-effect.mjs
+++ b/src/module/data/pseudo-documents/power-roll-effects/damage-effect.mjs
@@ -33,10 +33,55 @@ export default class DamagePowerRollEffect extends BasePowerRollEffect {
   prepareDerivedData() {
     super.prepareDerivedData();
 
-    this.damage.tier1.value ??= 1;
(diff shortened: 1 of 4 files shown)


src/module/data/pseudo-documents/power-roll-effects/damage-effect.mjs
```diff
@@ -33,10 +33,55 @@ export default class DamagePowerRollEffect extends BasePowerRollEffect {
   prepareDerivedData() {
     super.prepareDerivedData();
 
-    this.damage.tier1.value ??= 1;
-    this.damage.tier2.value ??= 2 * this.damage.tier1.value;
-    this.damage.tier3.value ??= 3 * this.damage.tier1.value;
+    for (const n of [1, 2, 3]) {
```
JPMeehan, line 36: If it was any longer I'd say use Array.fromRange but for only 3 values this is probably just fine

src/module/data/pseudo-documents/power-roll-effects/damage-effect.mjs
```diff
@@ -33,10 +33,55 @@ export default class DamagePowerRollEffect extends BasePowerRollEffect {
...
+      this.damage[`tier${n}`].value ??= this.#defaultDamageValue(n);
+    }
 
     this.text ||= "{{damage}}";
```
JPMeehan, line 40: I take it this is a future PR problem?

JPMeehan (approved):

src/module/data/pseudo-documents/power-roll-effects/damage-effect.mjs
krbz999: The text? Yeah I think so. We should properly figure out the purpose of it - or I could remove it for now.

JPMeehan: I think comment out for now and see where that leads us

PR #448, opened 2025-05-07
Implemented _getSegmentStyle for DrawSteelTokenRuler

JPMeehan:
Simple green/yellow/red, using the same colors as the default token disposition. Added some special handling for teleports

image

### DIFF
--- a/src/module/canvas/placeables/tokens/token-ruler.mjs
+++ b/src/module/canvas/placeables/tokens/token-ruler.mjs
@@ -118,14 +118,28 @@ export default class DrawSteelTokenRuler extends foundry.canvas.placeables.token
   _getSegmentStyle(waypoint) {
     const segment = super._getSegmentStyle(waypoint);
 
-    const value = foundry.utils.getProperty(this, "token.document.actor.system.movement.value") ?? Infinity;
-
-    // green for up to 1x, yellow for up to 2x, red for 3x
+    // color order
     const colors = [0x33BC4E, 0xF1D836, 0xE72124];
 
-    const index = Math.min(2, Math.floor(waypoint.measurement.cost / value));
+    if (waypoint.actionConfig.teleport) {
+      // Teleports on creatures without a teleport speed are ignored for distance calculations
+      // It's possible we should be also subtracting them for mixed paths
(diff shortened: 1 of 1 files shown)


krbz999:
As commented on the league, I think grid highlights would be better.

JPMeehan:
Updated code to use both segment and grid highlight styles

image
image

src/module/canvas/placeables/tokens/token-ruler.mjs
```diff
@@ -112,35 +112,59 @@ export default class DrawSteelTokenRuler extends foundry.canvas.placeables.token
...
+    return style;
+  }
+
+  /**
+   * Adjusts the grid or segment style based on the token's movement characteristics
+   * @param {{ color?: PIXI.ColorSource }} style        - The calculated style properties from the parent class
+   * @param {DeepReadonly<TokenRulerWaypoint>} waypoint - The waypoint being adjusted
+   * @protected
+   */
+  _speedValueStyle(style, waypoint) {
```
krbz999, line 145: Do we expext this to be subclassed? If not, I think we should default to private.

src/module/canvas/placeables/tokens/token-ruler.mjs
```diff
@@ -112,35 +112,59 @@ export default class DrawSteelTokenRuler extends foundry.canvas.placeables.token
...
+
+  /**
+   * @inheritdoc
+   * @param {DeepReadonly<Omit<TokenRulerWaypoint, "index"|"center"|"size"|"ray">>} waypoint
+   * @param {DeepReadonly<foundry.grid.types.GridOffset3D>} offset
+   */
+  _getGridHighlightStyle(waypoint, offset) {
+    const style = super._getGridHighlightStyle(waypoint, offset);
+
+    this._speedValueStyle(style, waypoint);
```
krbz999, line 134: Some unneccessary spacing between lines. Also add some dividers.

krbz999 (approved):

krbz999:
LGTM. We can adjust the style of the individual nodes, too, if we're feeling extra but I don't feel strongly.

PR #449, opened 2025-05-07
Simple embed implementation

JPMeehan:
Added a basic implementation of toEmbed for active effects

### DIFF
--- a/src/module/data/effect/base.mjs
+++ b/src/module/data/effect/base.mjs
@@ -60,4 +60,19 @@ export default class BaseEffectModel extends foundry.abstract.TypeDataModel {
       label: this.durationLabel,
     };
   }
+
+  /** @inheritdoc */
+  async toEmbed(config, options = {}) {
+
+    // TODO: Figure out what roll data is "supposed" to be for effects
+
+    const enriched = await foundry.applications.ux.TextEditor.implementation.enrichHTML(this.parent.description, options);
+
+    const embed = document.createElement("div");
+    embed.classList.add("draw-steel", this.parent.type);
+    embed.innerHTML = enriched;
+
+    return embed;
+  }
+
 }
--- a/src/module/data/effect/base.mjs
+++ b/src/module/data/effect/base.mjs
@@ -1,3 +1,4 @@
+import enrichHTML from "../../utils/enrichHTML.mjs";
 import FormulaField from "../fields/formula-field.mjs";
 
 /**
@@ -64,8 +65,13 @@ export default class BaseEffectModel extends foundry.abstract.TypeDataModel {
   /** @inheritdoc */
   async toEmbed(config, options = {}) {
 
-    options.rollData ??= this.parent.getRollData();
-    const enriched = await foundry.applications.ux.TextEditor.implementation.enrichHTML(this.parent.description, options);
+    // Override document-related options with this document's info so it's consistent wherever embedded
+    const enriched = await enrichHTML(this.description.value, {
+      ...options,
+      relativeTo: this.parent,
+      rollData: this.parent.getRollData(),
+      secrets: this.parent.isOwner,
+    });
 
     const embed = document.createElement("div");
     embed.classList.add("draw-steel", this.parent.type);
--- a/src/module/data/effect/base.mjs
+++ b/src/module/data/effect/base.mjs
@@ -65,13 +65,7 @@ export default class BaseEffectModel extends foundry.abstract.TypeDataModel {
   /** @inheritdoc */
   async toEmbed(config, options = {}) {
 
-    // Override document-related options with this document's info so it's consistent wherever embedded
-    const enriched = await enrichHTML(this.description.value, {
-      ...options,
-      relativeTo: this.parent,
-      rollData: this.parent.getRollData(),
-      secrets: this.parent.isOwner,
-    });
+    const enriched = await enrichHTML(this.description.value, { ...options, relativeTo: this.parent });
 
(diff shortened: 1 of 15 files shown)


krbz999:
But why do we need this?

JPMeehan:
> But why do we need this?

Provides a basic & easy fallback since we otherwise have ownership of the system models.

krbz999:
> > But why do we need this?
> 
> Provides a basic & easy fallback since we otherwise have ownership of the system models.

My question is rather, what are we implementing right now that is requiring this? I don't think we should add new features just for the sake of adding features, they have to serve a purpose.

JPMeehan:
> > > But why do we need this?
> > 
> > 
> > Provides a basic & easy fallback since we otherwise have ownership of the system models.
> 
> My question is rather, what are we implementing right now that is requiring this? I don't think we should add new features just for the sake of adding features, they have to serve a purpose.

I guess there's nothing _right now_, I was mostly prompted by some discussion of expanding the range of core embeds. In terms of practical use, I could see minimizing duplicate text by using `@Embed[.ActiveEffect.relativeId]` to grab text from a child AE in an item.

krbz999:
> I guess there's nothing _right now_, I was mostly prompted by some discussion of expanding the range of core embeds. In terms of practical use, I could see minimizing duplicate text by using `@Embed[.ActiveEffect.relativeId]` to grab text from a child AE in an item.

That's a good use case, though not one we need right now - but perhaps soon? This isn't a very large PR so I don't feel strongly, but I'm worried about bloat if implementing features for the sake of features.

src/module/data/effect/base.mjs
```diff
@@ -60,4 +60,19 @@ export default class BaseEffectModel extends foundry.abstract.TypeDataModel {
       label: this.durationLabel,
     };
   }
+
+  /** @inheritdoc */
+  async toEmbed(config, options = {}) {
+
+    // TODO: Figure out what roll data is "supposed" to be for effects
```
krbz999, line 67: If we can figure this out, then I'll be happy with it. My suggestion would be to implement `ActiveEffect#getRollData` which would just return `{ ...this.system }`.

JPMeehan: Put together a more comprehensive function. I expect in the future we'll want to figure out if we also need to expand that for applied effects but it's a pretty good start.

src/module/data/effect/base.mjs
```diff
@@ -60,4 +60,19 @@ export default class BaseEffectModel extends foundry.abstract.TypeDataModel {
...
+
+    const enriched = await foundry.applications.ux.TextEditor.implementation.enrichHTML(this.parent.description, options);
```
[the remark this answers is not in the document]
JPMeehan: So, I'm working up a possible implementation, but in places where it's _not_ a one-off like our ActorSheet, I use `const TextEditor = foundry.applications.ux.TextEditor.implementation;`. I agree the long path is really annoying and I wish core had considered that when they made these adjustments.

src/module/data/item/kit.mjs
```diff
@@ -102,14 +103,13 @@ export default class KitModel extends BaseItemModel {
...
-        rollData: this.parent.getRollData(),
-        relativeTo: this.parent,
-      },
-    );
+    // Override document-related options with this document's info so it's consistent wherever embedded
+    context.enrichedDescription = await enrichHTML(this.description.value, {
+      ...options,
+      relativeTo: this.parent,
+      rollData: this.parent.getRollData(),
+      secrets: this.parent.isOwner,
```
krbz999, line 111: rollData and secrets are done in the helper method

src/module/data/item/base.mjs
```diff
@@ -78,9 +79,13 @@ export default class BaseItemModel extends SubtypeModelMixin(foundry.abstract.Ty
...
   async toEmbed(config, options = {}) {
-
-    options.rollData ??= this.parent.getRollData();
-    const enriched = await foundry.applications.ux.TextEditor.implementation.enrichHTML(this.description.value, options);
+    // Override document-related options with this document's info so it's consistent wherever embedded
+    const enriched = await enrichHTML(this.description.value, {
+      ...options,
+      relativeTo: this.parent,
+      rollData: this.parent.getRollData(),
+      secrets: this.parent.isOwner,
```
krbz999, line 87: Roll data and secrets are done in the helper method. imo we dont have a reason to enforce and override what is passed into `options`

src/module/data/effect/base.mjs
```diff
@@ -60,4 +61,23 @@ export default class BaseEffectModel extends foundry.abstract.TypeDataModel {
...
+
+  /** @inheritdoc */
+  async toEmbed(config, options = {}) {
+
+    // Override document-related options with this document's info so it's consistent wherever embedded
+    const enriched = await enrichHTML(this.description.value, {
+      ...options,
+      relativeTo: this.parent,
+      rollData: this.parent.getRollData(),
+      secrets: this.parent.isOwner,
```
krbz999, line 73: same here

krbz999 (approved):

src/module/data/item/kit.mjs
JPMeehan: I'm only null coalescing there, while this is is a hard set. Should I change the helper to hard override?

The basic case I'm envisioning is that you embed say an item in some other text block that is getting roll data — we want to preserve that items roll data context. Furthermore, if you embed the description of an item you don't own, it shouldn't display that items secrets.

PR #450, opened 2025-05-07
Update embed of power rolls

krbz999:
Adds a `toText` method to PowerRollEffects that formats them as a string. For a damage effect, this formats as shown:

image

(Unrelated addition, I had annoying difficulties with line endings, so I'm making an attempt at a permanent fix with `.gitattributes`.)

### DIFF
--- a/src/module/data/pseudo-documents/power-roll-effects/damage-effect.mjs
+++ b/src/module/data/pseudo-documents/power-roll-effects/damage-effect.mjs
@@ -94,9 +94,8 @@ export default class DamagePowerRollEffect extends BasePowerRollEffect {
     let damageTypes;
     let i18nString = "DRAW_STEEL.PSEUDO.POWER_ROLL_EFFECT.DAMAGE.formatted";
     if (types.size) {
-      damageTypes = new Intl.ListFormat(game.i18n.lang, {
-        style: "long", type: "disjunction",
-      }).format(Array.from(types).map(type => ds.CONFIG.damageTypes[type].label));
+      const formatter = game.i18n.getListFormatter({ type: "disjunction" })
+      damageTypes = formatter.format(Array.from(types).map(type => ds.CONFIG.damageTypes[type].label));
     } else {
       i18nString += "Typeless";
     }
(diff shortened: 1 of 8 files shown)


src/module/data/pseudo-documents/power-roll-effects/damage-effect.mjs
```diff
@@ -84,4 +84,22 @@ export default class DamagePowerRollEffect extends BasePowerRollEffect {
...
+  /** @inheritdoc */
+  toText(tier) {
+    const { value, types } = this.damage[`tier${tier}`];
+
+    let damageTypes;
+    let i18nString = "DRAW_STEEL.PSEUDO.POWER_ROLL_EFFECT.DAMAGE.formatted";
+    if (types.size) {
+      damageTypes = new Intl.ListFormat(game.i18n.lang, {
+        style: "long", type: "disjunction",
+      }).format(Array.from(types).map(type => ds.CONFIG.damageTypes[type].label));
```
JPMeehan, line 99: you can use `game.i18n.getListFormatter`, which also defaults to `style="long"`
```suggestion
      const formatter = game.i18n.getListFormatter({ type: "disjunction" })
      damageTypes = formatter.format(Array.from(types).map(type => ds.CONFIG.damageTypes[type].label));
```

templates/item/embeds/ability.hbs
```diff
@@ -28,28 +28,20 @@
...
-{{#if (or tier1 tier2 tier3 system.effect)}}
+{{#if powerRolls}}
 <section class="powerResult">
-  {{#if system.powerRoll.enabled}}
   <p>
     <strong>{{{localize "DRAW_STEEL.Roll.Power.RollPlusBonus" bonus=powerRollBonus}}}</strong>
   </p>
-  {{/if}}
   <dl class="power-roll-display">
-    {{#if tier1}}
```
JPMeehan, line 39: I believe the reason we have these individual checks is for the chat card results.

JPMeehan (commented):
Also, not sure if that should be this PR or not, but we do need to update the _types file documenting the schema & data prep-derived properties.

templates/item/embeds/ability.hbs
krbz999: Gonna need to redo that as well, yes. I can do this as part of this PR if preferred.

krbz999: So I had a hell of a time deciphering the current roll pipeline, but rolls once again appear in chat and the roll config dialog works.

JPMeehan (approved):

PR #458, opened 2025-05-10
Remove unnecessary unregisterSheet calls

### LINKED ISSUE #433: It is no longer required to unregister the AppV1 sheets
In 13.341+, any document that has no sheet registered instead uses `BaseSheet`, which is AppV2. This sheet does not need to be unregistered.

zarick1342:
Closes #433


JPMeehan (approved):
Could you check your line endings settings? Not sure why the other lines show change

krbz999:
> Could you check your line endings settings? Not sure why the other lines show change

Possibly (or probably) a system-wide issue with some files having committed line endings. We should remove the `unix` enforcement in the eslint file, too.

PR #459, opened 2025-05-11
Add an option to spend career points to Project context menu

### LINKED ISSUE #457: Add "Spend Career Points" to Project context menu
As a player it would be nice to have an option to easily spend my career points on a project. 

I'll tackle this one next week, just wanted an issue made first.

cdverrett94:
Closes #457 

- Adds an option to project context menu to spend career project points. Only shows if you have career points to spend and the project goal hasn't been met. This prompts an input with a range picker to choose how many points to spend.
- Moved the project complete stuff from the `roll` method into the `_onUpdate`, so that item creation automation runs no matter how the points are updated (rolling, spending career points, project sheet, etc).

### DIFF
--- a/src/module/applications/sheets/actor-sheet.mjs
+++ b/src/module/applications/sheets/actor-sheet.mjs
@@ -455,6 +455,29 @@ export default class DrawSteelActorSheet extends DSDocumentSheetMixin(sheets.Act
           await this.render();
         },
       },
+      // Project specific options
+      {
+        name: "DRAW_STEEL.Item.Project.SpendCareerPoints.Title",
+        icon: "<i class=\"fa-solid fa-hammer\"></i>",
+        condition: (target) => {
+          const item = this._getEmbeddedDocument(target);
+          if (item.type !== "project") return false;
+
+          const careerPoints = this.actor.system.career.system.projectPoints ?? 0;
+          const pointsToCompletion = item.system.goal - item.system.points;
+
+          return careerPoints && pointsToCompletion;
+        },
+        callback: async (target) => {
+          const item = this._getEmbeddedDocument(target);
+          if (!item) {
+            console.error("Could not find item");
+            return;
+          }
+          await item.system.spendCareerPoints();
+          await this.render();
+        },
+      },
       // All applicable options
       {
         name: "View",
--- a/src/module/applications/sheets/actor-sheet.mjs
+++ b/src/module/applications/sheets/actor-sheet.mjs
@@ -463,7 +463,7 @@ export default class DrawSteelActorSheet extends DSDocumentSheetMixin(sheets.Act
           const item = this._getEmbeddedDocument(target);
           if (item.type !== "project") return false;
 
-          const careerPoints = this.actor.system.career.system.projectPoints ?? 0;
+          const careerPoints = foundry.utils.getProperty(this.actor, "system.career.system.projectPoints") ?? 0;
           const pointsToCompletion = Math.max(0, item.system.goal - item.system.points);
 
           return careerPoints && pointsToCompletion;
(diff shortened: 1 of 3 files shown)


src/module/applications/sheets/actor-sheet.mjs
```diff
@@ -455,6 +455,29 @@ export default class DrawSteelActorSheet extends DSDocumentSheetMixin(sheets.Act
...
       },
+      // Project specific options
+      {
+        name: "DRAW_STEEL.Item.Project.SpendCareerPoints.Title",
+        icon: "<i class=\"fa-solid fa-hammer\"></i>",
+        condition: (target) => {
+          const item = this._getEmbeddedDocument(target);
+          if (item.type !== "project") return false;
+
+          const careerPoints = this.actor.system.career.system.projectPoints ?? 0;
```
JPMeehan, line 466: If there's no career it will error at `career.system`. Just use `foundry.utils.getProperty` rather than try to optional chain this.

src/module/data/item/project.mjs
```diff
@@ -88,6 +88,38 @@ export default class ProjectModel extends BaseItemModel {
...
+  }
+
+  /**
+   * @inheritdoc
+   */
+  _onUpdate(changed, options, userId) {
+    super._onUpdate(changed, options, userId);
+
+    // When the project is completed, notify the user and create any yielded item.
+    if (options.completeProject) {
```
JPMeehan, line 111: _onUpdate runs on every client, so you probably want to filter by userId

src/module/data/item/project.mjs
```diff
@@ -88,6 +88,38 @@ export default class ProjectModel extends BaseItemModel {
...
   }
 
+  /**
+   * @inheritdoc
+   */
+  async _preUpdate(changes, options, user) {
+    const allowed = await super._preUpdate(changes, options, user);
+    if (allowed === false) return false;
+
+    if (("system" in changes) && ("points" in changes.system)) {
```
JPMeehan, line 98: `foundry.utils.hasProperty` is probably more straightforward

src/module/data/item/project.mjs
```diff
@@ -88,6 +88,38 @@ export default class ProjectModel extends BaseItemModel {
...
+      // Mark the project for completion only if the points meet the goal and it hasn't already been completed.
+      options.completeProject = (changes.system.points >= this.goal) && (this.points < this.goal);
```
JPMeehan, line 100: Feels like should also check for `this.parent` here.

src/module/data/item/project.mjs
```diff
@@ -190,4 +199,66 @@ export default class ProjectModel extends BaseItemModel {
 
     return DrawSteelChatMessage.create(messageData);
   }
+
+  /**
+   * Spend a variable amount of the actor's project points from their career on this project
+   */
+  async spendCareerPoints() {
+    const careerPoints = this.actor.system.career.system.projectPoints ?? 0;
```
JPMeehan, line 207: Issues with no actor or no career, should more gracefully fail

JPMeehan (changes requested):
Issues with nullish handling

src/module/applications/sheets/actor-sheet.mjs
cdverrett94: Oops. Thought I had the optional chaining in for the career each time I checked for it. Will use that util instead! Thanks.

src/module/data/item/project.mjs
cdverrett94: Always forget about it running on every client.

src/module/data/item/project.mjs
cdverrett94: Good call.

src/module/data/item/project.mjs
cdverrett94: Added checks for both here and added an actor check in the complete crafting project method as well.

JPMeehan (approved):

PR #468, opened 2025-05-12
[#467] List power roll effects.

### LINKED ISSUE #467: The `Create Power Roll Effect` button doesn't overflow properly with more options
Image

krbz999:
Closes #467.

image

image

Turned the list of powers into, well, a list. Also added `disabled` attribute to the formGroups and fixed the tab not saving scroll position.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -462,6 +462,9 @@
           },
           "power": {
             "label": "Power Roll",
+            "effects": {
+              "label": "Power Roll Effects"
+            },
             "roll": {
               "flat": {
                 "label": "Flat"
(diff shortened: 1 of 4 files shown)


JPMeehan:
Given that we can end up with upwards of two paragraphs of text in the Effect, is there any consideration to trying to just make it flex and wrap nicely rather than have so much whitespace?

krbz999:
> Given that we can end up with upwards of two paragraphs of text in the Effect, is there any consideration to trying to just make it flex and wrap nicely rather than have so much whitespace?

We could make it look nicer, though I don't see value in showing a lot of the effects' properties on the item sheet - this can all be shown in the (derived) description, no?

JPMeehan:
> > Given that we can end up with upwards of two paragraphs of text in the Effect, is there any consideration to trying to just make it flex and wrap nicely rather than have so much whitespace?
> 
> We could make it look nicer, though I don't see value in showing a lot of the effects' properties on the item sheet - this can all be shown in the (derived) description, no?

My concern is that only one per line seems to be pretty space-inefficient. Even a double column would be huge

PR #470, opened 2025-05-13
Fixed various issues with ability internals

JPMeehan:
- Kit info
- Characteristic info
- Typings

### DIFF
--- a/src/module/data/pseudo-documents/_types.d.ts
+++ b/src/module/data/pseudo-documents/_types.d.ts
@@ -4,3 +4,15 @@ import { ApplicationConfiguration } from "@client/applications/_types";
 import { DialogV2Configuration, DialogV2WaitOptions } from "@client/applications/api/dialog.mjs";
 
 export interface TypedPseudoDocumentCreateDialogOptions extends ApplicationConfiguration, DialogV2Configuration, DialogV2WaitOptions {}
+
+declare module "./pseudo-document.mjs" {
+  export default interface PseudoDocument {
+    _id: string;
+  }
+}
+
+declare module "./typed-pseudo-document.mjs" {
+  export default interface TypedPseudoDocument {
+    type: string;
+  }
+}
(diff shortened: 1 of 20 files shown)


src/module/data/pseudo-documents/_types.d.ts
```diff
@@ -4,3 +4,15 @@ import { ApplicationConfiguration } from "@client/applications/_types";
...
 export interface TypedPseudoDocumentCreateDialogOptions extends ApplicationConfiguration, DialogV2Configuration, DialogV2WaitOptions {}
+
+declare module "./pseudo-document.mjs" {
+  export default interface PseudoDocument {
+    _id: string;
+  }
+}
+
+declare module "./typed-pseudo-document.mjs" {
+  export default interface TypedPseudoDocument {
```
krbz999, line 15: This extends `PseudoDocument`, are we not able to annotate that?

src/module/data/pseudo-documents/power-roll-effects/base-power-roll-effect.mjs
```diff
@@ -45,6 +47,24 @@ export default class BasePowerRollEffect extends TypedPseudoDocument {
 
   /* -------------------------------------------------- */
 
+  /**
+   * Reference to the grandparent item
+   * @type {DrawSteelItem}
+   */
+  get item() {
+    return this.parent?.parent ?? null;
+  }
```
krbz999, line 56: This getter already exists as `document`. We can add an alias, sure, if we're dead certain we won't reuse this for non-items.

```suggestion
  get item() {
    return this.document;
  }
```

src/module/data/pseudo-documents/power-roll-effects/base-power-roll-effect.mjs
```diff
@@ -45,6 +47,24 @@ export default class BasePowerRollEffect extends TypedPseudoDocument {
...
+
+  /**
```
krbz999, line 58: Add divider.

krbz999 (changes requested):

src/module/data/pseudo-documents/_types.d.ts
JPMeehan: Like all of our other type annotations the extension doesn't need to be repeated, we just need to declaration merge here. The TS server knows it's an extension because the class itself has the extension.

krbz999 (approved):

PR #476, opened 2025-05-13
417 floating damage numbers

### LINKED ISSUE #417: Display scrolling damage numbers when applying damage
The damage types seem to be set in stone and damage application is functional. It would be good if tokens displayed scrolling damage numbers, color-coded by the damage type.

zarick1342:
Fixes #417 

I think maybe the colors will need review, but it seems to work pretty well. I'm not sure if we want to add this to a hook somewhere so that it happens for any change in stamina, right now it's specifically on the "apply damage".

### DIFF
--- a/src/module/config.mjs
+++ b/src/module/config.mjs
@@ -482,7 +482,7 @@ preLocalize("skills.list", { key: "label" });
 
 Object.defineProperty(DRAW_STEEL.skills, "optgroups", {
   /** @type {FormSelectOption[]} */
-  get: function () {
+  get: function() {
     const config = ds.CONFIG.skills;
     return Object.entries(config.list).reduce((arr, [value, { label, group }]) => {
       arr.push({ label, group: config.groups[group].label, value });
@@ -1102,7 +1102,7 @@ preLocalize("abilities.forcedMovement", { key: "label" });
 
 Object.defineProperty(DRAW_STEEL.abilities.keywords, "optgroups", {
   /** @type {FormSelectOption[]} */
-  get: function () {
+  get: function() {
     const sortedKeywords = Object.entries(ds.CONFIG.abilities.keywords).sort(([keyA, valueA], [keyB, valueB]) => {
       // When no group, sort between their keys
       if ((valueA.group === undefined) && (valueB.group === undefined)) return keyA.localeCompare(keyB);
--- a/src/module/config.mjs
+++ b/src/module/config.mjs
@@ -99,30 +99,39 @@ DRAW_STEEL.speedOptions = ["teleport", "fly", "walk", "swim", "burrow", "climb"]
 DRAW_STEEL.damageTypes = {
   acid: {
     label: "DRAW_STEEL.DamageTypes.Acid",
+    color: "#14ff14",
   },
   cold: {
     label: "DRAW_STEEL.DamageTypes.Cold",
+    color: "#14ffd0",
   },
   corruption: {
     label: "DRAW_STEEL.DamageTypes.Corruption",
+    color: "#7b00a8",
   },
   fire: {
     label: "DRAW_STEEL.DamageTypes.Fire",
+    color: "#ff870f",
   },
   holy: {
     label: "DRAW_STEEL.DamageTypes.Holy",
+    color: "#ffed61",
   },
   lightning: {
     label: "DRAW_STEEL.DamageTypes.Lightning",
+    color: "yellow",
   },
   poison: {
     label: "DRAW_STEEL.DamageTypes.Poison",
+    color: "#008500",
   },
   psychic: {
     label: "DRAW_STEEL.DamageTypes.Psychic",
+    color: "#d40cc3",
   },
   sonic: {
     label: "DRAW_STEEL.DamageTypes.Sonic",
+    color: "#999",
   },
 };
 preLocalize("damageTypes", { key: "label" });
@@ -473,7 +482,7 @@ preLocalize("skills.list", { key: "label" });
 
 Object.defineProperty(DRAW_STEEL.skills, "optgroups", {
   /** @type {FormSelectOption[]} */
-  get: function() {
+  get: function () {
     const config = ds.CONFIG.skills;
     return Object.entries(config.list).reduce((arr, [value, { label, group }]) => {
       arr.push({ label, group: config.groups[group].label, value });
@@ -1093,7 +1102,7 @@ preLocalize("abilities.forcedMovement", { key: "label" });
 
 Object.defineProperty(DRAW_STEEL.abilities.keywords, "optgroups", {
   /** @type {FormSelectOption[]} */
-  get: function() {
+  get: function () {
     const sortedKeywords = Object.entries(ds.CONFIG.abilities.keywords).sort(([keyA, valueA], [keyB, valueB]) => {
       // When no group, sort between their keys
       if ((valueA.group === undefined) && (valueB.group === undefined)) return keyA.localeCompare(keyB);
--- a/src/module/data/actor/base.mjs
+++ b/src/module/data/actor/base.mjs
@@ -234,7 +234,7 @@ export default class BaseActorModel extends SubtypeModelMixin(foundry.abstract.T
    * @param {DrawSteelCombatant} combatant The combatant representation
    * @abstract
    */
-  async _onStartTurn(combatant) { }
+  async _onStartTurn(combatant) {}
 
   /**
    * Prompt the user for what types
--- a/src/module/data/actor/base.mjs
+++ b/src/module/data/actor/base.mjs
@@ -3,6 +3,7 @@ import { PowerRoll } from "../../rolls/power.mjs";
 import { damageTypes, requiredInteger, setOptions } from "../helpers.mjs";
 import SizeModel from "../models/size.mjs";
 import SubtypeModelMixin from "../subtype-model-mixin.mjs";
+import { showFloatyText } from "../../helpers/tokenUtils.mjs";
 
 /** @import { DrawSteelActor, DrawSteelCombatant, DrawSteelCombatantGroup } from "../../documents/_module.mjs"; */
 /** @import AbilityModel from "../item/ability.mjs" */
@@ -233,7 +234,7 @@ export default class BaseActorModel extends SubtypeModelMixin(foundry.abstract.T
    * @param {DrawSteelCombatant} combatant The combatant representation
    * @abstract
    */
-  async _onStartTurn(combatant) {}
+  async _onStartTurn(combatant) { }
 
   /**
    * Prompt the user for what types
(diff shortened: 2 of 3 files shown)


src/module/config.mjs
```diff
@@ -99,30 +99,39 @@ DRAW_STEEL.speedOptions = ["teleport", "fly", "walk", "swim", "burrow", "climb"]
 DRAW_STEEL.damageTypes = {
   acid: {
     label: "DRAW_STEEL.DamageTypes.Acid",
+    color: "#14ff14",
```
krbz999, line 102: These should be added to the annotation above as well.

I think we should instantiate as a `foundry.utils.Color`, too.

src/module/helpers/tokenUtils.mjs
```diff
@@ -0,0 +1,43 @@
+/**
+ * @param {DrawSteelActor} actor
+ * @param input
+ * @param fill
+ * @returns {Promise<void>}
+ */
+export async function showFloatyText(actor, input, fill) {
```
krbz999, line 7: I don't feel great about this being a utility function (and then it would be in the wrong folder anyway). The `Actor#_onUpdate` method is where we would want this, I think.

krbz999: And consider a name like `#displayStaminaChange` or such.

src/module/helpers/tokenUtils.mjs
```diff
@@ -0,0 +1,43 @@
...
+  if (!canvas.scene) {
+    return;
+  }
+
+  const [token] = actor.getActiveTokens();
```
krbz999, line 12: No reason not to display the floating text on all tokens.

src/module/helpers/tokenUtils.mjs
```diff
@@ -0,0 +1,43 @@
...
+
+  if (token && (typeof input === "number")) {
+    const gridSize = canvas.scene.grid.size;
+    const scrollingTextArgs = [
+      { x: token.x + gridSize / 2, y: token.y + gridSize - 20 },
```
krbz999, line 17: To get the center position of a token, simply use `token.center`. This current approach ignores token sizes.

src/module/helpers/tokenUtils.mjs
```diff
@@ -0,0 +1,43 @@
...
+      Math.abs(input),
+      {
+        fill: fill ?? (input < 0 ? "lightgreen" : "white"),
+        fontSize: 32,
+        stroke: 0x000000,
+        strokeThickness: 4,
+      },
+    ];
+    await token._animation;
```
krbz999, line 26: Is this necessary? What happens if we do not await this?

src/module/helpers/tokenUtils.mjs
```diff
@@ -0,0 +1,43 @@
...
+    await canvas.interface?.createScrollingText(...scrollingTextArgs);
```
krbz999, line 27: If we have a token, then the canvas is ready, and then `canvas.interface` is definitely ready, too.

krbz999 (changes requested):

src/module/data/actor/base.mjs
```diff
@@ -352,7 +387,7 @@ export default class BaseActorModel extends SubtypeModelMixin(foundry.abstract.T
     const remainingDamage = Math.max(0, damage - damageToTempStamina);
     if (remainingDamage > 0) staminaUpdates.value = this.stamina.value - remainingDamage;
 
-    return this.parent.update({ "system.stamina": staminaUpdates });
+    return this.parent.update({ "system.stamina": staminaUpdates, "damageColor": damageColor });
```
zarick1342, line 390: I don't know about this technique (passing something to the changes that isn't really system data), but otherwise I'm not sure how to fulfill the request to make the damage different colors per type.

src/module/config.mjs
zarick1342: I added the annotation but I'm not really sure the proper way to instantiate the utils as requested.

src/module/data/actor/base.mjs
krbz999: Yeah we shouldn't do that. We can pass additional information to the second parameter instead.

src/module/config.mjs
krbz999: `color: new foundry.utils.Color("...")` would do it.

This does change how you use the property throughout but the class is very handy and contains lots of information.

src/module/data/actor/base.mjs
```diff
@@ -203,6 +203,8 @@ export default class BaseActorModel extends SubtypeModelMixin(foundry.abstract.T
         },
       });
     }
+
+    this.displayStaminaChange(changes);
```
krbz999, line 207: This should be in `_onUpdate` to appear correctly for all users.

cdverrett94:
This won't currently trigger for minion damage. Would need to add something to the `SquadModel#_onUpdate` as well.

src/module/data/actor/base.mjs
JPMeehan: Yeah so the way this should be done is `this.parent.update({ "system.stamina: staminaUpdates }, { damageColor })`

zarick1342:
I've got this done for regular actors I think, but the minion bit seems tricky. I figured out how to extend this into squads to catch the stamina changes, but I'm not sure how to make it display on the minions.

From a user standpoint what would be the expected behavior when squad HP changes? Should a damage number show on all minions? The targeted minion? If the latter, how do we discover which minion the stamina change was requested on?

Also from my quick attempt to put that together it seems like the actor on the minion doesn't necessarily inherit from the same place as "normal" actors? So I was struggling figuring out how to call the `displayStaminaChanges` method there.

krbz999 (changes requested):
It's shaping up! I have some minor concerns that should be easily ammended.

cdverrett94:
> I've got this done for regular actors I think, but the minion bit seems tricky. I figured out how to extend this into squads to catch the stamina changes, but I'm not sure how to make it display on the minions.
> 
> From a user standpoint what would be the expected behavior when squad HP changes? Should a damage number show on all minions? The targeted minion? If the latter, how do we discover which minion the stamina change was requested on?
> 
> Also from my quick attempt to put that together it seems like the actor on the minion doesn't necessarily inherit from the same place as "normal" actors? So I was struggling figuring out how to call the `displayStaminaChanges` method there.

In `SquadModel#_onUpdate` you should be able to do something like this:

```suggestion
for (const combatant of this.parent.members) {
    combatant.actor?.system.displayStaminaChange(/* whatever the final arguments are */);
}
```

If you only wanted it to show on the triggering minion, you could probably pass the actor id as an option to the update method. Then just get that actor and only call `displayStaminaChange` on that actor.

zarick1342:
Made the requested changes (I think), and tried my hand at the minion update. For now it shows on all minions in the squad but that does sort of accurately represent what's happening.

The damage type also doesn't carry through even when using the button to apply damage, couldn't figure out how that works.

Also, I haven't done much work on a public Github with this kind of pull request and was trying to get the latest changes, so that's what the force push in the history was (resetting this branch to not include a bunch of changes that aren't mine), so if you can look to make sure I didn't screw anything up by doing that it'd be appreciated.

src/module/data/actor/base.mjs
```diff
@@ -232,6 +244,43 @@ export default class BaseActorModel extends SubtypeModelMixin(foundry.abstract.T
     }
   }
 
+  /**
+   * Display actor stamina changes on active tokens.
+   *
+   * @param {int} diff The amount the actor's stamina has changed.
+   * @param {string} damageType (optional) The type of damage being dealt.
```
krbz999, line 251:
```suggestion
   * @param {number} diff               The amount the actor's stamina has changed.
   * @param {string} [damageType=""]    The type of damage being dealt.
```

src/module/config.mjs
```diff
@@ -95,35 +95,44 @@ DRAW_STEEL.speedOptions = ["teleport", "fly", "walk", "swim", "burrow", "climb"]
 
 /**
  * Configuration information for damage types
- * @type {Record<string, {label: string}>}
+ * @type {Record<string, {label: string, color: string}>}
```
JPMeehan, line 98: doesn't `fromString` turn the strings into a Color instance, making this type inaccurate?

src/module/data/actor/base.mjs
```diff
@@ -352,7 +401,9 @@ export default class BaseActorModel extends SubtypeModelMixin(foundry.abstract.T
     const remainingDamage = Math.max(0, damage - damageToTempStamina);
     if (remainingDamage > 0) staminaUpdates.value = this.stamina.value - remainingDamage;
 
-    return this.parent.update({ "system.stamina": staminaUpdates });
+    options.ds ??= {};
```
JPMeehan, line 404: are you actually using the other options forwarded (type, ignoredImmunities)? Or should this be a fresh `operation` object?

zarick1342: Not sure I understand this one, can you elaborate?

JPMeehan: So, here you're reusing the`options` object for the function and passing it into the update. I am asking if this should be a fresh object instead.

JPMeehan:
Hi @zarick1342 any chance you can get to this in the next ~12 hours? Would like to push out 0.7.2 with this before the weekend

src/module/data/actor/base.mjs
```diff
@@ -232,6 +244,43 @@ export default class BaseActorModel extends SubtypeModelMixin(foundry.abstract.T
     }
   }
 
+  /**
+   * Display actor stamina changes on active tokens.
+   *
+   * @param {int} diff The amount the actor's stamina has changed.
```
krbz999, line 250: `number`, not `int`

src/module/data/actor/base.mjs
```diff
@@ -232,6 +244,43 @@ export default class BaseActorModel extends SubtypeModelMixin(foundry.abstract.T
...
+   * @param {string} [damageType=""] The type of damage being dealt.
+   */
+  async displayStaminaChange(diff, damageType = "") {
+    if (!canvas.scene) {
+      return;
+    }
```
krbz999, line 256: Should also abort if `!diff`.

src/module/data/actor/base.mjs
krbz999: Yeah it should be a new object - the `options` object that is a parameter of the `takeDamage` method should not also be used as the operation object for updates.

krbz999 (changes requested):
Only minor stuff left here imo. Looking good so far.

src/module/config.mjs
```diff
@@ -95,35 +95,44 @@ DRAW_STEEL.speedOptions = ["teleport", "fly", "walk", "swim", "burrow", "climb"]
 
 /**
  * Configuration information for damage types
- * @type {Record<string, {label: string}>}
+ * @type {Record<string, {label: string, color: Color}>}
```
JPMeehan, line 98:
```suggestion
 * @type {Record<string, {label: string, color: foundry.utils.Color}>}
```

You can validate something works in the editor by mousing over it.

JPMeehan (approved):

PR #477, opened 2025-05-13
"Other" power roll effect type

JPMeehan:
- Removed generic "text" property
- Added "name" property
- Added Other type, which has a tiered text field
- Refactored the labels so they're at least moderately i18n-friendly for now

### DIFF
--- a/src/module/data/pseudo-documents/power-roll-effects/base-power-roll-effect.mjs
+++ b/src/module/data/pseudo-documents/power-roll-effects/base-power-roll-effect.mjs
@@ -4,6 +4,9 @@ import TypedPseudoDocument from "../typed-pseudo-document.mjs";
 
 const { SchemaField, StringField } = foundry.data.fields;
 
+/**
+ * Pseudodocument used by abilities to represent the tiered results of a power roll.
+ */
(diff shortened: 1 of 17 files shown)


src/module/data/pseudo-documents/power-roll-effects/base-power-roll-effect.mjs
```diff
@@ -67,6 +71,14 @@ export default class BasePowerRollEffect extends TypedPseudoDocument {
 
   /* -------------------------------------------------- */
 
+  prepareDerivedData() {
```
krbz999, line 74: inheritdoc

src/module/data/pseudo-documents/pseudo-document.mjs
```diff
@@ -3,6 +3,9 @@
 
 const { DocumentIdField } = foundry.data.fields;
 
+/**
+ * A special subclass of data model that can be treated as a system-defined embedded document
```
krbz999, line 7:
```suggestion
 * A special subclass of data model that can be treated as a system-defined embedded document.
```

krbz999 (approved):

PR #480, opened 2025-05-14
Applied Power Roll Effect

### LINKED ISSUE #401: NPC Potency Not Calculating With Formula
When using @potency.weak/average/strong, it shows as a 0 on the ability/chat card. Using a number shows it correctly.

Image

Tier 1 was set with @potency/weak, tier 2/3 were set with using a number instead.

Foundry V13 Testing 4 (339)
Draw Steel v0.7
Modules: None

### LINKED ISSUE #463: Define AE PowerRollEffect type
0.6 definition for reference

```ts
type AEPowerRoll = {
    type: "ae";
    always: Set<string>;
    success: Set<string>;
    failure: Set<string>;
    potency: Potency;
    display: string;
  }
```

JPMeehan:
image

image

image

Closes #463

Closes #401

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -1043,11 +1043,11 @@
           },
           "success": {
             "label": "Success",
-            "hint": "Effects that apply when the target's characteristic is equal to or greater than the potency value"
+            "hint": "Effects that apply when the target's characteristic is equal to or greater than the potency value."
           },
           "failure": {
             "label": "Failure",
-            "hint": "Effects that apply when the target's characteristic is less than the potency value"
+            "hint": "Effects that apply when the target's characteristic is less than the potency value."
           },
           "potency": {
             "label": "Potency",
--- a/lang/en.json
+++ b/lang/en.json
@@ -1051,9 +1051,6 @@
           },
           "potency": {
             "label": "Potency",
-            "enabled": {
-              "label": "Has Potency"
-            },
             "value": {
               "label": "Potency Value"
             },
--- a/lang/en.json
+++ b/lang/en.json
@@ -1033,9 +1033,37 @@
           },
           "text": {
             "label": "Text"
+          },
+          "display": {
+            "label": "Display",
+            "hint": "'{{potency}}' will be replaced with the derived potency info (i.e. M<2)."
+          },
+          "always": {
+            "label": "Always"
+          },
+          "success": {
+            "label": "Success",
+            "hint": "Effects that apply when the target's characteristic is equal to or greater than the potency value"
+          },
+          "failure": {
+            "label": "Failure",
+            "hint": "Effects that apply when the target's characteristic is less than the potency value"
+          },
+          "potency": {
+            "label": "Potency",
+            "enabled": {
+              "label": "Has Potency"
+            },
+            "value": {
+              "label": "Potency Value"
+            },
+            "characteristic": {
+              "label": "Potency Characteristic"
+            }
           }
         },
         "TYPES": {
+          "ae": "Applied Effect",
           "damage": "Damage",
           "other": "Other"
         },
--- a/lang/en.json
+++ b/lang/en.json
@@ -1039,7 +1039,8 @@
             "hint": "'{{potency}}' will be replaced with the derived potency info (i.e. M<2)."
           },
           "always": {
-            "label": "Always"
+            "label": "Always",
+            "hint": "Effects that apply regardless of potency checks."
           },
           "success": {
             "label": "Success",
--- a/lang/en.json
+++ b/lang/en.json
@@ -1042,26 +1042,27 @@
             "label": "Always",
             "hint": "Effects that apply regardless of potency checks."
           },
-          "success": {
-            "label": "Success",
-            "hint": "Effects that apply when the target's characteristic is equal to or greater than the potency value."
-          },
-          "failure": {
-            "label": "Failure",
-            "hint": "Effects that apply when the target's characteristic is less than the potency value."
-          },
           "potency": {
             "label": "Potency",
             "value": {
               "label": "Potency Value"
             },
             "characteristic": {
-              "label": "Potency Characteristic"
+              "label": "Potency Characteristic",
+              "hint": "Tiers 2 and 3 will default to the chosen characteristic of the prior tier."
+            },
+            "failure": {
+              "label": "Failure",
+              "hint": "Effects that apply when the target's characteristic is less than the potency value."
+            },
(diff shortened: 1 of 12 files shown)


lang/en.json
```diff
@@ -1033,9 +1033,34 @@
...
+          "display": {
+            "label": "Display",
+            "hint": "'{{potency}}' will be replaced with the derived potency info (i.e. M<2)."
+          },
+          "always": {
+            "label": "Always"
+          },
+          "success": {
+            "label": "Success",
+            "hint": "Effects that apply when the target's characteristic is equal to or greater than the potency value"
```
krbz999, line 1046:
```suggestion
            "hint": "Effects that apply when the target's characteristic is equal to or greater than the potency value."
```

lang/en.json
```diff
@@ -1033,9 +1033,34 @@
...
+          },
+          "failure": {
+            "label": "Failure",
+            "hint": "Effects that apply when the target's characteristic is less than the potency value"
```
krbz999, line 1050:
```suggestion
            "hint": "Effects that apply when the target's characteristic is less than the potency value."
```

src/module/config.mjs
```diff
@@ -1124,6 +1124,10 @@ DRAW_STEEL.PowerRollEffect = {
     label: "DRAW_STEEL.PSEUDO.POWER_ROLL_EFFECT.TYPES.damage",
     documentClass: pseudoDocuments.powerRollEffects.DamagePowerRollEffect,
   },
+  ae: {
+    label: "DRAW_STEEL.PSEUDO.POWER_ROLL_EFFECT.TYPES.ae",
+    documentClass: pseudoDocuments.powerRollEffects.AppliedPowerRollEffect,
+  },
```
krbz999, line 1130: Oh no please, let us pick a better internal key than "ae".

"effect"? "grantEffect"? Anything but a two-letter abbreviation.

src/module/data/pseudo-documents/power-roll-effects/applied-effect.mjs
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  /* -------------------------------------------------- */
+
+  /** @inheritdoc */
+  prepareDerivedData() {
+    super.prepareDerivedData();
+
+    for (const n of [1, 2, 3]) {
+      this.ae[`tier${n}`].potency.value ||= this.schema.getField(["ae", `tier${n}`, "potency", "value"]).initial;
```
krbz999, line 58: We should set `display` as well.

src/module/data/pseudo-documents/power-roll-effects/applied-effect.mjs
```diff
@@ -0,0 +1,129 @@
...
+            value: this.ae[`tier${n}`].potency.characteristic,
+            src: this._source.ae[`tier${n}`].potency.characteristic,
+            name: `${path}.potency.characteristic`,
+            options: Object.entries(ds.CONFIG.characteristics).map(([value, { label }]) => ({ value, label })),
+          },
+        },
+      };
+    }
+  }
+  /* -------------------------------------------------- */
```
krbz999, line 113:
```suggestion
  }

  /* -------------------------------------------------- */
```

krbz999 (changes requested):
Code looks good but I feel strongly about changing `ae`.

We should set the tier 2 + tier 3 (other text and ae display) as derived from the lower tier. Are there cases where a higher tier can be blank while the lower tier is not?

cdverrett94:
Should the always and success/failure effects have conditions set to only show the field based on if there is a potency characteristic set? 

Or should the always effect field have a hint to indicate it's only for potency-less effects.

JPMeehan:
I think always should probably always be shown, but I would be open to moving the success/failure within potency. I think a blank potency characteristic is a good indicator for "not in use".

src/module/config.mjs
JPMeehan: `applied` is also an option, not thrilled by how often we're being pushed to reuse the word "effect"

src/module/data/pseudo-documents/power-roll-effects/applied-effect.mjs
JPMeehan: I think this one is a good case for bubbling up rather than using field initials.

cdverrett94:
That's fair. Then maybe just a hint under always to highlight it's for AEs that aren't based on a potency? Just so it's clear what it's for.

src/module/config.mjs
```diff
@@ -1124,6 +1124,10 @@ DRAW_STEEL.PowerRollEffect = {
     label: "DRAW_STEEL.PSEUDO.POWER_ROLL_EFFECT.TYPES.damage",
     documentClass: pseudoDocuments.powerRollEffects.DamagePowerRollEffect,
   },
+  applied: {
+    label: "DRAW_STEEL.PSEUDO.POWER_ROLL_EFFECT.TYPES.ae",
```
krbz999, line 1128:
```suggestion
    label: "DRAW_STEEL.PSEUDO.POWER_ROLL_EFFECT.TYPES.applied",
```

krbz999 (approved):
Looks good. Let's add a "Default" option to the characteristic as well. Purpose being to default to the lower tier's characteristic so users only have to set the value in one tier.

krbz999:
> I think always should probably always be shown, but I would be open to moving the success/failure within potency. I think a blank potency characteristic is a good indicator for "not in use".

Yeah that sounds good. One option for each chr, a Default (same as lower tier), and a "None" option.

JPMeehan:
> > I think always should probably always be shown, but I would be open to moving the success/failure within potency. I think a blank potency characteristic is a good indicator for "not in use".
> 
> Yeah that sounds good. One option for each chr, a Default (same as lower tier), and a "None" option.

Value-wise, I think `''` as none and `null` as default? Need to check how the form data is processed to see if that works.

krbz999:
> > > I think always should probably always be shown, but I would be open to moving the success/failure within potency. I think a blank potency characteristic is a good indicator for "not in use".
> > 
> > 
> > Yeah that sounds good. One option for each chr, a Default (same as lower tier), and a "None" option.
> 
> Value-wise, I think `''` as none and `null` as default? Need to check how the form data is processed to see if that works.

Seems tricky. I think blank and "none" are fine.

JPMeehan:
Added in blank as default, "none" as the un-option. Default is only available for tiers 2 & 3.

image

PR #483, opened 2025-05-15
Add properties field to damage and forced PREs

cdverrett94:
Added a properties field to the damage (Ignores Immunities) and force (Vertical, Ignores Stability) PREs.

Also updated some inconsistencies in the type for `SetField`s and in where multi-select value/label options were being created.

### DIFF
--- a/src/module/config.mjs
+++ b/src/module/config.mjs
@@ -1121,12 +1121,17 @@ Object.defineProperty(DRAW_STEEL.abilities.keywords, "optgroups", {
 
 /**
  * Valid types for the PowerRollEffect pseudo-document
- * @type {Record<string, { label: string; documentClass: pseudoDocuments.powerRollEffects.BasePowerRollEffect }>}
+ * @type {Record<string, { label: string; documentClass: pseudoDocuments.powerRollEffects.BasePowerRollEffect, properties?: Record<string, { label: string; }> }>}
  */
 DRAW_STEEL.PowerRollEffect = {
   damage: {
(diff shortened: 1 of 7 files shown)


src/module/config.mjs
```diff
@@ -1121,12 +1121,17 @@ Object.defineProperty(DRAW_STEEL.abilities.keywords, "optgroups", {
 
 /**
  * Valid types for the PowerRollEffect pseudo-document
- * @type {Record<string, { label: string; documentClass: pseudoDocuments.powerRollEffects.BasePowerRollEffect }>}
+ * @type {Record<string, { label: string; documentClass: pseudoDocuments.powerRollEffects.BasePowerRollEffect, properties?: Record<string, { label: string; }> }>}
```
JPMeehan, line 1124: I think we could break this object out into a typedef/interface

JPMeehan (approved):
Looks good, clean work

src/module/config.mjs
cdverrett94: Is this what the typedef should look like?

```
/**
 * @typedef {Object} PowerRollEffectProperty
 * @property {string} label
 */

/**
 * @typedef {Object} PowerRollEffectType
 * @property {string} label
 * @property {pseudoDocuments.powerRollEffects.BasePowerRollEffect} documentClass
 * @property {Record<string, PowerRollEffectProperty} [properties]
 */
```

JPMeehan: Yeah, although I actually don't like the implicit extending Object and would just leave that out.

cdverrett94: Dope. Updated.

PR #484, opened 2025-05-15
Split effect into before & after

### LINKED ISSUE #478: Split effects into before & after
Abilities like "Their Lack of Focus Is Their Undoing" can have an Effect _before_ the Power Roll, and abilities like "That One Is Mine!" can have an Effect before _and_ after. Data-wise this should just be `effect.before` and `effect.after`, but a solution for the sheet should also be found.

JPMeehan:
- Also found some display bugs in the description tab, I suspect from CSS changes by core between early v13 builds and stable
- Widened the sheet a bit so the prosemirror headers don't overflow by default when selecting a p element

Closes #478

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,7 +54,7 @@
   - Power roll, spend, and effects have been moved to the new Impact tab
   - Power roll effects are now organized by type first (e.g. Damage)
     - Each type has data for each tier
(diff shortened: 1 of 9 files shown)


krbz999 (changes requested):
The editors are nonfunctional in my testing.

krbz999 (approved):

PR #487, opened 2025-05-16
Have potency apply to all effect types

### LINKED ISSUE #486: Damage PREs need potency
I also think other effects can have a potency. Is there a way to make potency characteristic/value included in the base PRE stuff instead of having to define it for every type? With AE having extra schema fields, not sure if there's a way to add extra fields easily to a schema field.

cdverrett94:
Closes #486

- Move the potency schema inside the `BasePowerRollEffect.duplicateTierSchema` so that potency auto applies to all effects.
   - For this I had to move the applied effects success/failure to the top level of the applied effects schema.
- To make the above easier I used `this.constructor.TYPE` a lot. This meant in the other effect schema I renamed `text` to `other` and `value` to `text`. This has the benefit of being consistent with the top level being the type. 
   - Should `text` be display instead? The applied and forced use display instead of text.
   - Should damage get a display field too? Right now, I'm inferring that potency should be first, but with a display field, it'd be configurable.
- With all of the effects now having a potency, should there just be one potency section in the HTML instead of inside each `{{#with}}`?

### DIFF
--- a/src/module/data/item/ability.mjs
+++ b/src/module/data/item/ability.mjs
@@ -277,7 +277,7 @@ export default class AbilityModel extends BaseItemModel {
 
     context.powerRollEffects = Object.fromEntries([1, 2, 3].map(tier => [
       `tier${tier}`,
-      { text: powerRollEffectFormatter.format(this.power.effects.contents.map(effect => effect.toText(tier))) },
+      { text: this.power.effects.contents.map(effect => effect.toText(tier)).join("; ") },
     ]));
     context.powerRolls = this.power.effects.size > 0;
 
--- a/src/module/data/item/ability.mjs
+++ b/src/module/data/item/ability.mjs
@@ -273,8 +273,6 @@ export default class AbilityModel extends BaseItemModel {
 
     context.characteristics = Object.entries(ds.CONFIG.characteristics).map(([value, { label }]) => ({ value, label }));
 
-    const powerRollEffectFormatter = game.i18n.getListFormatter({ type: "unit" });
-
     context.powerRollEffects = Object.fromEntries([1, 2, 3].map(tier => [
       `tier${tier}`,
       { text: this.power.effects.contents.map(effect => effect.toText(tier)).join("; ") },
--- a/templates/sheets/pseudo-documents/power-roll-effect-sheet/details-tiers.hbs
+++ b/templates/sheets/pseudo-documents/power-roll-effect-sheet/details-tiers.hbs
@@ -3,6 +3,20 @@
   {{formGroup value.field value=value.src name=value.name placeholder=value.placeholder localize=true}}
   {{formGroup types.field value=types.src name=types.name options=@root.fields.damageTypes localize=true}}
   {{formGroup properties.field value=properties.src name=properties.name options=@root.fields.properties localize=true}}
+  {{#with potency}}
+  <fieldset>
+    <legend>{{localize field.label}}</legend>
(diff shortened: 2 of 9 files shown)


JPMeehan:
1. Definitely just switch `text` to `display` - now that `other` has potency, the hint is applicable.
2. I think the most reusable method for the potency bit is an inline partial at the top

templates/sheets/pseudo-documents/power-roll-effect-sheet/details-tiers.hbs
```diff
@@ -19,8 +33,10 @@
     localize=true
     }}
     {{formGroup value.field value=value.src name=value.name placeholder=value.field.initial localize=true}}
-    {{formGroup failure.field value=failure.src name=failure.name options=../effectOptions localize=true}}
-    {{formGroup success.field value=success.src name=success.name options=../effectOptions localize=true}}
+    {{#with ../../tierFields.applied}}
+    {{formGroup failure.field value=failure.src name=failure.name options=../../effectOptions localize=true}}
+    {{formGroup success.field value=success.src name=success.name options=../../effectOptions localize=true}}
+    {{/with}}
```
krbz999, line 39: Would prefer a `#with @root.some.longer.path.tierFields.applied` over a relative path.

cdverrett94:
> 1. Definitely just switch `text` to `display` - now that `other` has potency, the hint is applicable.
> 
>     2. I think the most reusable method for the potency bit is an inline partial at the top

One #2, I moved to the inline partial, but I'm not sure if you approve of the way I did it. I was trying to prevent having to do the fieldset multiple times while still allowing the applied effects stuff to add additional fields inside it.

JPMeehan:
> > 1. Definitely just switch `text` to `display` - now that `other` has potency, the hint is applicable.
> >    
> >    2. I think the most reusable method for the potency bit is an inline partial at the top
> 
> One #2, I moved to the inline partial, but I'm not sure if you approve of the way I did it. I was trying to prevent having to do the fieldset multiple times while still allowing the applied effects stuff to add additional fields inside it. It also prevented having to do the log path for the success failure fields per Zhell's review.

I think the partial block will get you there?

JPMeehan:
There's some deep magic in Handlebars that I think goes under-appreciated by the average Foundry developer

JPMeehan (approved):
Looks good, could you also do the formatter => join with semicolon adjustment?

cdverrett94:
Yeah, I'll go ahead and do that. While I'm here, thoughts on adding a display field to damage that I mentioned in the OP?

JPMeehan:
I think for now damage doesn't need display, if we need to adjust we can without too many problems

cdverrett94:
Fair enough!

src/module/data/item/ability.mjs
```diff
@@ -277,7 +277,7 @@ export default class AbilityModel extends BaseItemModel {
 
     context.powerRollEffects = Object.fromEntries([1, 2, 3].map(tier => [
       `tier${tier}`,
-      { text: powerRollEffectFormatter.format(this.power.effects.contents.map(effect => effect.toText(tier))) },
+      { text: this.power.effects.contents.map(effect => effect.toText(tier)).join("; ") },
```
JPMeehan, line 280: Formatter is probably unused now and no longer needs to be declared?

cdverrett94: ahh oops

PR #488, opened 2025-05-16
Fixed typo in build, enabled manual wiki run

JPMeehan:
Unfortunately it's annoying to test these build changes, the typo fix is real idk if this actually enables manual wiki runs but I think it will


krbz999 (approved):
Here's hoping 🤞

PR #489, opened 2025-05-16
Move mode toggle and update source buttons to header

### LINKED ISSUE #413: Consider moving Edit toggle and Source config into header
Something I've done in Artichron is add these buttons in `renderFrame` - here pictured a button for the edit toggle and a button for "opacity" toggle (adds a class which reduces opacity to 0.1 unless hovered).

Adding header buttons is something of a fashion faux pas but I think systems should be free to do so when it exists for all sheets, ensuring this shared functionality can always be found in the same location.

Image

cdverrett94:
Closes #413 

Screenshot 2025-05-16 112438
Screenshot 2025-05-16 112433
Screenshot 2025-05-16 112101


JPMeehan (approved):
Works well. You can also delete... basically the entirety of our `elements` stuff as that was all copied from dnd5e just for the mode slide toggle. If we want to bring it back later we can, but it's no longer in use and honestly I like the smaller profile there.

cdverrett94:
I think I've got everything related to the slide toggle element removed.

PR #499, opened 2025-05-18
Automate edges from flanking

### LINKED ISSUE #451: Automate Flanking
Flanking occurs on melee strikes when both entities are adjacent and on opposite sides/corners of the target. 

Flanking requires line of effect and the ability to take triggered actions. The triggered actions part can make use of the existing `system.restrictions.type`.

cdverrett94:
Closes #451 

- Automates flanking checks. To flank, you need to be adjacent, have line of effect, be able to take triggered actions, and be on opposite sides/corners.
- Adds `DrawSteelToken#isFlanking` with several helper methods.
   - Most notably is `hasLineOfEffect` that checks each corner of each occupied space of the token with  each corner of each occupied space of the target. This can also be used for automating the taunted condition later.
 - Added an input to the Wall Config for if a wall should block LoE. In `hasLineOfEffect`, verify that each wall collision blocks LoE. This defaults to true

<details closed>
<summary>Wall Config image</summary>

</details>

### DIFF
--- a/src/module/canvas/placeables/token.mjs
+++ b/src/module/canvas/placeables/token.mjs
@@ -22,7 +22,7 @@ export default class DrawSteelToken extends foundry.canvas.placeables.Token {
   /* -------------------------------------------------- */
 
   /**
-   * Does the actor have restrictions preventing flanking ability
+   * Can the token's actor flank? Effects that restrict the ability to take triggered actions block flanking.
    * @type {boolean}
    */
   get canFlank() {
@@ -30,13 +30,17 @@ export default class DrawSteelToken extends foundry.canvas.placeables.Token {
     return !this.actor.system.restrictions.type.has("triggered");
   }
 
+  /* -------------------------------------------------- */
+
   /**
-   * Can the actor be flanked?
+   * Can the token's actor be flanked?
    * @type {boolean}
    */
   get canBeFlanked() {
     return this.actor.system.flankable ?? true;
   }
+
+  /* -------------------------------------------------- */
 
   /**
    * An array of inset vertices for each of the token's grid spaces.
@@ -60,6 +64,8 @@ export default class DrawSteelToken extends foundry.canvas.placeables.Token {
     return vertices;
   }
 
+  /* -------------------------------------------------- */
+
   /**
    * Determine if any of the token's occupied squares are adjacent to any of the target's occupied squares.
    * @param {DrawSteelToken} target
@@ -79,6 +85,8 @@ export default class DrawSteelToken extends foundry.canvas.placeables.Token {
     return false;
   }
 
+  /* -------------------------------------------------- */
+
   /**
    * Return all allies adjacent to the target with line of effect
    * @param {*} target
@@ -95,6 +103,8 @@ export default class DrawSteelToken extends foundry.canvas.placeables.Token {
 
     return alliedTokens;
   }
+
+  /* -------------------------------------------------- */
 
   /**
    * Does this token have line of effect to the target token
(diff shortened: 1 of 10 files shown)


src/module/canvas/placeables/token.mjs
```diff
@@ -16,6 +17,155 @@ export default class DrawSteelToken extends foundry.canvas.placeables.Token {
...
+  /*   Flanking                                    */
+  /* -------------------------------------------------- */
+
+  /**
+   * Can the token's actor flank? Effects that restrict the ability to take triggered actions block flanking.
+   * @type {boolean}
+   */
+  get canFlank() {
+    // Checking if active effects have restricted triggered abilities
+    return !this.actor.system.restrictions.type.has("triggered");
```
JPMeehan, line 30: Actor can be null if the base actor is deleted while the token is on the scene; need to add handling throughout

src/module/canvas/placeables/token.mjs
```diff
@@ -16,6 +17,155 @@ export default class DrawSteelToken extends foundry.canvas.placeables.Token {
...
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Does this token have line of effect to the target token
+   * @param {DrawSteelToken} token    The token that is moving.
+   * @returns {boolean}
+   */
+  hasLineOfEffect(target) {
+    // TODO: Add flag to walls for "blocks line of effect" that defaults to true.
```
JPMeehan, line 115: I think the solution here is to extend either the wall document or placeable with a getter that can handle the nullish check

JPMeehan (commented):
Still need to actually test but some initial static analysis

src/module/canvas/placeables/token.mjs
cdverrett94: For these, should the default if there's no actor be to return true?

src/module/canvas/placeables/token.mjs
cdverrett94: To add the field to the sheet, is subclassing the `WallConfig` preferred or adding via hook?

JPMeehan: I guess subclassing with use of _onRender to add the input

src/module/canvas/placeables/token.mjs
JPMeehan: I guess?

JPMeehan (commented):
OK, so the 1x1 actor case seems well-covered. But I'm struggling for actors with larger sizes, both in the "is flanking" and "is being flanked" cases.

cdverrett94:
I'll take a look tomorrow and see what's up.

cdverrett94:
Okay, I think the different sizes are all working properly now.

JPMeehan:
Larger actors being flanked appears to work but struggling to get them *to* flank. Testing with huge (3x3)

JPMeehan (approved):

src/module/documents/wall.mjs
```diff
@@ -0,0 +1,8 @@
+/**
+ * A document subclass adding system-specific behavior and registered in CONFIG.Wall.documentClass
+ */
+export default class DrawSteelWallDocument extends foundry.documents.WallDocument {
+  get blocksLineOfEffect() {
+    return foundry.utils.getProperty(this.flags, "draw-steel.blocksLineOfEffect") ?? true;
```
JPMeehan, line 6: wait why isn't this just a getFlag?

cdverrett94: Honestly? Totally forgot that was a thing. Good call.

PR #501, opened 2025-05-18
Fix setting power roll characteristic for negative values

cdverrett94:
Set the initial value to the minimum score of -5 so that any score of -5 to +5 will set the value/key.


JPMeehan (approved):
Can you also setup the changelog here?

cdverrett94:
Done!

PR #523, opened 2025-05-22
Improved error handling in formula evaluation

### LINKED ISSUE #500: Ability becomes unviewable/uneditable when forced movement tag is removed
I ran into this bug while trying to set an ability that has forced movement in Tier 2 and 3 but not 1.

When I clicked on the `X` to remove the default `pull` key word for the forced movement in Tier 1, the ability window disappeared, and trying to bring it up again from the character sheet did not work. Reloading the page didn't work, and according to `Find the Culprit` the error was with the system and not a module.

The console displayed the following error when the ability window became unavailable:

Image

After the ability can't be displayed or edited anymore, it still can be rolled for and works fine that way.

I also tested if this only happens when changing the forced movement this way on an ability on a character sheet, or if it also happens to an ability in the item tab, and it only happens to actor abilities.

Upon further investigation, if I pull the ability from the actor sheet into the items menu, it can be viewed/edited again. If I deleted that ability from the actor sheet and replace it with the ability from the item tab, then it is not viewable/editable again, so it is somehow actor sheet related?

JPMeehan:
Closes #500


src/module/data/pseudo-documents/power-roll-effects/forced-movement-effect.mjs
```diff
@@ -111,7 +111,13 @@ export default class ForcedMovementPowerRollEffect extends BasePowerRollEffect {
     const tierValue = this.forced[`tier${tier}`];
     let distanceValue = tierValue.distance;
     if (this.actor) {
-      distanceValue = new DSRoll(distanceValue, this.item.getRollData()).evaluateSync({ strict: false }).total;
+      try {
+        const evaluatedDistance = new DSRoll(distanceValue, this.item.getRollData()).evaluateSync().total;
```
krbz999, line 115: Is the error due to the formula, `distanceValue`, being malformed?

If so, perhaps let's introduce a utility method for these kinds of formula field evaluations, one that covers these edge cases.

JPMeehan: So there's three errors this can catch
1. Empty string
2. Invalid formula ("foobar")
3. Nondeterministic formula (which could be alleviated with returning strict: false)

I agree a utility method might be justified

krbz999: sgtm, make it so.

JPMeehan: I also realized we forgot to flag the relevant fields as deterministic in the FormulaField, so that should help a bit.

src/module/utils/evaluateUserFormula.mjs
```diff
@@ -0,0 +1,24 @@
...
+ * User-input formulas may throw if blank or otherwise contain invalid terms
+ * @param {string} formula                    The roll formula. May be blank or otherwise invalid
+ * @param {object} [rollData]                 The roll data for parsing
+ * @param {object} [options]                  Options for this method to forward
+ * @param {boolean} [options.strict]          Forwarded to {@linkcode DSRoll.evaluateSync}
+ * @param {boolean} [options.allowStrings]    Forwarded to {@linkcode DSRoll.evaluateSync}
+ * @param {string} [options.contextName]      Helpful string put into the error message
+ * @returns {string | number} Returns the original formula if it failed to evaluate, or the total if it succeeded
+ */
+export default function evaluateUserFormula(formula, rollData = {}, { strict, allowStrings, contextName = "unknown" } = {}) {
```
krbz999, line 14: I think we can safely default to `strict=false`.

Not sold on the name here for the function. What about just `evaluateFormula`? `deriveTotal`?

Also see no reason we cannot simply return `0` if it throws an error.

krbz999: let's do that, then. I think that clarifies what its purpose is.

JPMeehan: Also decided to normalize our file names in the utils folder to use kebab case

krbz999 (approved):

PR #534, opened 2025-05-24
Document update

KJTailor:
Updated Roll data and the system automations page, and wrote the combat encounters document. Please let me know if you want anything else included in the combat document.


src/docs/System Automations.md
```diff
@@ -77,6 +77,12 @@ With the implementation of the token ruler in Foundry V13 the system automatical
 
 Choosing one of the displayed movement types, such as climbing, crawling, or swimming automatically adjusts the available movement distance.
 
+## Flanking
```
cdverrett94, line 80: Flanking is already included at the bottom of the document. If it's being added here, then the other one should be removed.

PR #542, opened 2025-05-28
Prevent characteristic scores from becoming `null`

cdverrett94:
If a characteristic was `0`, changing other values on the sheet would make the characteristic scores `null` because a score of 0 makes the input empty.


JPMeehan (approved):
Could you also add a changelog entry

JPMeehan (approved):

PR #567, opened 2025-06-08
Improved Combat UI

### LINKED ISSUE #532: Intiative Value reset at new round
If an actor is set in the initiative tracker to have multiple initiative turns with an initiative value of >1 (e.g. Solos with 2 turns per round), this initiative value is reset to standard=1 at the beginning of a new round.

Either, once a initiative value >1 is set, it is not reset to 1 at the beginning of a new round, and the set value is restored at the beginning of each round, or there should be the option somewhere in the token settings to set the standard initiative value for an actor's token in combat

JPMeehan:
image

image

image

Closes #532

### DIFF
--- a/src/module/applications/apps/actor-combat-stats-input.mjs
+++ b/src/module/applications/apps/actor-combat-stats-input.mjs
@@ -6,6 +6,15 @@ import DocumentInput from "../api/document-input.mjs";
  */
 export default class ActorCombatStatsInput extends DocumentInput {
   /** @inheritdoc */
+  static DEFAULT_OPTIONS = {
+    classes: ["actor-combat"],
+    window: {
+      title: "DRAW_STEEL.Actor.base.NicheCombatDialog.Title",
+      icon: "fa-solid fa-swords",
+    },
+  };
+
+  /** @inheritdoc */
   static PARTS = {
     body: {
       template: systemPath("templates/sheets/document-input/actor-combat-stats-input.hbs"),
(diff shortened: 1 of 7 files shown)


src/module/applications/api/document-input.mjs
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  /* -------------------------------------------------- */
+
+  /**
+   * A function that assembles the HTML content to display.
+   * It is acceptable for `this` to be bound to something other than the DocumentInput sheet.
+   * @type {() => Promise<string>}
+   */
+  get contentFunc() {
```
krbz999, line 46: If it is required, then we should enforce it by throwing an error in the constructor.

krbz999 (changes requested):
Gonna mainly offer my thoughts on the overarching direction here. I get wanting to optimize. And I think the solution proposed here, to only have a single, easily re-used application with barely any handlebars, is pretty genius.

However, I think we'll hit eventually hit a point where it won't be as reusable as we hope, and I would lean towards a low-brow solution of a single, basic `DocumentConfig` application that is extended with subclasses providing handlebars templates.

Ignoring the LoCs that would be the small bit of boilerplate for the subclasses, I think it'd be a net reduction in lines of code as there's be much less context preparation to do; it would simply provide `config`, `document`, `fields`, and `systemFields`, and the handlebars would be trivial´.

krbz999 (approved):
Yeah that really simplifies things.

PR #587, opened 2025-06-12
Made DSID read only, added top-level getter

### LINKED ISSUE #389: Create DSID getter on Item
Should be the one that people "actually" access, with `system._dsid` being the data but not what code should be using/filtering for. We should also consider extending Item-related collections to make this an easier getter.

Inspired by dnd5e's `Item#identifier` getter

```js
  /**
   * Return an item's identifier.
   * @type {string}
   */
  get identifier() {
    if ( this.system.identifier ) return this.system.identifier;
    const identifier = this.name.replaceAll(/(\w+)([\\|/])(\w+)/g, "$1-$3");
    return identifier.slugify({ strict: true });
  }
```

JPMeehan:
I guess I could've let this linger as a "good first issue" but w/e it was a fast API improvement

Closes #389

### DIFF
--- a/src/module/data/item/base.mjs
+++ b/src/module/data/item/base.mjs
@@ -35,7 +35,7 @@ export default class BaseItemModel extends SubtypeModelMixin(foundry.abstract.Ty
 
     /**
      * The Draw Steel ID, indicating a unique game rules element
-     * @remarks readonly: true makes this non-iterable
+     * @remarks `readonly: true` makes this non-iterable
      */
     schema._dsid = new fields.StringField({ blank: false, readonly: true });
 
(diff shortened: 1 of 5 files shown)


src/module/data/item/base.mjs
```diff
@@ -35,8 +35,9 @@ export default class BaseItemModel extends SubtypeModelMixin(foundry.abstract.Ty
 
     /**
      * The Draw Steel ID, indicating a unique game rules element
+     * @remarks readonly: true makes this non-iterable
```
krbz999, line 38:
```suggestion
     * @remarks `readonly: true` makes this non-iterable
```

krbz999 (approved):

PR #600, opened 2025-06-15
Add equipment stacking

### LINKED ISSUE #558: [Suggestion] Make consumables stackable
It would be nice if consumables could be counted as a stack, which displays how many of that one item are in the inventory, rather than having the same item X times in the inventory separately.

cdverrett94:
Closes #558 

- Adds `quantity` to the `EquipmentModel`.
- Adds a quantity column to the equipment item lists with -/+ buttons for easy decrease/increase.
- Updated the crafting project method to use the new quantity field instead of creating a new item for each yield. 
   - It attempts to find an existing item by dsid and increases that quantity instead
   - Incidentally fixes an issue with project rolls using the old `prompt` method instead of the new `create` method.

equipment-stacking

### DIFF
--- a/src/module/data/item/equipment.mjs
+++ b/src/module/data/item/equipment.mjs
@@ -54,16 +54,6 @@ export default class EquipmentModel extends BaseItemModel {
   }
 
   /** @inheritdoc */
-  _onUpdate(changed, options, userId) {
-    if (this.quantity === 0) {
-      this.parent.delete();
-      return;
-    }
-
-    super._onUpdate(changed, options, userId);
-  }
-
-  /** @inheritdoc */
   async getSheetContext(context) {
     context.categories = Object.entries(ds.CONFIG.equipment.categories).map(([value, { label }]) => ({ value, label }));
 
--- a/src/module/data/item/equipment.mjs
+++ b/src/module/data/item/equipment.mjs
@@ -1,6 +1,6 @@
 import { systemPath } from "../../constants.mjs";
 import FormulaField from "../fields/formula-field.mjs";
-import { setOptions } from "../helpers.mjs";
+import { requiredInteger, setOptions } from "../helpers.mjs";
 import BaseItemModel from "./base.mjs";
 
 /** @import { DrawSteelActor } from "../../documents/_module.mjs"; */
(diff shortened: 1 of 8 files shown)


src/module/data/item/equipment.mjs
```diff
@@ -51,6 +53,16 @@ export default class EquipmentModel extends BaseItemModel {
     return schema;
   }
 
+  /** @inheritdoc */
+  _onUpdate(changed, options, userId) {
+    if (this.quantity === 0) {
```
JPMeehan, line 58: I am moderately concerned about "auto-deletion" here, maybe put this behind a setting? Or maybe use deleteDialog() instead of direct delete()?

src/module/data/item/project.mjs
```diff
@@ -253,11 +253,19 @@ export default class ProjectModel extends BaseItemModel {
...
     const yieldRoll = await new DSRoll(this.yield.amount).evaluate();
     const amount = yieldRoll.total;
-    const itemArray = Array(amount).fill(item.toObject());
 
-    await this.actor.createEmbeddedDocuments("Item", itemArray);
+    // If there's an existing item, add the amount to the item's quantity, otherwise create a new item with the quantity amount
+    if (existingItem) {
+      await existingItem.update({ "system.quantity": existingItem.system.quantity + amount });
+    } else {
+      const itemData = item.toObject();
```
JPMeehan, line 264: I believe that since this is going to frequently pull from a compendium we should be using `game.items.fromCompendium(item)` instead of `item.toObject()` to properly populate the _stats field.

src/module/data/item/project.mjs
```diff
@@ -253,11 +253,19 @@ export default class ProjectModel extends BaseItemModel {
...
     const item = await fromUuid(this.yield.item);
+    const existingItem = this.actor.items.find(i => i.dsid === item.dsid);
     const yieldRoll = await new DSRoll(this.yield.amount).evaluate();
     const amount = yieldRoll.total;
-    const itemArray = Array(amount).fill(item.toObject());
 
-    await this.actor.createEmbeddedDocuments("Item", itemArray);
+    // If there's an existing item, add the amount to the item's quantity, otherwise create a new item with the quantity amount
+    if (existingItem) {
+      await existingItem.update({ "system.quantity": existingItem.system.quantity + amount });
```
JPMeehan, line 262: I wonder if for UX reasons this should pop a notification?

JPMeehan: Ah yeah that's probably sufficient

JPMeehan (approved):

PR #602, opened 2025-06-15
Advancements: Class item advancement tab

krbz999:
image

A simple implementation, each advancement rendered in a standard form. Using the new standard API to create, edit, and delete.

### DIFF
--- a/templates/item/advancement.hbs
+++ b/templates/item/advancement.hbs
@@ -1,6 +1,4 @@
 <section class="tab standard-form {{tab.cssClass}}" data-group="primary" data-tab="advancement" data-subtype="{{document.type}}">
-  {{log this level="warn"}}
-
   <section class="advancements" data-pseudo-document-name="Advancement">
 
     <h3 class="section-header">


src/module/data/item/class.mjs
```diff
@@ -64,6 +64,21 @@ export default class ClassModel extends AdvancementModel {
   async getSheetContext(context) {
     context.characteristics = Object.entries(ds.CONFIG.characteristics).map(([value, { label }]) => ({ value, label }));
     context.kitOptions = Array.fromRange(3).map(number => ({ label: number, value: number }));
+
+    // Advancements
+    const advs = {};
+    const models = this.parent.getEmbeddedPseudoDocumentCollection("Advancement")[
+      context.isPlay ? "contents" : "sourceContents"
```
JPMeehan, line 71: Okay so outside of play mode we can do the raw data without the models?

src/module/data/item/class.mjs
```diff
@@ -64,6 +64,21 @@ export default class ClassModel extends AdvancementModel {
   async getSheetContext(context) {
     context.characteristics = Object.entries(ds.CONFIG.characteristics).map(([value, { label }]) => ({ value, label }));
     context.kitOptions = Array.fromRange(3).map(number => ({ label: number, value: number }));
+
+    // Advancements
+    const advs = {};
```
JPMeehan, line 69: Is this logic you can move into the shared AdvancementModel class? Or is that a future PR thing

JPMeehan (approved):
Overall makes sense to me but want to know more about design direction

src/module/data/item/class.mjs
krbz999: `sourceContents` is not the raw data, it's the array of sudokus that exist on the document prior to data prep. For example, this would ignore sudokus added by AEs. In hindsight likely irrelevant in this case of advancements.

src/module/data/item/class.mjs
krbz999: I have general thoughts on our sheet preparation logic, but I don't want to introduce a different design paradigm at this point. I'll see about it when I work on other item types that inherit from AdvancementModel. Easier to remove duplicated code after the fact.

src/module/data/item/class.mjs
JPMeehan: Whatever makes most sense to you, I'm good with

PR #603, opened 2025-06-15
Pseudo-document upkeep

krbz999:
Pseudo-documents support having embedded sudokus of their own, so data prep should be relegated to the parent sudo.

Also turns out we can use `DocumentTypeField`. Controversial? Perhaps.

Also fixes the fact that `isSource` always threw an error for AEs adding sudos - this was something that had come about from a change before sudos were added to DS. This new implementation of `isSource` should fix that by relying on `metadata.embedded`.

Also adds drag data method to sudos which will be relevant down the line.

### DIFF
--- a/src/module/data/pseudo-documents/pseudo-document.mjs
+++ b/src/module/data/pseudo-documents/pseudo-document.mjs
@@ -114,7 +114,14 @@ export default class PseudoDocument extends foundry.abstract.DataModel {
    * Prepare base data. This method is not called automatically; it is the responsibility
    * of the parent document to ensure pseudo-documents prepare base and derived data.
    */
-  prepareBaseData() {}
+  prepareBaseData() {
+    const documentNames = Object.keys(this.constructor.metadata.embedded);
+    for (const documentName of documentNames) {
+      for (const pseudoDocument of this.getEmbeddedPseudoDocumentCollection(documentName)) {
+        pseudoDocument.prepareBaseData();
+      }
+    }
+  }
 
   /* -------------------------------------------------- */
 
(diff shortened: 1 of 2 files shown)


JPMeehan:
What's the context of sudokus that have sudokus?

src/module/data/pseudo-documents/pseudo-document.mjs
```diff
@@ -155,7 +197,13 @@ export default class PseudoDocument extends foundry.abstract.DataModel {
    * @type {boolean}
    */
   get isSource() {
-    const source = foundry.utils.getProperty(this.document._source, this.fieldPath);
+    const docName = this.documentName;
+    const fieldPath = this.parent.constructor.metadata.embedded[docName];
+
+    const source = foundry.utils.getProperty(
+      this.parent._source,
+      this.parent instanceof foundry.abstract.TypeDataModel ? fieldPath.slice(7) : fieldPath,
```
JPMeehan, line 205: Is there a way to make this less of a magic number?

JPMeehan (approved):

src/module/data/pseudo-documents/pseudo-document.mjs
krbz999: Well if the parent is a TypeDataModel, then the path always starts with `system.`. I think core foundry actually does the same in some places. Can do a `replace("system.", "")` but we can't actually have a collection's path not start with that unless the parent is another sudoku.

JPMeehan: I'm happy with either

PR #607, opened 2025-06-16
Dev docs

JPMeehan:


krbz999:
The moment we decide to support removal of properties like skills and languages, I will insist we implement a custom field that shows "invalid" values on the document. E.g. if an actor has `stevebob: SteveBob` as a language, but a director removes it, it should still show `stevebob` and allow for it to be removed, and more importantly allow for it to stick around.

Unlike dnd5e, I want to make sure modules can *actually* be enabled or disabled at will without loss of data just by viewing a sheet.

PR #611, opened 2025-06-17
Document update

KJTailor:
Updated the document to include missed ability roll data, and represent the active effect duration in the System Automations page


src/docs/Active-Effects.md
```diff
@@ -11,12 +11,44 @@ The Foundry [Knowledge Base](https://foundryvtt.com/article/active-effects/) has
...
+
+Applying any of these DS core conditions applies a bespoke active effect, details for which can be found in the `System Automations` section.
+
+When creating a new temporary active effect by clicking on the `+ New Effect` button on the right of a active effect section header on this page, the duration is by default set to EoT (End of Turn).
+
+## Active Effect Window
+
+At any given time you can change the icon and name of an active effect.
+
+### Details
```
JPMeehan, line 26: Structurally, I'd prefer that since this *isn't* actually unique to Draw Steel it's not our responsibility to explain this stuff. But I also recognize that core docs aren't super great and *somebody* has gotta explain this stuff.

src/docs/Active-Effects.md
```diff
@@ -11,12 +11,44 @@ The Foundry [Knowledge Base](https://foundryvtt.com/article/active-effects/) has
...
+
+On the details page, you can give an effect a informative description. If you tick the `Effect Suspended` tick box, the effect will exist, but not take effect until activated. If you tick the `Apply Effect to Actor` tick box, the effect will be applied to the actor themselves.
```
JPMeehan, line 28: I think if we're going to talk about Effect Disabled & Apply Effect to Actor we probably need to talk about them in more detail. For example
- Effects are *also* automatically suppressed if they have a conventional duration (e.g. in rounds & turns) that has run over.
- We use the "Apply Effect to Actor" as a way to decide if an effect on an item is an "Applied Effect"; unchecked means it's available for that, but we won't have that application handling until 0.8 at least

src/docs/Active-Effects.md
```diff
@@ -11,12 +11,44 @@ The Foundry [Knowledge Base](https://foundryvtt.com/article/active-effects/) has
...
+If it is one of the automated conditions that require a targeted token as the source (`Frightened`, `Grabbed`, `Taunted`), and a token is targeted while applying the Active Effect, this targeted token will function as source for those conditions, applying edges and banes accordingly.
+
+### Duration
+
+Draw Steel has three different predefined effect durations `End of Turn (EoT)`, `Save Ends`, and `End of Encounter`.
+Additionally, the Save Ends duration allows for defining the saving throw formula. The default is `1d10 + @combat.save.bonus`.
+
++ `EoT` will automatically self-disable when an actor starts their turn after the affected actor had theirs.
++ `Save Ends` will create prompts for owners that allow to change the save threshold (e.g. due to Ancestry effects) and a text field to enter situational bonuses. The roll message has a button to spend a hero token to automatically succeed. If multiple players own an actor, the active GM will receive a dialog to help delegate rolls.
++ End of Encounter effects will automatically self-disable alongside the encounter.
```
JPMeehan, line 43: Could you improve the consistency with spelling out the duration label & using backticks?

src/docs/System-Automation.md
```diff
@@ -120,14 +124,28 @@ If a kit provides a bonus other than stamina or damage, these bonuses also get a
 
 If an ability is programmed with damage, untyped or typed, then when the ability is used, a button will appear in the chat window that is showing the power roll result, which will automatically apply the damage to a selected token. This also immediately takes damage weaknesses and immunities into account *(see below)*
 
+## Enrichers and applied damage from chat messages
```
JPMeehan, line 127: I would rather just point people to the Enrichers page.

src/docs/System-Automation.md
```diff
@@ -170,3 +188,20 @@ If the `Weakened` status is applied to an actor, all power rolls of that actor a
 
 If an actor falls below 50 % of their maximum stamina, they automatically get the `winded` status effect applied to them. This status does not have any effects by itself.
 If the actor's stamina gets healed above the 50 % threshold, the `winded` status automatically if removed.
+
+## Active Effects
+
+The above automated conditions are bespoke active effects usable via the token HUD, but may further be enhanced directly via active effects. This is explained in more detail in the `Active Effect` page.
```
JPMeehan, line 194:
```suggestion
The above automated conditions are canonical statuses applicable via the token HUD, but may further be enhanced directly via active effects. This is explained in more detail in the `Active Effect` page.
```

src/docs/System-Automation.md
```diff
@@ -170,3 +188,20 @@ If the `Weakened` status is applied to an actor, all power rolls of that actor a
...
+
+### Linking status effects
```
JPMeehan, line 196: We should try to minimize how many times we repeat information; this page should *only* be used for explaining information that's not otherwise covered.

JPMeehan (changes requested):

PR #615, opened 2025-06-18
Major refactor of DS Rolls & type updates

JPMeehan:
- Removed _prepareContext in favor of core's _prepareChatContext, which has identical behavior to what we had but with a slightly different function name
- Adjusted Rolls to use default exports
- General migration of `@link` to `@linkcode` in the doc strings
- Ensured consistency for links pointing to instance methods; VSCode has its own standard of *only* using . methods, so need to use the pipe labels 
image

### DIFF
--- a/src/module/data/actor/base.mjs
+++ b/src/module/data/actor/base.mjs
@@ -429,7 +429,7 @@ export default class BaseActorModel extends SubtypeModelMixin(foundry.abstract.T
 
   /**
    * Fetch information about the core resource for this actor subtype.
-   * {@linkcode AbilityModel.use | AbilityModel#use}
+   * @see {@linkcode AbilityModel.use | AbilityModel#use}
    * @abstract
    * @returns {{
    *  name: string;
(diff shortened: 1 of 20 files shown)


src/module/data/actor/base.mjs
```diff
@@ -429,7 +429,7 @@ export default class BaseActorModel extends SubtypeModelMixin(foundry.abstract.T
 
   /**
    * Fetch information about the core resource for this actor subtype.
-   * {@link AbilityModel#use}
+   * @see {@linkcode AbilityModel.use | AbilityModel#use}
```
krbz999, line 432: Thanks I hate it.

src/module/rolls/power.mjs
```diff
@@ -277,8 +277,9 @@ export class PowerRoll extends DSRoll {
     return flavorlessRoll.formula;
   }
 
-  async _prepareContext({ flavor, isPrivate }) {
-    const context = await super._prepareContext({ flavor, isPrivate });
+  /** @inheritdoc */
```
krbz999, line 280: Add a divider while we're here.

krbz999 (approved):
Good simplification. I don't love having to essentially document something twice but eh.

PR #616, opened 2025-06-19
Rolls/damage button

JPMeehan:
Refactored damage button logic so now base messages with damage rolls have take damage buttons

TODO: Healing, Documentation

image

### DIFF
--- a/src/module/rolls/damage.mjs
+++ b/src/module/rolls/damage.mjs
@@ -5,6 +5,22 @@ import DSRoll from "./base.mjs";
  */
 export default class DamageRoll extends DSRoll {
   /**
+   * Button callback to apply damage to selected actors
+   * @param {PointerEvent} event
+   */
+  static async applyDamageCallback(event) {
+    if (!canvas.tokens.controlled.length) return ui.notifications.error("DRAW_STEEL.Messages.AbilityUse.NoTokenSelected", { localize: true });
+
+    const type = event.target.dataset.type;
+    let amount = Number(event.target.dataset.amount);
+    if (event.shiftKey) amount = Math.floor(amount / 2);
+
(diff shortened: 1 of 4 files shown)


src/module/rolls/damage.mjs
```diff
@@ -4,21 +4,68 @@ import DSRoll from "./base.mjs";
  * Contains damage-specific info like damage types
  */
 export default class DamageRoll extends DSRoll {
+  /**
+   * Button callback to apply damage to selected actors
+   * @param {PointerEvent} event
+   */
+  static async applyDamageCallback(event) {
+    if (!canvas.tokens.controlled.length) return ui.notifications.error("DRAW_STEEL.Messages.AbilityUse.NoTokenSelected", { localize: true });
```
krbz999, line 12:
```suggestion
    if (!canvas.tokens.controlled.length) return void ui.notifications.error("DRAW_STEEL.Messages.AbilityUse.NoTokenSelected", { localize: true });
```

src/module/rolls/damage.mjs
```diff
@@ -4,21 +4,68 @@ import DSRoll from "./base.mjs";
...
+
+    const type = event.target.dataset.type;
+    let amount = Number(event.target.dataset.amount);
+    if (event.shiftKey) amount = Math.floor(amount / 2);
+
+    for (const actor of ds.utils.selectedActors()) {
+      await actor.system.takeDamage(amount, { type });
+    }
```
krbz999, line 20: Should use `event.currentTarget`, which is the element to which the listener was bound.

Secondly, are we unable to derive this from the message itself? Perhaps give the button `data-index` pointing to a roll in the `rolls` array?

krbz999 (approved):
Generally seems fine.

PR #619, opened 2025-06-19
Adjusted edit vs play mode logic

JPMeehan:
New actors and items now start in edit mode. Items no longer default to edit mode if opened while editable.


src/module/applications/api/document-sheet-mixin.mjs
```diff
@@ -95,6 +96,10 @@ export default base => {
     /** @inheritdoc */
     async _prepareContext(options) {
       const context = await super._prepareContext(options);
+      // New sheets should always start in edit mode
+      if (["createItem", "createActor"].includes(options.renderContext)) {
+        this._mode = DSDocumentSheet.MODES.EDIT;
+      }
```
krbz999, line 102: We can future-proof this for other document types.
```js
if (options.renderContext === `create${this.document.documentName}`) { ... }
```

PR #629, opened 2025-06-22
Player UI malice controls

### LINKED ISSUE #498: Malice does not show in counter until menu has been used to `give token`.
like the title says. When I load a load a map with an enemy on it, the Malice display does not show until after I gave the players a Hero Token.

Before giving token:

Image

After giving token:

Image

### LINKED ISSUE #515: Add manual controls for Malice gain/expenditure to lower left widget
A button labelled 'Add / Remove Malice' that displays a query requesting an integer (neg to remove, pos to add) should be added to the lower left control widget beneath the network display.

Image

cdverrett94:
Closes #515
Closes #498

- Adds an Adjust Malice and Reset Malice option to the player UI context menu.
   - updated the context menu methods to `_metaCurrencyContextMenuOptions` to reflect both malice and HT instead of just HT.
   - Adjust Malice prompts an input to increase/decrease the current malice value.
   - Malice options only show if there's a combat which is when the malice value only shows too
- Renders the player ui on combat create/delete.
   - Previously the render relied on the `MaliceModel.onChange` method rendering it, so it only rendered on Begin Combat. 
   - Rendering on delete is because if the malice was already 0, `MaliceModel.onChange` wouldn't trigger a render.

Screenshot 2025-06-22 113342

### DIFF
--- a/src/module/applications/ui/players.mjs
+++ b/src/module/applications/ui/players.mjs
@@ -2,6 +2,7 @@ import { systemID, systemPath } from "../../constants.mjs";
 
 /** @import { ContextMenuEntry } from "@client/applications/ux/context-menu.mjs" */
 /** @import { HeroTokenModel } from "../../data/settings/hero-tokens.mjs"; */
+/** @import { MaliceModel } from "../../data/settings/malice.mjs"; */
 
 /**
  * An extension of the core Players display that adds controls for hero tokens and malice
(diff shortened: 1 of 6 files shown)


JPMeehan (approved):
Please add changelog entry

cdverrett94:
> Please add changelog entry

Done

src/module/applications/ui/players.mjs
```diff
@@ -66,6 +67,26 @@ export default class DrawSteelPlayers extends foundry.applications.ui.Players {
...
+        condition: li => game.user.isGM && game.combat,
+        callback: async li => {
+          /** @type {MaliceModel} */
+          const malice = game.actors.malice;
+          await malice.adjustMalice();
+        },
+      },
+      {
+        name: "DRAW_STEEL.Setting.Malice.ResetMalice",
+        icon: "<i class=\"fa-solid fa-rotate\"></i>",
```
krbz999, line 82: Icons should all have `fa-fw`.

PR #633, opened 2025-06-24
Added support for negative heroic resources

### LINKED ISSUE #508: Add specialized support for heroic resources that can go negative
Some classes [the Talent] can dip into the negatives for their heroic resource. We should support that by extending our implementation of `Actor#modifyTokenAttribute`.

On a data level, we should add a FormulaField to the ClassModel that defaults to 0, but classes like the Talent can implement to provide a formula to derive the intended minimum - for the talent that would probably be `1 + @R`

JPMeehan:
- Token bar will clamp at the calculated negative value
- Added appropriate warnings to the ability use dialog
- Added lots more dividers

Closes #508

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -682,6 +682,10 @@
           "turnGain": {
             "label": "Gain Per Turn"
           },
+          "minimum": {
+            "label": "Resource Minimum",
+            "hint": "Can take a formula like '1 + @R'."
+          },
           "characteristics": {
             "core": {
               "label": "Core Characteristics"
--- a/lang/en.json
+++ b/lang/en.json
@@ -684,7 +684,7 @@
           },
           "minimum": {
             "label": "Resource Minimum",
-            "hint": "Can take a formula like '1 + @R'."
+            "hint": "Can take a formula like '-1 - @R'."
           },
           "characteristics": {
             "core": {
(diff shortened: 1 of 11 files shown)


cdverrett94:
Just a thought. Should the heroic resource stuff get grouped into its own `SchemaField`? Like `system.resources.primary.minimum`? It seems like it'd be more descriptive that way just looking at the data paths.

lang/en.json
```diff
@@ -682,6 +684,10 @@
           "turnGain": {
             "label": "Gain Per Turn"
           },
+          "minimum": {
+            "label": "Resource Minimum",
+            "hint": "Can take a formula like '-1 - @R'."
```
krbz999, line 689: This looks like something that should be a placeholder, not a hint.

src/module/documents/actor.mjs
```diff
@@ -46,10 +54,28 @@ export default class DrawSteelActor extends BaseDocumentMixin(foundry.documents.
...
+      case "stamina": return this.#modifyStamina(value, isDelta);
+      case "hero.primary.value": return this.#modifyHeroicResource(value, isDelta);
+      default: return super.modifyTokenAttribute(attribute, value, isDelta, isBar);
+    }
+  }
+
+  /* -------------------------------------------------- */
 
+  /**
+   * Handle how changes to a Stamina are applied to the Actor.
```
krbz999, line 71:
```suggestion
   * Handle how changes to Stamina are applied to the Actor.
```

krbz999 (approved):

PR #639, opened 2025-06-26
Added end turn button

JPMeehan:
- I think Foundry changed the handling of endCombat because now it always returns the Combat so we were always executing the award victories prompt
- Added a bunch of custom logic to end the current turn without advancing to a new combatant's turn

### DIFF
--- a/src/module/applications/sidebar/tabs/combat-tracker.mjs
+++ b/src/module/applications/sidebar/tabs/combat-tracker.mjs
@@ -20,6 +20,8 @@ export default class DrawSteelCombatTracker extends sidebar.tabs.CombatTracker {
     },
   };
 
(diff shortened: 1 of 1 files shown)


src/module/applications/sidebar/tabs/combat-tracker.mjs
```diff
@@ -78,6 +78,20 @@ export default class DrawSteelCombatTracker extends sidebar.tabs.CombatTracker {
     return context;
   }
 
+  async _prepareCombatContext(context, options) {
```
krbz999, line 81: missing annotation

krbz999: and dividers. B)

Issue #638, opened 2025-06-26
Button for Heroes/Monsters to End Turn


finaljas90:
Currently the only way for a character to make a save is for another character to start their turn. If the party or the Director wants to wait to see what the character that ends their turn gets on their save to determine who goes next, this can provide a significant tactical choice. In order to circumvent this, as well as make saving throws more accessible, perhaps there could be an "End Turn" or "Saving Throw" button that could make this dialogue pop up without having to rely on another character starting their turn?

Issue #613, opened 2025-06-18
[Active Effects] - Distance attributes not adjusting Ability range


Voltan-rev1:
- Foundry: 13.345
- DS Module: v0.7.2 #
- Discord Source: [Conversation](https://canary.discord.com/channels/332362513368875008/1342298358664138805/1384603222508961882)

Purpose:

- Add Active Effects that impact ability range within ability sheet on characters and related rolls.

Setup:
- Apply abilities with Ranged or Melee distance to a character.
- Apply Active Effect that increases the distance by a specific number. [Source of Active Effects:](https://github.com/MetaMorphic-Digital/draw-steel/wiki/Active-Effects)
*  e.g. - system.abilityBonuses.ranged.distance     |    Add    |     2
- Observe increase in Ability range distance within the Abilities list on the Character sheet and relevant rolls for that ability.

Current Behavior:
- Attempted two Active effects regarding range, Melee and Ranged as passive Active Effects.
*  system.abilityBonuses.ranged.distance
*  system.abilityBonuses.melee.distance
- After activating the Active Effect, no change in Ability range is observed.

Expected Behavior:
- Ability distances should change according to the active effect within the Character sheets and related rolls.

Notes:
- Kits increase this distance value as expected on the Abilities tab.
- Other Active Effects such as movement distance were successful in the same effect configuration.
- Not all Active Effects were attempted, others could have the same issue but the focus of this test was Melee and Ranged Distance modifiers for Active Effects which are common.

Image Example (Active Effect should change Range 10 to Range 12):

![Image](https://github.com/user-attachments/assets/248d3366-3ff1-4c17-95ba-e2e59911a1bb)

JPMeehan:
Ah, OK so the behavior here is simply that we're apply the bonuses from kits *before* Active Effects. Need to think about the correct behavior here but it's clear we got ahead of things on the docs.

Issue #609, opened 2025-06-17
implement that empty ability and equipment headers are hidden in play mode


KJTailor:
How would you feel about only showing categories in play mode that have something in them?

On player sheets that would be probably superfluous (except maybe free maneuvers and free triggered actions), but especially on enemy sheets that would cut down a lot of bloat

Same would go for equipment headers.

Issue #599, opened 2025-06-15
Light Mode Coloring Error


Zetesofos:
As of 0.7.2, the light mode has an issue with the Header Bar Coloring - hard to read:

![Image](https://github.com/user-attachments/assets/7f9e8073-fbc6-47e1-8265-6eeeee64fc8e)

JPMeehan:
Not seeing it?

![Image](https://github.com/user-attachments/assets/b70a2cbf-3642-4a51-8e0c-ca9f1b13aaeb)

Zetesofos:
So, for some reason, I'm still getting this. Are you testing in the client, or in a browswer by chance?

![Image](https://github.com/user-attachments/assets/d35bf0e4-49ba-41ac-ab70-1465d3fc755d)

Zetesofos:
It seems to persist across both firefox and edge (did two browser tests).

Zetesofos:
Works for me. Thank you for looking at it again.

PR #641, opened 2025-06-27
Release 0.7.3

JPMeehan:
### Added

- Added support for heroic resources that can go negative. (#508)
- Add Adjust Malice and Reset Malice context menu options to the player UI. (#515)
- Equipment now has a "quantity" property. Completing a project will now add to an equipment's quantity if it already exists on the actor, based on matching `dsid` values. (#558)
- Added Healing enricher, which lets you recover stamina or temporary stamina with `[[/heal]]`. See the wiki for more details. (#559)
- Added Damage enricher, which lets you add damage rolls to text with `[[/damage]]`. See the wiki for more details. (#568)
- Added a button in the Combat Tracker footer to end the current combatant's turn without beginning a new one. (#638)

### Changed

- Overhauled the Effects tab of the Item & Actor sheets. (#146)
  - Actor sheets now feature a "Status Conditions" section which lists the conditions specific to Draw Steel.
    - Clicking the buttons will toggle the status
    - If you have a status, it will be highlighted in orange
    - If the status is granted by a non-canonical condition, the button will be disabled
  - Hitting "New Effect" in the "Temporary Effects" section will now set the duration to End of Turn, instead of 1 round.
  - An effect's enriched description is now available by toggling the carat, like already existed for items.
- Replaced the default token status menu with a new one that allows applying a status with one of Draw Steel's unique durations. (#261)
- Adjusted display of the actor sheets in play mode. (#609)
  - Unused ability and equipment sections are no longer displayed in play mode.
  - Add Item buttons only display in edit mode.
- New actors and items now start in edit mode. Items no longer default to edit mode if opened while editable.
- Show malice value in the player UI on combat creation instead of only on beginning combat. (#498)
- Moved ability bonus calculations to after active effect application. (#613)

### Fixed

- Fixed various color issues on sheets that had their theme set individually. (#599)
- Fixed malice generation applying the first round bonus twice.
- The "Award Victories" prompt will no longer appear if you cancel out of ending a combat.


krbz999 (approved):
LGTM!

PR #642, opened 2025-06-27
Item Grant and Trait advancements

krbz999:
- Update types.
- Update BaseAdvancement, removing not-yet-used code.
- Update ItemGrantAdvancement.
- Add TraitAdvancement.
- Heavily update AdvancementChainLink class.

### DIFF
--- a/src/module/config.mjs
+++ b/src/module/config.mjs
@@ -1554,4 +1554,9 @@ preLocalize("projects.types", { key: "label" });
  * The options for a trait advancement.
  * @type {Record<string, TraitConfiguration>}
  */
-DRAW_STEEL.TRAITS = {};
+DRAW_STEEL.TRAITS = {
+  test: {
+    label: "Test Trait",
+    field: new foundry.data.fields.NumberField({ min: 1, max: 5, integer: true, nullable: false }),
+  },
+};
--- a/src/module/config.mjs
+++ b/src/module/config.mjs
@@ -1205,7 +1205,7 @@ Object.defineProperty(DRAW_STEEL.abilities.keywords, "optgroups", {
  */
 DRAW_STEEL.PowerRollEffect = {
   damage: {
-    label: "DRAW_STEEL.PSEUDO.POWER_ROLL_EFFECT.TYPES.damage",
+    label: "TYPES.PowerRollEffect.damage",
     documentClass: pseudoDocuments.powerRollEffects.DamagePowerRollEffect,
     properties: {
       ignoresImmunity: {
@@ -1214,11 +1214,11 @@ DRAW_STEEL.PowerRollEffect = {
     },
   },
   applied: {
-    label: "DRAW_STEEL.PSEUDO.POWER_ROLL_EFFECT.TYPES.applied",
+    label: "TYPES.PowerRollEffect.applied",
     documentClass: pseudoDocuments.powerRollEffects.AppliedPowerRollEffect,
   },
   forced: {
-    label: "DRAW_STEEL.PSEUDO.POWER_ROLL_EFFECT.TYPES.forced",
+    label: "TYPES.PowerRollEffect.forced",
     documentClass: pseudoDocuments.powerRollEffects.ForcedMovementPowerRollEffect,
     properties: {
       ignoresImmunity: {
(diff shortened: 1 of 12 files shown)


src/module/config.mjs
```diff
@@ -1537,3 +1541,17 @@ DRAW_STEEL.projects = {
   },
 };
 preLocalize("projects.types", { key: "label" });
+
+/* -------------------------------------------------- */
+
+/**
+ * @typedef TraitConfiguration
+ * @property {string} label                                           Human-readable label of the trait.
+ * @property {InstanceType<foundry.data.fields.DataField>} [field]    A field used to render the input.
```
JPMeehan, line 1550: Feels like a type import of DataField at the top of the file would be reasonable since I doubt it will be the only place we use it.


Also, is this a place where widget might be helpful? Widgets allow us to provide functions that render fields normally non-renderable (e.g. an embedded data field)

src/module/data/pseudo-documents/advancements/trait-advancement.mjs
```diff
@@ -0,0 +1,131 @@
...
+export default class TraitAdvancement extends BaseAdvancement {
+  /** @inheritdoc */
+  static defineSchema() {
+    return Object.assign(super.defineSchema(), {
+      requirements: new SchemaField({
+        level: new NumberField({ integer: true, min: 1, max: 10, nullable: false, initial: 1 }),
+      }),
+      traits: new TypedObjectField(new SchemaField({
+        label: new StringField({ required: true }),
+        trait: new StringField({ required: true, blank: false, choices: () => ds.CONFIG.TRAITS }),
```
JPMeehan, line 14: I think with choices we don't need required & blank? The defaults are different

src/module/data/pseudo-documents/advancements/trait-advancement.mjs
```diff
@@ -0,0 +1,131 @@
...
+      ? traits.filter(([k]) => node.selected[k])
+      : item.isEmbedded
+        ? foundry.utils.getProperty(item, path) ?? []
+        : [];
+
+    const content = [];
+    for (const [traitId, trait] of traits) {
+      const fgroup = foundry.applications.fields.createFormGroup({
+        label: trait.label,
+        input: foundry.utils.parseHTML(`<input type="checkbox" value="${traitId}" name="choices" ${chosen.includes(traitId) ? "checked" : ""}>`),
```
JPMeehan, line 98: I think I'd prefer `foundry.applications.cteateCheckboxInput`?

JPMeehan (changes requested):
Currently having issues rendering the Trait advancement, but here's my static analysis

src/module/data/pseudo-documents/advancements/trait-advancement.mjs
krbz999: This needs a better application regardless, so leaving as a TODO for when we do the actual leveling-up part.

src/module/data/pseudo-documents/typed-pseudo-document.mjs
```diff
@@ -51,6 +51,16 @@ export default class TypedPseudoDocument extends PseudoDocument {
 
   /* -------------------------------------------------- */
 
+  /** @inheritdoc */
+  prepareDerivedData() {
+    super.prepareDerivedData();
+    if (!this.name) {
+      this.name = game.i18n.localize(`TYPES.${this.documentName}.${this.type}`);
```
JPMeehan, line 58: Could also one liner it with ||=

src/module/data/pseudo-documents/advancements/trait-advancement.mjs
```diff
@@ -0,0 +1,131 @@
...
+export default class TraitAdvancement extends BaseAdvancement {
+  /** @inheritdoc */
+  static defineSchema() {
+    return Object.assign(super.defineSchema(), {
+      requirements: new SchemaField({
+        level: new NumberField({ integer: true, min: 1, max: 10, nullable: false, initial: 1 }),
+      }),
+      traits: new TypedObjectField(new SchemaField({
+        label: new StringField({ required: true }),
+        trait: new StringField({ choices: () => ds.CONFIG.TRAITS }),
```
JPMeehan, line 14: I was mistaken, still need required: true, *but* it does do blank: false

JPMeehan (commented):
Errors on actor drop now

src/module/data/pseudo-documents/advancements/trait-advancement.mjs
krbz999: We can add `required: true`, but it's always made with an initial value so it shouldn't matter.

JPMeehan (approved):

PR #644, opened 2025-06-27
Updated system minimums

JPMeehan:
Increased minimums to 13.346 so we can rely on various v13 era features without worrying about compatibility errors


krbz999:
Getting a bit ahead of ourselves there, let's see how advancements go, though.

PR #663, opened 2025-06-29
Removed duplicate function

JPMeehan:
Somehow a merge added a second copy of this function to the mixin class


krbz999 (approved):
Probably a bad rebase from the advancement development branch.

PR #666, opened 2025-06-29
[#658] Add description to advancement sheets.

### LINKED ISSUE #658: Add descriptions to Advancements
As a content creator, I would like to include the basic description for features that grant abilities without making that a whole feature that grants the ability.

krbz999:
Closes #658.

Not enriching, cus I didn't see any point in making this toggled.

image

### DIFF
--- a/system.json
+++ b/system.json
@@ -45,7 +45,7 @@
           "advancements.*.img": ["IMAGE"],
           "advancements.*.traits.*.img": ["IMAGE"]
         },
-        "htmlFields": ["description.value", "description.gm"],
+        "htmlFields": ["description.value", "description.gm", "advancements.*.description"],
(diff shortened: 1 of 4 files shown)


system.json
```diff
@@ -41,7 +41,7 @@
         "filePathFields": {
           "advancements.*.img": ["IMAGE"]
         },
-        "htmlFields": ["description.value", "description.gm"],
+        "htmlFields": ["description.value", "description.gm", "advancements.*.description"],
```
JPMeehan, line 44: Should we just add it to the others we know will have advancements now so we don't forget?

JPMeehan (approved):
I agree no point on toggling here, we can consider enrichment wherever we actually display it (The advancement flow UI?)

PR #672, opened 2025-07-01
Added Apply Effect Button

### LINKED ISSUE #214: Add apply effect button to chat cards
As a Director, I would like to be able to apply linked active effects from ability use messages.

JPMeehan:
This doesn't quite handle the potency logic, but it creates buttons that allow applying effects.

image

Closes #214

### DIFF
--- a/src/module/data/message/ability-use.mjs
+++ b/src/module/data/message/ability-use.mjs
@@ -31,6 +31,8 @@ export default class AbilityUseModel extends BaseMessageModel {
     return schema;
   }
 
+  /* -------------------------------------------------- */
+
   /**
    * The displayed result tier
    * @type {1 | 2 | 3 | undefined}
@@ -136,6 +138,7 @@ export default class AbilityUseModel extends BaseMessageModel {
 
       for (const actor of ds.utils.tokensToActors()) {
         // reusing the ID will block creation if it's already on the actor
+        // TODO: Update when https://github.com/foundryvtt/foundryvtt/issues/11898 is implemented
         actor.createEmbeddedDocuments("ActiveEffect", [tempEffect.toObject()], { keepId: noStack });
       }
     });
--- a/src/module/data/message/ability-use.mjs
+++ b/src/module/data/message/ability-use.mjs
@@ -1,7 +1,11 @@
+
 import BaseMessageModel from "./base.mjs";
 
-/** @import AbilityModel from "../item/ability.mjs" */
-/** @import DrawSteelItem from "../../documents/item.mjs" */
+/**
+ * @import AbilityModel from "../item/ability.mjs";
+ * @import AppliedPowerRollEffect from "../pseudo-documents/power-roll-effects/applied-effect.mjs";
+ * @import DrawSteelItem from "../../documents/item.mjs"
+ */
 
 const fields = foundry.data.fields;
 
@@ -25,6 +29,15 @@ export default class AbilityUseModel extends BaseMessageModel {
     return schema;
   }
 
+  /**
+   * The displayed result tier
+   * @type {1 | 2 | 3 | undefined}
+   */
+  get tier() {
+    // First roll is always the base roll, second is the result. Only one tier per message.
+    return this.parent.rolls[1]?.product;
+  }
+
   /* -------------------------------------------------- */
 
   /** @inheritdoc */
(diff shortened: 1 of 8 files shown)


src/module/data/message/ability-use.mjs
```diff
@@ -25,6 +31,25 @@ export default class AbilityUseModel extends BaseMessageModel {
     return schema;
   }
 
+  /**
+   * The displayed result tier
+   * @type {1 | 2 | 3 | undefined}
+   */
+  get tier() {
```
krbz999, line 38: divideeeeeers

src/module/data/message/ability-use.mjs
```diff
@@ -87,5 +109,35 @@ export default class AbilityUseModel extends BaseMessageModel {
...
+        transfer: true,
+        origin: pre.uuid,
+        system: {},
+      };
+      if (config.end) updates.system.end = { type: config.end };
+      tempEffect.updateSource(updates);
+
+      for (const actor of ds.utils.tokensToActors()) {
+        // reusing the ID will block creation if it's already on the actor
+        actor.createEmbeddedDocuments("ActiveEffect", [tempEffect.toObject()], { keepId: noStack });
```
krbz999, line 139: I was unable to find the core issue, but yknow the one about being able to make only a single db operation - let's leave a comment here to improve this in the future (hopefully v14)

src/module/data/pseudo-documents/power-roll-effects/applied-effect.mjs
```diff
@@ -138,4 +138,42 @@ export default class AppliedPowerRollEffect extends BasePowerRollEffect {
     const potencyString = this.toPotencyText(tier);
     return this.applied[`tier${tier}`].display.replaceAll("{{potency}}", potencyString);
   }
+
+  /* -------------------------------------------------- */
+
+  /**
+   * @inheritdoc
+   * @param {1 | 2 | 3} tier
+   */
```
krbz999, line 147: is just inheritdoc not sufficient?

src/module/data/pseudo-documents/power-roll-effects/applied-effect.mjs
```diff
@@ -138,4 +138,42 @@ export default class AppliedPowerRollEffect extends BasePowerRollEffect {
...
+  /* -------------------------------------------------- */
+
+  /**
+   * Fetch the applicable effect by its key, or fallback to a simple object
+   * @param {string} key
+   * @returns {DrawSteelActiveEffect | StatusEffectConfig | { name: string }}
+   * @protected
+   */
+  _getEffect(key) {
+    return this.item.effects.get(key) || CONFIG.statusEffects.find(s => key === s.id) || { name: key };
```
krbz999, line 177: Feels like we have one method here serving three purposes. I'm fine with writing more code to be more explicit about what it's accomplishing.

src/module/utils/construct-html-button.mjs
```diff
@@ -1,25 +1,28 @@
...
   const button = document.createElement("button");
   button.type = type;
 
   for (const [key, value] of Object.entries(dataset)) {
     button.dataset[key] = value;
   }
   button.classList.add(...classes);
-  if (icon) icon = `<i class="${icon}"></i> `;
+  let image = "";
+  if (img) image = `<img src="${img}" alt=${label} />`;
```
krbz999, line 22:
```suggestion
  if (img) image = `<img src="${img}" alt="${label}">`;
```

src/styles/system/applications/sidebar/tabs/chat.css
```diff
@@ -10,6 +11,17 @@
     footer {
       button {
         margin-top: 3px;
+        &.apply-effect {
+          /* Status conditions should be inverted to dark on the light mode chat */
+          .themed.theme-light &[data-type="status"] {
+            img {
+              filter: invert(1);
+            }
```
krbz999, line 19: This is where svgs and some css variables (that depend on light/dark mode) would be very handy. For a follow-up PR.

krbz999 (changes requested):
Seems mostly good. Small comments re style, and I feel somewhat strongly that multi-purpose methods should either get yote or reworked possibly into multiple methods.

src/module/data/pseudo-documents/power-roll-effects/applied-effect.mjs
JPMeehan: So, this is about translating the keys of our TOF into whatever they're supposed to be representing, I think what it's doing is pretty consistent. The TOF is where we've got the multi-tasking, because we're tracking both statuses *and* local active effects. The third param is just a fallback in case we can't actually find our key. The middle bit will be better in v14 when they switch from an Array to a Record

krbz999 (approved):

PR #674, opened 2025-07-02
Add level up configuration dialog

krbz999:
- Adds a very simple advancement configuration dialog.
- This presents the UI for configuring the 'nodes' of advancements chains.
- This also supports an entire range of levels, using the `levels` property of advancements.
- For testing, I changed the level 1 advancement of Tactician to be a 'Choose 1'.
- There is currently no follow-up from the dialog; it configures the chains. Future PR will handle the actual item creation from item grants and so forth.
- Also deleted some old advancement files found in `/documents`. I think they were a legacy attempt?

image


src/module/applications/apps/advancement/chain-configuration-dialog.mjs
```diff
@@ -0,0 +1,142 @@
...
+
+  /** @type {import("../../../_types").ContextPartHandler} */
+  async _preparePartContextChains(context, options) {
+    context.ctx = { chains: this.#chains.map(c => c.active()) };
+    return context;
+  }
+
+  /* -------------------------------------------------- */
+
+  /** @type {import("../../../_types").ContextPartHandler} */
```
JPMeehan, line 96: Stylistically I'd prefer one type import at the top that's reused

src/module/applications/sheets/character.mjs
```diff
@@ -280,7 +280,22 @@ export default class DrawSteelCharacterSheet extends DrawSteelActorSheet {
...
       }
-      await ds.data.pseudoDocuments.advancements.BaseAdvancement.performChanges(this.document, item);
+
+      const advancements = item.getEmbeddedPseudoDocumentCollection("Advancement");
+      const chains = [];
+      const range = [this.document.system.level, this.document.system.level + 1];
+      for (const advancement of advancements) {
+        console.warn(advancement.levels);
+        const validRange = advancement.levels.some(level => level.between(...range));
+        console.warn({ validRange });
```
JPMeehan, line 290: logging?

src/module/applications/apps/advancement/chain-configuration-dialog.mjs
krbz999: Oops, looks like that was part of my scaffolding in arti.

PR #687, opened 2025-07-05
Moved recoveries out of `hero`, made max stamina and recoveries derived

### LINKED ISSUE #627: Make character stamina and recoveries derived entirely from class
- Remove the max stamina input from the sheet for characters 
- 0 out the property in prepareBaseData 

All max stamina adjustments for characters must go through Active Effects.

JPMeehan:
With these two properties now purely derived, I removed them from the edit mode of the sheet.

image

image

Closes #627

### DIFF
--- a/src/module/data/item/class.mjs
+++ b/src/module/data/item/class.mjs
@@ -1,6 +1,6 @@
 import { systemPath } from "../../constants.mjs";
 import FormulaField from "../fields/formula-field.mjs";
-import { setOptions } from "../helpers.mjs";
+import { requiredInteger, setOptions } from "../helpers.mjs";
 import AdvancementModel from "./advancement.mjs";
 
 /**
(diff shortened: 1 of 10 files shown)


src/module/data/item/class.mjs
```diff
@@ -87,7 +87,15 @@ export default class ClassModel extends AdvancementModel {
   /** @inheritdoc */
   _onCreate(data, options, userId) {
     if (this.actor && (this.actor.type === "character") && (game.userId === userId)) {
-      this.actor.update({ "system.hero.recoveries.value": this.recoveries });
+      this.actor.update({ "system.recoveries.value": this.recoveries });
+    }
+  }
+
+  /** @inheritdoc */
```
krbz999, line 94: Divider.

krbz999 (approved):
LGTM. We may consider moving the remaining two properties into the root as well.

PR #688, opened 2025-07-05
Record trait advancements during data prep

krbz999:
When preparing items, grab trait advancements and record what changes they are going to make. It is then the responsibility of the actor model (hero) to apply these traits at the correct step.

Leaving as draft for now until we have actual changes with which to test.


src/module/documents/item.mjs
```diff
@@ -20,6 +21,29 @@ export default class DrawSteelItem extends BaseDocumentMixin(foundry.documents.I
 
   /* -------------------------------------------------- */
 
+  /** @inheritdoc */
+  prepareBaseData() {
```
JPMeehan, line 25: Is there a reason we're doing this in Item instead of AdvancementModel?

src/module/data/actor/character.mjs
```diff
@@ -270,6 +270,16 @@ export default class CharacterModel extends BaseActorModel {
 
   /* -------------------------------------------------- */
 
+  /**
+   * Internal record used to cache trait advancements to apply their changes during data prep.
+   * This record is populated during `prepareEmbeddedDocuments`.
+   * @type {Record<string, Set>}
+   * @internal
+   */
+  _traits;
```
JPMeehan, line 279: I think we can reasonably initialize this to {}

krbz999:
Adding traits during derived data. I don't *currently* see a reason why we would want to apply these earlier (eg before embedded documents).

src/module/data/actor/character.mjs
```diff
@@ -281,6 +288,16 @@ export default class CharacterModel extends BaseActorModel {
 
   /* -------------------------------------------------- */
 
+  /**
+   * Internal record used to cache trait advancements to apply their changes during data prep.
+   * This record is populated during `prepareEmbeddedDocuments`.
+   * @type {Record<string, Set>}
+   * @internal
+   */
+  _traits = {};
```
JPMeehan, line 297: Should we be resetting this in prepareBaseData? Right now stuff can get cached and won't go away if you remove the advanceable. Or do we only want to reset this cache _onDelete? That would be *before* the `reset` call, so that could work if we're really concerned about performance or something.

src/module/data/item/advancement.mjs
```diff
@@ -24,4 +25,27 @@ export default class AdvancementModel extends BaseItemModel {
...
+    if ((this.actor?.type !== "character")) return;
+
+    const record = this.actor.system._traits;
+    const flags = this.parent.flags[systemID]?.advancement ?? {};
+    const addTrait = (type, trait) => {
+      record[type] ??= new Set();
+      for (const k of trait) record[type].add(k);
+    };
+    for (const advancement of this.advancements) {
+      if (!["skill", "language"].includes(advancement.type)) continue;
```
JPMeehan, line 44: I did add an `isTrait` getter we can use instead of the `includes`

src/module/data/actor/character.mjs
JPMeehan: Oh, that's a fair point that since it's down here there will be a new instance. Let me check with a log

JPMeehan: Yep it is properly resetting itself upon reconstruction

JPMeehan:
Need to check who's causing the problem, but currently *every* possible skill is being added

image

src/module/data/item/advancement.mjs
```diff
@@ -24,4 +25,27 @@ export default class AdvancementModel extends BaseItemModel {
...
+    const flags = this.parent.flags[systemID]?.advancement ?? {};
+    const addTrait = (type, trait) => {
+      record[type] ??= new Set();
+      for (const k of trait) record[type].add(k);
+    };
+    for (const advancement of this.advancements) {
+      if (!advancement.isTrait) continue;
+      const selected = advancement.isChoice
+        ? flags[advancement.id]?.selected ?? []
+        : Object.keys(advancement.traitChoices);
```
JPMeehan, line 47: This is every possible trait, `traitOptions` is the narrowed version. So, would need to map `traitOptions` to just the `value` property.

JPMeehan (changes requested):

src/module/data/item/advancement.mjs
JPMeehan: I think we could turn `traitChoices` into a static method to make it clearer?

krbz999: I tested with languages and that worked just fine. I don't see what is wrong here; if it's a choice (`isChoice`) then we use only the selected ones, otherwise it's *all* the options.

JPMeehan: Double checking; this is actually the only spot that actually directly uses `traitChoices`, otherwise stuff uses `traitOptions`, so we can probably just dump that getter?

krbz999:
Re-tested with the Shadow and a Language advancement at level 2 - everything seems to work as expected.

JPMeehan (approved):

PR #695, opened 2025-07-06
Create and update items during advancement

krbz999:
Seems to work in my testing.
Added a temporary skill advancement to the Tactician class to test the updating of existing items.


JPMeehan (commented):
Dragging another class on appears to prompt a level up? That seems unintended

JPMeehan (approved):

PR #696, opened 2025-07-06
I18n update

JPMeehan:
- Keys inside of DRAW_STEEL either match a value of `CONST.DOCUMENT_TYPES` or are in all caps
- Moved a number of keys around
- Caught some double & missing localizations
- Factored out a number of `lookup` calls


krbz999 (approved):
A proper review would require going through the entire system. I feel like that's too much to ask, so given that it is just i18n, they won't cause major issues if some keys are misaligned.

JPMeehan:
I did some testing using the new `CONFIG.debug.i18n` which was actually pretty helpful.

PR #700, opened 2025-07-07
Non class advancements

JPMeehan:
- Advancements now properly display on non-class items
- Refactored the drop logic generally
- Fixed the _onCreate/_onDelete logic for Careers
- Made Kits & Features extend AdvancementModel
- Added starter compendiums for new advanceables

### DIFF
--- a/src/module/applications/sheets/character.mjs
+++ b/src/module/applications/sheets/character.mjs
@@ -338,7 +338,7 @@ export default class DrawSteelCharacterSheet extends DrawSteelActorSheet {
         ui.notifications.error("DRAW_STEEL.ADVANCEMENT.WARNING.cannotAddNewType", {
           format: { type: game.i18n.localize(CONFIG.Item.typeLabels[item.type]) },
         });
-        return;
+        throw new Error("Cannot add a new class to an actor already with one");
       }
       return this.document.system.advance({ levels: 1, item });
     } else if (item.system instanceof AdvancementModel) {
@@ -356,7 +356,7 @@ export default class DrawSteelCharacterSheet extends DrawSteelActorSheet {
           });
           // TODO: Undo previous item's advancements
           if (replace) await existing.delete();
-          else return;
+          else throw new Error(`Cannot add a new ${item.type} to a character that already has one without replacing the old.`);
         }
       }
       else if (item.type === "kit") {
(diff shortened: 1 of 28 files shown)


src/module/applications/sheets/character.mjs
```diff
@@ -325,12 +335,48 @@ export default class DrawSteelCharacterSheet extends DrawSteelActorSheet {
     if (item.type === "class") {
       const cls = this.document.system.class;
       if (cls && (cls.dsid !== item.dsid)) {
-        ui.notifications.error("DRAW_STEEL.ADVANCEMENT.WARNING.cannotAddNewClass", { localize: true });
+        ui.notifications.error("DRAW_STEEL.ADVANCEMENT.WARNING.cannotAddNewType", {
```
krbz999, line 338: In hindsight, we need to throw an error here. Returning falsy indicates the drop was not handled - which it was.

src/module/applications/sheets/character.mjs
```diff
@@ -325,12 +335,48 @@ export default class DrawSteelCharacterSheet extends DrawSteelActorSheet {
...
+        const actorClass = this.document.system.class;
+        if (actorClass?.system.kits === 0) {
+          const message = game.i18n.format("DRAW_STEEL.Item.kit.NotAllowedByClass", { class: actorClass.name });
+          ui.notifications.error(message);
+          return false;
+        }
+        const swapKit = await item.system.kitSwapDialog(this.document);
+        if (swapKit === false) return false;
+      }
+      return item.system.applyAdvancements({ actor: this.document, levelStart: 1, levelEnd: this.document.system.level });
```
krbz999, line 372: For classes, we use `range` (a tuple of numbers). We should align there - I'm fine with `levelEnd` and `Start` but perhaps `range: { start, end }`?

src/module/applications/sheets/character.mjs
```diff
@@ -325,12 +335,48 @@ export default class DrawSteelCharacterSheet extends DrawSteelActorSheet {
...
     }
 
-    return super._onDropItem(event, item);
+    // Fixed default implementation
```
krbz999, line 375: Did anyone make an issue on the core repo

krbz999 (approved):
It just works. Terrifying, really.

src/module/applications/sheets/character.mjs
JPMeehan: https://github.com/foundryvtt/foundryvtt/issues/13166 done

src/module/applications/sheets/character.mjs
krbz999: I think those names would make more intuitive sense.

Issue #716, opened 2025-07-08
Block Advancement related item creation under conditions 


JPMeehan:
- Ancestry/Career/Class/Subclass/Kit should only be created inside of a compendium
- Allow cultures in world since no item grants
- Block non-compendium items from being added to item grants

All of this should involve expansive UI notifications pointing people to the setup screen module maker

PR #724, opened 2025-07-12
[#716] Block advancement-related creation

JPMeehan:
Do you also want to do item-type restrictions here?


JPMeehan:
Note that while BaseItemModel has packOnly: false, it won't show up in the create dialog because ClientDocument.createDialog always filters the `"base"` type out.

PR #727, opened 2025-07-14
[#694] Add next respite effect ends option

### LINKED ISSUE #694: Add "Until Next Respite" duration
Some rare features provide a bonus until the "next respite". This should be added alongside the existing End of Turn/Save Ends/End of Encounter durations.

cdverrett94:
Closes #694


JPMeehan:
Could you update the Changelog? Also, are there any concerns about the text width in the effects tab?

cdverrett94:
Done!

PR #730, opened 2025-07-15
Cleaned up item context options

JPMeehan:
- Adjusted i18n
- simplified the (probably unnecessary) error check to 1 line


krbz999:
Agree the error checks are not needed. Feel free to remove them - if we get an actual error there, we have bigger issues imo

PR #746, opened 2025-07-22
[#712] Advancement Restrictions

### LINKED ISSUE #712: Implement sensible restrictions on advancements depending on item type
For example, item grants can't grant any item type. For non-class items, the pool of types is also smaller.

And not all item types that allow for advancements in general should allow item grant advancements.

JPMeehan:
- Implemented create-level controls on what advancements can go on what item types.
  - Ancestries & Kits only have Item Grants
  - Cultures only have Skill, Language, and their custom types
  - Everything else (career, class, complication, feature, subclass) has access to all three advancement types
- Implemented restrictions on what can be dropped on an Item Grant. I kept Kits because of the unique Stormwright kits, maybe that was incorrect.
- Refactored our handling of PseudoDocument.createDialog to use more helper functions and no total overrides.

I think we could re-evaluate exactly how we're deciding TypedPseudoDocument.TYPES, seems slightly redundant in construction.

Closes #712

### DIFF
--- a/src/module/data/pseudo-documents/advancements/item-grant-advancement.mjs
+++ b/src/module/data/pseudo-documents/advancements/item-grant-advancement.mjs
@@ -31,6 +31,14 @@ export default class ItemGrantAdvancement extends BaseAdvancement {
 
   /* -------------------------------------------------- */
 
+  /**
+   * Item types that cannot be added to an Item Grant.
+   * @type {Set<string>}
+   */
+  static RESTRICTED_TYPES = new Set(["ancestry", "career", "class", "complication", "culture", "subclass"]);
+
+  /* -------------------------------------------------- */
+
   /** @inheritdoc */
   get levels() {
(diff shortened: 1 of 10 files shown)


src/module/data/pseudo-documents/advancements/item-grant-advancement.mjs
```diff
@@ -31,6 +31,14 @@ export default class ItemGrantAdvancement extends BaseAdvancement {
 
   /* -------------------------------------------------- */
 
+  /**
+   * Item types that cannot be added to an Item Grant.
+   * @type {Set<string>}
+   */
+  static RESTRICTED_TYPES = new Set(["ancestry", "career", "class", "complication", "culture", "subclass"]);
```
krbz999, line 38: Let's invert this, list item types that *can* be added.

krbz999 (approved):

PR #751, opened 2025-07-25
[#734] Fix save and project rolls not applying roll mode

### LINKED ISSUE #734: Saving Throw Roll ignores roll mode from dialog
Need to come up with a common solution to the roll mode logic; maybe override _processFormData in the RollDialog class?

cdverrett94:
Closes #734 

Also noticed that project rolls were not applying roll mode too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@
 - Fixed display of popped out ability messages
 - Fixed display of multi-result abilities
 - Fixed unrequited prompting for end of turn events (#673)
+- Fixed project and save rolls not respecting roll mode settings. (#734)
 
 ## 0.7.3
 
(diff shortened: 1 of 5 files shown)


JPMeehan (approved):

cdverrett94:
> My only structural question is if we should move some of this roll options handling to RollDialog?

I didn't do it that way because the `PowerRollDialog` and `SavingThrowDialog` don't really share a lot of logic in `_processFormData` other than grabbing the `rollMode`. The `SavingThrowDialog` use `formData` exclusively (other than `rollMode`). `PowerRollDialog` primarily uses `options.context` with a bit of `formData`. SavingThrowDialog is also a single roll config whereas the `PowerRollDialog` needs to be an array of roll configs. 

I suppose we could have `RollDialog#_processFormData` return something like `{ rollMode: this.options.context.rollMode }` and then add the specific roll(s) in the subclass. Would have to add `const formData = foundry.utils.expandObject(formData.object);` to both subclasses though.

If you think that's a better way to handle it, I can definitely make that change.

JPMeehan:
I think we can revisit later if we change our minds, but otherwise it's probably fine to just handle it individually.

Issue #537, opened 2025-05-24
Epic Resources


JPMeehan:
Tracking ticket for Epic Resources once they are added to some publicly licensed document.

PR #753, opened 2025-07-27
Resource buttons

JPMeehan:
Adjusted the resource label-buttons to be clearer

Editable
image

Non-editable
image

Also switched a few flexrow to form-group to add the native gap


JPMeehan:
Currently blocking #537

PR #755, opened 2025-07-31
Add Censor Class to Compendium

cdverrett94:
I've got the Censor level 1 stuff added. Right now I'm hoping to just verify that how I'm doing the advancements and compendium folders is desirable.

For features that grant an ability, I did the feature (with ability embed) and ability. I started doing it that way because the Judgment feature has some extra text outside of the ability and then just continued doing it that way.

Once I get validation on what I have so far, I'll work on finishing the rest of the class.


JPMeehan:
I think, generally speaking, I would prefer to have features-that-grant-abilities (e.g. My Life For Yours) to *not* have an associated feature and instead just have that extra description in the Advancement. However, I see that Judgment is weird; my gut instinct is honestly that the Judgment Order Benefit should probably be in the *subclass*, since I assume any 3rd party/future subclasses should also grant a Judgment benefit.

PR #784, opened 2025-08-02
Implemented showing the compendium source of an item or NPC

JPMeehan:
Antecedent to having a header button that performs an "update from compendium", thought it was a good place to render this info.

image

### DIFF
--- a/src/module/applications/apps/document-source-input.mjs
+++ b/src/module/applications/apps/document-source-input.mjs
@@ -30,6 +30,8 @@ export default class DocumentSourceInput extends DocumentInput {
     const context = await super._prepareContext(options);
     context.sourceValues = this.document.system.source._source;
     context.sourceFields = this.document.system.source.schema.fields;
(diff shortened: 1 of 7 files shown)


src/module/applications/apps/document-source-input.mjs
```diff
@@ -30,6 +33,20 @@ export default class DocumentSourceInput extends DocumentInput {
...
+    const compendiumSource = await fromUuid(this.document._stats.compendiumSource);
+    context.sourceLink = compendiumSource?.toAnchor();
     return context;
   }
+
+  /* -------------------------------------------------- */
+
+  /**
+     * Clear the document's compendium source.
+     * @this {DSDocumentSheet}
```
krbz999, line 45: Need indentation fix here.

krbz999 (approved):

krbz999:
"Update from pack" functionality should wait until there's been much more live testing of the system and compendiums. Then it will become clear whether it is needed and what functions it should serve.

PR #785, opened 2025-08-03
[#55] I Speak Their Language

### LINKED ISSUE #55: Implement "I Speak Their Language"
Depends on #51 

Language advancements uniquely can be expected to be delayed; should make sure UI showcases "unselected languages"

JPMeehan:
Improved flexibility of language advancements, allowing players to defer selection with a reminder element in the UI

image
image

Closes #55


krbz999 (commented):
I disagree with this implementation.

Despite the rules treating languages as a special case, we should go the extra mile and implement a generic notification on the sheet to remind users to make the choices of unresolved advancements, no matter what the type is.

Languages get a specific call-out in the rules - sure - but we won't be going afoul of the ruleset if we implemented something identical for eg skills or other traits.

As for the actual implementation of the language feature here:

The UX is slightly confusing getting 1 dialog to display all relevant advancements, which pops other dialogs for each advancement. We should either combine these into one, or look into having the initial dialog close itself once all advancements are resolved.

Ideally the changes would not happen until all advancements were resolved either.

JPMeehan:
I think what distinguishes "I Speak Their Language" is that they don't *all* need to be resolved together; if I've got five undeclared languages, I can allocate them individually across the duration of the campaign.

krbz999:
> I think what distinguishes "I Speak Their Language" is that they don't _all_ need to be resolved together; if I've got five undeclared languages, I can allocate them individually across the duration of the campaign.

I don't see that as a valuable distinction that necessitates a unique implementation.

JPMeehan:
UX-wise
- While nested dialogs is not *strictly* ideal, I don't really see a good way around it while also maintaining consistency; it's the same way that adding advancements is its own dialog that you then pop out to configure.
- I think having the "I Speak Their Language" close itself upon resolution is probably reasonable

As for Skill parity; would you basically just like a similar button in the Skills section?

krbz999:
> As for Skill parity; would you basically just like a similar button in the Skills section?

Essentially, yes. The ability to be notified of, and configure, any arbitrary advancement directly on the sheet is an overarching feature that would resolve #55 as well.

This is why I suggest that we simply do *that* rather than implement a bespoke feature explicitly for languages.

JPMeehan:
Updated to support skills

image

src/module/data/pseudo-documents/advancements/trait-advancement.mjs
```diff
@@ -112,14 +112,16 @@ export default class TraitAdvancement extends BaseAdvancement {
       const submit = dialog.element.querySelector(".form-footer [type=submit]");
       multiCheckbox.addEventListener("change", () => {
         for (const checkbox of multiCheckbox.querySelectorAll("input")) checkbox.disabled = !multiCheckbox.value.includes(checkbox.value) && (multiCheckbox.value.length >= chooseN);
-        submit.disabled = multiCheckbox.value.length !== chooseN;
+        // You don't have to pick all languages when they are granted
+        if (this.type !== "language") submit.disabled = multiCheckbox.value.length !== chooseN;
+        else submit.disabled = multiCheckbox.value.length > chooseN;
```
krbz999, line 117: I think it's fine to take the same approach for skills, too. Users can pick 1 of N and leave the rest for later.

```suggestion
        submit.disabled = multiCheckbox.value.length > chooseN;
```

src/module/data/pseudo-documents/pseudo-document.mjs
```diff
@@ -111,6 +112,16 @@ export default class PseudoDocument extends foundry.abstract.DataModel {
 
   /* -------------------------------------------------- */
 
+  /**
+   * Fetches the collection this PseudoDocument is contained in.
+   * @type {ModelCollection<this>}
+   */
+  get collection() {
+    return foundry.utils.getProperty(this.document, this.fieldPath);
```
krbz999, line 120: I don't think this will work in case of nested pseudo documents, and it would also be better to use the `getEmbeddedPseudoDocumentCollection` method.

Better yet, we make this a getter for a private property instantiated when the pseudo document is initialized. We do this already for the model, I think.

krbz999 (approved):
Code looks fine. UX is also fine, though I would like to see two changes happen.

Issue #259, opened 2025-03-03
Decide on implementation for Wards and similar features for non-kit classes


JPMeehan:
> E.g. features that offer a choice which can be swapped on a respite.

cdverrett94:
I wonder if a lot of these could be handled with the existing Feature item and Active Effects? We would need to allow for and resolve roll data in AE values.

That wouldn't help the conditional bonuses though I suppose.

JPMeehan:
95% handled by #653 with a touch of implementing some additional AE paths

PR #786, opened 2025-08-03
[#259] Added stamina.bonuses.echelon

JPMeehan:
Covers the Dwarf and a number of kit-style bonuses available within the system, e.g. the Elementalist's "Enchantment of Permanence"

### DIFF
--- a/src/module/data/actor/base.mjs
+++ b/src/module/data/actor/base.mjs
@@ -109,7 +109,12 @@ export default class BaseActorModel extends DrawSteelSystemModel {
       dsid: new Set(),
     };
 
-    this.stamina.min = 0;
+    Object.assign(this.stamina, {
+      min: 0,
+      bonuses: {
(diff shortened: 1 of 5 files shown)


src/module/data/actor/base.mjs
```diff
@@ -109,7 +109,12 @@ export default class BaseActorModel extends DrawSteelSystemModel {
       dsid: new Set(),
     };
 
-    this.stamina.min = 0;
+    Object.assign(this.stamina, {
+      min: 0,
+      bonuses: {
+        echelon: 0,
```
krbz999, line 115: I trust we have no use case at the moment for allowing roll data?

JPMeehan: Haven't seen one for applying to the actor themself. I think there is a use case for Applied effects, but that can be a one time parse when the effect is instantiated

PR #788, opened 2025-08-03
[#703] Add complications to features tab

### LINKED ISSUE #703: Add display for Complications

cdverrett94:
Closes #703 

In adding this, I think there's room to generalize some of the individual context types. Instead of having an `ActorSheetComplicationsContext`, `ActorSheetEquipmentContext`, etc that redfine the same typing, we could just have an `ActorSheetItemCategoryContext` that is just : `{
  label: string;
  item: ActorSheetItemContext[];
  showAdd: boolean;
  showHeader: boolean;
}`.  The ability context one could just extend the general and override the type for item. Let me know your thoughts. If you agree, I can do a separate PR for that.


JPMeehan (approved):
Yeah, I think by 1.0 we want to move to a more unified approach to our item lists, but right now we've been still figuring out what they need.

PR #789, opened 2025-08-03
[#713] Ability Modifier AE subtype

### LINKED ISSUE #713: Consider dedicated AE subtype for ability adjustments
An AE subtype that gets applied to actors (so not an enchantment) that modifies all abilities on the actor

Reasoning
1. Schema could include conditions, being able to restrict to certain keywords 
2. Cleaner data paths, as the structure could be defined entirely separately from the actor or item data.

Functionality covered
- Give abilities a new keyword, e.g. magic
- Give abilities with a keyword bonus damage, push distance, potency
- Allow abilities to ignore damage inmunities

JPMeehan:
- Open to adjusting the subtype name/label
- Keyword filters are sufficient for various kit-style bonuses, e.g. + damage for psionic abilities
- The _applyAbilityBonuses function is not super generalizable for arbitrary keys because of the complexity of abilities, including the use of sudokus. 
- Removed Reach because it was more annoying to calculate and doesn't appear to be standard to display now.

Creation
image

The filters section
image

Closes #713


templates/sheets/active-effect/details.hbs
```diff
@@ -0,0 +1,20 @@
+<section class="tab scrollable{{#if tab.active}} active{{/if}}" data-group="{{tab.group}}" data-tab="{{tab.id}}">
+  {{formGroup fields.tint value=source.tint rootId=rootId placeholder="#ffffff"}}
+  {{#if systemFields.filters}}
+  <fieldset class="filters">
+    <legend>{{localize "DRAW_STEEL.ActiveEffect.Filters"}}</legend>
+    {{formGroup systemFields.filters.fields.keywords value=system.filters.keywords options=keywordOptions}}
```
cdverrett94, line 6: The existing filters aren't showing when you load the AE config.

```suggestion
    {{formGroup systemFields.filters.fields.keywords value=source.system.filters.keywords options=keywordOptions}}
```

krbz999 (approved):
Only glanced at the docs.

Issue #765, opened 2025-08-01
Kits


JPMeehan:
- [x] Arcane Archer
- [x] Battlemind
- [x] Cloak and Dagger
- [x] Dual Wielder
- [ ] Guisarmier
- [ ] Martial Artist
- [ ] Mountain
- [ ] Panther
- [ ] Pugilist
- [ ] Raider
- [ ] Ranger
- [ ] Rapid-Fire
- [ ] Retiarius
- [x] Shining Armor
- [ ] Sniper
- [ ] Spellsword
- [ ] Stick and Robe
- [ ] Swashbuckler
- [ ] Sword and Board
- [ ] Warrior Priest
- [ ] Whirlwind

mike8912:
**Adding the following kits as a test:**
- Arcane Archer
- Battlemind
- Cloak and Dagger
- Dual Wielder
- Shining Armor

Once approved I'll add the rest

PR #792, opened 2025-08-04
[#765] Added Arcane Archer, Battlemind, Cloak and Dagger, Dual Wielder, and…

mike8912:
Five kits for #765


JPMeehan (approved):
Looks good, noticed some bugs in the advancement code while testing but that's a me problem not a you problem

mike8912:
image

@JPMeehan Is this the bug you are refering to? 
Because I think that may have been on my end.

Issue #768, opened 2025-08-01
Projects


JPMeehan:

cdverrett94:
I will tackle this. It will require adding a new compendium pack type, which I can handle.

PR #793, opened 2025-08-04
Add project and roll table compendiums with 1 item each

cdverrett94:
Adds project and roll table compendiums. Used the Build or Repair a Road project and its corresponding events roll table as the initial items in the compendiums.

Sets up further work for #768.


src/packs/tables/Project_Events_Tgb2Z68gf9SJ5VIg/tables_Build_or_Repair_Road_Events_tXtAQLmmJP1vHK6R.json
```diff
@@ -0,0 +1,300 @@
+{
+  "name": "Build or Repair Road Events",
+  "_id": "tXtAQLmmJP1vHK6R",
+  "img": "icons/svg/d20-grey.svg",
```
JPMeehan, line 4: Would appreciate an icon for the table, don't need to pick icons for the results

JPMeehan (approved):
Looks good, could you pick out icons for the documents?

cdverrett94:
Added the one I thought was best fitting. I used the same icon for the project and the event since they are tied to each other.

PR #801, opened 2025-08-06
Level up button

JPMeehan:
Implemented level-up button using existing `advance` API

image

### DIFF
--- a/src/module/applications/sheets/character.mjs
+++ b/src/module/applications/sheets/character.mjs
@@ -79,6 +79,9 @@ export default class DrawSteelCharacterSheet extends DrawSteelActorSheet {
   async _preparePartContext(partId, context, options) {
     await super._preparePartContext(partId, context, options);
     switch (partId) {
+      case "header":
+        context.showLevelUp = (this.actor.system.hero.xp > (this.actor.system.nextLevelXP ?? Infinity));
+        break;
       case "stats":
         context.skills = this._getSkillList();
         break;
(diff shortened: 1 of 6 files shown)


src/module/applications/sheets/character.mjs
```diff
@@ -78,6 +79,9 @@ export default class DrawSteelCharacterSheet extends DrawSteelActorSheet {
   async _preparePartContext(partId, context, options) {
     await super._preparePartContext(partId, context, options);
     switch (partId) {
+      case "header":
+        context.showLevelUp = (this.actor.system.hero.xp > (this.actor.system.nextLevelXP ?? Infinity));
```
krbz999, line 83: This is a good contender for some API on the actor. I'm thinking either a getter or derived value attached to `HeroData.hero`.

src/module/applications/sheets/character.mjs
```diff
@@ -276,6 +280,18 @@ export default class DrawSteelCharacterSheet extends DrawSteelActorSheet {
...
+   * Advance this character one level.
+   * @this DrawSteelCharacterSheet
+   * @param {PointerEvent} event   The originating click event.
+   * @param {HTMLElement} target   The capturing HTML element which defined a [data-action].
+   */
+  static async #levelUp(event, target) {
+    await this.actor.system.advance();
+  }
+
+  /* -------------------------------------------------- */
```
krbz999, line 293: Are we awaiting these calls for any particular reason?

krbz999 (approved):

src/module/applications/sheets/character.mjs
JPMeehan: I figure if some caller wants to know when the click listener is resolved they want the interior promise info.

src/module/applications/sheets/character.mjs
JPMeehan: Went with a getter, `advancementReady`

PR #802, opened 2025-08-06
Refactored "Replace items with advancements" logic

JPMeehan:
- Kit swap dialog will no longer show the current kit as a radio option
- Dropping an item as a replacement now goes through the "advancement deletion prompt" workflow

We had only been testing manually deleting then adding, the breaks were in this replacement workflow

### DIFF
--- a/src/module/documents/item.mjs
+++ b/src/module/documents/item.mjs
@@ -83,8 +83,8 @@ export default class DrawSteelItem extends BaseDocumentMixin(foundry.documents.I
    * An alternative to the document delete method, this deletes the item as well as any items that were
    * added as a result of this item's creation via advancements.
    * @param {Object} options
-   * @param {boolean} [options.replacement=false]   Should the window title indicate that this is a replacement.
-   * @param {boolean} [options.skipDialog=false]    Whether to skip the confirmation dialog, e.g. If there's already been another.
+   * @param {boolean} [options.replacement=false]   Should the window title indicate that this is a replacement?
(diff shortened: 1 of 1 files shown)


src/module/documents/item.mjs
```diff
@@ -82,9 +82,12 @@ export default class DrawSteelItem extends BaseDocumentMixin(foundry.documents.I
   /**
    * An alternative to the document delete method, this deletes the item as well as any items that were
    * added as a result of this item's creation via advancements.
+   * @param {Object} options
+   * @param {boolean} [options.replacement=false]   Should the window title indicate that this is a replacement.
+   * @param {boolean} [options.skipDialog=false]    Whether to skip the confirmation dialog, e.g. If there's already been another.
```
krbz999, line 87:
```suggestion
   * @param {boolean} [options.skipDialog=false]    Whether to skip the confirmation dialog, e.g., if there's already been another.
```

src/module/documents/item.mjs
```diff
@@ -82,9 +82,12 @@ export default class DrawSteelItem extends BaseDocumentMixin(foundry.documents.I
   /**
    * An alternative to the document delete method, this deletes the item as well as any items that were
    * added as a result of this item's creation via advancements.
+   * @param {Object} options
+   * @param {boolean} [options.replacement=false]   Should the window title indicate that this is a replacement.
```
krbz999, line 86:
```suggestion
   * @param {boolean} [options.replacement=false]   Should the window title indicate that this is a replacement?
```

krbz999 (approved):

PR #803, opened 2025-08-06
Revised unpacking script

JPMeehan:
- Purge non-system flags
- Normalize ownership like we did _stats
- Add _stats handling for grandchild items in anticipation of bestiary data


tools/pushLDBtoJSON.mjs
```diff
@@ -73,10 +73,18 @@ async function transformEntry(entry) {
     modifiedTime: null,
     lastModifiedBy: null,
   });
+  // Remove module flags
+  for (const key of Object.keys(entry.flags)) if (key !== "draw-steel") delete entry.flags[key];
```
krbz999, line 77: What about `core`?

krbz999 (approved):

tools/pushLDBtoJSON.mjs
JPMeehan: oh good call

Issue #761, opened 2025-08-01
Ancestries


JPMeehan:
- [x] Devil
- [x] Dragon Knight
- [x] Dwarf
- [x] Wode Elf
- [x] High Elf
- [x] Hakaan
- [x] Human
- [x] Memonek
- [x] Orc
- [x] Polder
- [x] Revenant
- [x] Time Raider

PR #804, opened 2025-08-06
[#761] Devils

RulyTafzil:
Quick vibe check. 
Added descriptions and feature items for the purchased traits. 
Unsure how to include purchased traits in advancements since 2 or 3 can be added depending on trait cost.


JPMeehan:
For now, leave the purchased traits out of the advancements; we'll do ancestry point buy in 0.8.1

RulyTafzil:
Made the listed changes, thanks for the detailed feedback and examples, that was very helpful. 

For wings, made an active effect for the fly and weakness. Rules say that weakness only applies at lvl 3 or lower, was not sure how to do that.

JPMeehan (approved):
Much better, just a small tweak for the speed AE

Issue #762, opened 2025-08-01
Cultures


JPMeehan:
- [ ] Typical Ancestry Cultures
- [ ] Archetypical Cultures

PR #806, opened 2025-08-06
[#762] Cultures

mrafaj:
Added cultures to the compendium, both Typical Ancestry and Archetypal.

Closes https://github.com/MetaMorphic-Digital/draw-steel/issues/762


JPMeehan (changes requested):
I think structurally appending the aspect, e.g. "Urban Environment" instead of "Urban", isn't actually what I want in the compendium.

Also, could you add in the source data? This should be easily fixable with a bulk update since it's all page 53.

mrafaj:
Updated

JPMeehan (approved):
Great work, should help smooth out character creation

PR #817, opened 2025-08-07
[#768] Add projects to compendium

### LINKED ISSUE #768: Projects

cdverrett94:
Closes #768 

This covers all projects and the related roll tables and abilities.

For the imbue treasure, I made an individual project for each enchantment type. Let me know your thoughts on the DSID for theses projects. Example: `imbue-armor-5th-level-armor-enhancement`. I wasn't sure if I should include the level in the dsid or just make it `imbue-armor`. I also gave a different icon for each level (1st, 5th, 9th) to help differentiate in the compendium a little easier.

There's also a lot of repeated text in these for the rules of Imbue Treasure and the rules for the specific enchantment type (armor, implement, weapon). At some point, it could be good to add a new system journal for the Heroes book with a journal per chapter. Then there could be a section that includes that information and we could just link it to that. Until then, this was the best way I thought of to make sure the information is easily available in the system.


JPMeehan:
Woah nelly this is a big one. Will take time to review.

cdverrett94:
Totally fair. The overwhelming majority of the files are the different imbue treasure enchantment types. Each type (armor, implement, and weapon) has 3-levels (1st, 5th, 9th) and each level has like 9 enchantments per type.

JPMeehan:
Practical note: I think that if you split the roll tables off from this PR, we could easily commit *those*

krbz999:
Yes, that seems fine.

JPMeehan:
Probably yeah

Issue #704, opened 2025-07-07
Implement Condition Immunities


JPMeehan:
Need to figure out the data path, but basically just purge `actor.statuses` in data prep of whatever conditions are immune.

metrzero:
After thinking about this for a bit, I'm not as sure that it's the right approach to simply filter out during preparation. Imagine a scenario like:
* Actor equips an item that grants immunity to Restrained.
* Actor is assigned Restrained, with a save ends duration.
* Actor removes item.

The result is a long-lingering restrained condition that has been present the entire time and now suddenly has an effect.

I think the current approach is good for wording such as "immune to the effects of" such as:
<img width="513" height="128" alt="Image" src="https://github.com/user-attachments/assets/3cc76841-8377-4e73-986e-c5f52095e3a2" />
I propose we call this "ignoring" conditions.

Compare with true immunity, such as the below:
<img width="539" height="105" alt="Image" src="https://github.com/user-attachments/assets/f73b228a-40d3-4c46-af91-2d880b192529" />

I think the latter needs to prevent application of the condition and clear existing conditions that conflict with the immunity when the immunity is assigned.

Thoughts?

krbz999:
I am not sure I see a meaningful distinction between immunity to a condition and ignoring a condition that would require us to do anything differently.

PR #819, opened 2025-08-08
[761] Memonek, Orc, Polder, Revenant, Time Raider

RulyTafzil:
Added effects/abilities/maneuvers where I could / where appropriate.


JPMeehan (changes requested):
Good stuff; makes it clear that #704 should get done in 0.8.1 or so

RulyTafzil:
Added an applied effect to the Orc Bloodfire Rush ability, and a link to the applied effect in the text. So users can drag that onto their character sheet. Not sure if there's a more convenient way to let users apply an effect on themselves?

Duration is set to 1 round, but it seems bugged. On the start of the 2nd round the effect expires as it should, but the bonus to speed remains. On the 3rd round the effect wears off as expected.

PR #820, opened 2025-08-08
Sudoku touchups

JPMeehan:
Some small touch ups to sudokus that came to my mind
- Item grants with a single item should default to that item's icon. `fromUuidSync` is guaranteed to provide an index with `img` because we prevent dropping items with parents.
- Made sure applied effects have all the right attributes lined up

### DIFF
--- a/src/module/data/pseudo-documents/advancements/item-grant-advancement.mjs
+++ b/src/module/data/pseudo-documents/advancements/item-grant-advancement.mjs
@@ -56,6 +56,20 @@ export default class ItemGrantAdvancement extends BaseAdvancement {
 
   /* -------------------------------------------------- */
 
+  /** @inheritdoc */
+  prepareBaseData() {
+    super.prepareBaseData();
+    // Item grants that are only granting a single item should have a matching icon
+    const hasDefaultImage = (this.img === ds.CONFIG.Advancement.itemGrant.defaultImage);
+    const hasOneGrant = (this.pool.length === 1);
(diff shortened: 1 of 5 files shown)


src/module/data/pseudo-documents/advancements/item-grant-advancement.mjs
```diff
@@ -56,6 +56,20 @@ export default class ItemGrantAdvancement extends BaseAdvancement {
 
   /* -------------------------------------------------- */
 
+  /** @inheritdoc */
+  prepareBaseData() {
+    super.prepareBaseData();
+    // Item grants that are only granting a single item should have a matching icon
+    const hasDefaultImage = (this.img === ds.CONFIG.Advancement.itemGrant.defaultImage);
```
krbz999, line 63: it can be null as well

src/module/data/pseudo-documents/advancements/item-grant-advancement.mjs
```diff
@@ -56,6 +56,20 @@ export default class ItemGrantAdvancement extends BaseAdvancement {
...
+    const hasOneGrant = (this.pool.length === 1);
+    if (hasDefaultImage & hasOneGrant) {
+      const indexEntry = fromUuidSync(this.pool[0].uuid);
+      this.img = indexEntry.img;
```
krbz999, line 67:
```suggestion
      if (indexEntry) this.img = indexEntry.img;
```

krbz999 (approved):

PR #824, opened 2025-08-08
Add project event tables

cdverrett94:
Related to #768 

Split off from #812 per request.


JPMeehan (approved):
Clean and simple. Generally, would prefer smaller & more focused PRs for compendium content anyways

PR #830, opened 2025-08-09
Data fix

JPMeehan:
- Re-ran unpack to fix the ownership of a bunch of previously entered docs
- Fixed kit foldering
- Fixed Career advancements


krbz999 (approved):
Seems fine. I wonder if we should also do something with `sort`.

PR #831, opened 2025-08-09
fix: disable sheet interactions for non-owners using {{disabled}} helper (fixes #744)

### LINKED ISSUE #744: Block certain sheet interactions from non-owners
Certain sheet interactions, e.g. rolling or using abilities, should only be available to owners. The `disabled` CSS class should cover this, which can be set up with the `{{disabled bool}}` handlebars helper

apanettiere:
Disables ability tab interactions (rolling, using abilities) for non-owners by applying the {{disabled}} Handlebars helper to relevant elements. This ensures that restricted actions are only available to sheet owners, aligning with system permissions. Left the Resources section untouched due to ongoing work in another PR.

### DIFF
--- a/templates/sheets/actor/shared/abilities.hbs
+++ b/templates/sheets/actor/shared/abilities.hbs
@@ -13,7 +13,7 @@
       <div class="item-column item-distance">{{@root.abilityFields.distance.label}}</div>
       <div class="item-column item-target">{{@root.abilityFields.target.label}}</div>
       <div class="item-column item-controls">
-        {{#if abilityType.showAdd}}
+        {{#if (and abilityType.showAdd @root.isOwner)}}
         {{#with (localize "DRAW_STEEL.SHEET.Add" itemName=(localize "TYPES.Item.ability")) as |addItemTooltip|}}
         <a class="createDoc" data-action="createDoc" data-document-class="Item" data-type="ability" data-render-sheet="true" data-system.type="{{@key}}" data-tooltip-text="{{addItemTooltip}}">
           <i class="fa-solid fa-plus"></i>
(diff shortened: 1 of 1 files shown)


templates/sheets/actor/shared/abilities.hbs
```diff
@@ -13,7 +13,7 @@
       <div class="item-column item-distance">{{@root.abilityFields.distance.label}}</div>
       <div class="item-column item-target">{{@root.abilityFields.target.label}}</div>
       <div class="item-column item-controls">
-        {{#if abilityType.showAdd}}
+        {{#if (and abilityType.showAdd @root.isOwner)}}
```
JPMeehan, line 16: Not sure the relevance here; `showAdd` should never be true for non-owners

templates/sheets/actor/shared/abilities.hbs
```diff
@@ -26,7 +26,7 @@
       {{#each abilityType.abilities as |abilityContext|}}
       <li class="item ability draggable" data-item-id="{{abilityContext.item.id}}" data-document-class="Item">
         <div class="item-row">
-          <div class="item-column item-name rollable" data-action="useAbility">
+          <div class="item-column item-name rollable" data-action="useAbility" {{disabled (not @root.isOwner)}}>
```
JPMeehan, line 29: Since this is a `div` not a `button` you need to have this as a css class rather than an attribute
```suggestion
          <div class="item-column item-name rollable {{disabled (not @root.isOwner)}}" data-action="useAbility">
```

JPMeehan (changes requested):
I'm not sure why this is only the abilities tab; this should also be addressing the Characteristics and Project Roll elements. Should probably also fix up the permission logic for Free Strikes

Issue #720, opened 2025-07-12
[FEature request] Elaborate on individual languages


KJTailor:
### User Experience

currently the languages section is just a long list of different languages. It would be nice if each language maybe has an explainer behind it what that language is, or who typically speaks it.

for example something like the below list:
- **High Kuric** - Language of Giants and Giant-kin
- **Caelian** - Common language of Orden
- **Tholl** - Gnoll
- **Variac** - Common Language of the World Below

For the Vasloria Human Languages it could just include a short comment on the region it is spoken in
- **Vaslorian** - Language spoken in Vasloria
- **Khemharic** - Language spoken in Khemhara
- **Oaxuatl** - Language spoken in Ix

For the Dead Languages, it should probably go something like:
- **High Rhyvian** - Dead Language of the Sun Elves - Related to *Hyrallic* and  *Yllyric*
- **Rallarian** - Dead Language of the Steel Dwarves - Related to *Zaliaz*

JPMeehan:
I think this is something that will be handled just fine once we get the compendiums, we can maybe investigate adding tooltips to the language advancement multi-checkbox

HugeFanboy:
With the advancements now in, it might be useful to section off the languages in some way (like having the Dead Languages in their own section/header), much like skills are.

PR #846, opened 2025-08-14
Add initial Polyglot support

### LINKED ISSUE #640: Add "Polyglot" compatibility
Requesting compatibility with [Polyglot](https://github.com/mclemente/fvtt-module-polyglot)

nonstopmorningtrolley:
Closes #640. Adds Polyglot support for Draw Steel's base languages in chat messages and journals.

Also fixes the English localization string for one language's name: `Vasloria` > `Vaslorian`, per Heroes p56.

Sorry for the low priority surprise! :bow: It wasn't until my "just to get familiar" experiments had made real progress that I noticed my faux pas. I'm aware of the current milestone now; I'll start getting familiar shortly with how to make compendium contributions and find some work to stake a claim on.

This is my first time making a public code contribution-- to a Foundry project, and in general. If you spot a(nother) faux pas or have a word of advice, I'd be grateful to hear it. I usually work with C#, so I'm sure I've missed something obvious with list manipulation or JSDoc here.

A few notes on Polyglot support, here and in general:
* This implementation uses the Polyglot API to register a game system's default languages during the module's initialization. Another option would have been to commit a LanguageProvider class like this to the module's own repo.
* Regarding languages' default fonts-- I've made an effort to preserve language relationships with these font choices (such as elvish languages sharing the same cursive-like wingding, all dwarvish languages using nordic rune-like wingdings, humans all get chicken scratch, etc.). But my only exposure to this lore is from the core rulebooks; I'm expecting I've made an oversight or two.
* There should be no impact to the broader system. This is a new method, called on a Polyglot hook to translate existing Draw Steel language entities into objects that Polyglot expects. There's no overriding or alteration to the system.
* :exclamation: It looks like the system isn't currently awarding PCs with the common tongue? I don't see Draw Steel currently awarding player characters Caelian, not unless they explicitly select it as part of advancements. Am I missing a character creation step, or a system config setting? If not, this might be a new issue.
  * Heroes p56, quote: "All player characters know Caelian!".
  * This might be related to #720. Discussion there floats the idea of changes to the language selection form element.
  * A quick and hacky fix would be to manually include Caelian in `getUserLanguages(actor)`, in this PR.
  * A more maintainable solution might be to always include Caelian in any language advancement, perhaps with a new system setting like "All PCs know Caelian?".
* Polyglot allows Directors to add custom languages. I haven't brought these custom languages into the base Draw Steel system here-- I haven't overridden  `ds.CONFIG.languages` to include custom languages, to allow actors to gain proficiency with them. That struck me as something to open a new issue on, for discussion.
  * Directors can still add custom languages right now via Polyglot, and can use their global language selector to speak/write in these languages. Players and actors can't currently gain proficiency in these custom languages, though.
* Polyglot allows systems to distinguish between "spoken" (chat messages & journals) and "written" (journals only) languages. Heroes p57 describes Dead Languages as only written... but then talks about how Khamish is still (with limits) spoken today. Rather than preventing Directors from having NPCs who speak these dead languages, I've marked all languages as "spoken" here.

### DIFF
--- a/src/module/helpers/polyglot.mjs
+++ b/src/module/helpers/polyglot.mjs
@@ -164,20 +164,18 @@ export function polyglotInit(LanguageProvider) {
      *         If this is something we want, and are able to do without breaking other setups, then this is probably the place to do it.
      */
     async getLanguages() {
-      const outputLangs = {};
-
       if (this.replaceLanguages) { // if module setting to only use custom-set languages is enabled
         this.languages = {}; // remove the rpg system's own list of languages from Polyglot. (No impact to PC feature selection options.)
       }
       const languagesSetting = game.settings.get("polyglot", "Languages"); // User-set fonts and scrambling overrides for system langs
-      for (let lang in this.languages) {
+      this.languages = Object.keys(this.languages).reduce((outputLangs, lang) => {
         outputLangs[lang] = {
           label: ds.CONFIG.languages[lang].label,
           font: languagesSetting[lang]?.font || this.languages[lang]?.font || this.defaultFont,
           rng: languagesSetting[lang]?.rng ?? "default",
         };
-      }
-      this.languages = outputLangs;
+        return outputLangs;
+      }, {});
     }
 
     /**
@@ -185,7 +183,7 @@ export function polyglotInit(LanguageProvider) {
      * Called during init on user's designated character, before lang list is fully loaded, to establish user's chat language options. 
      * Called also by Director during regular play, whenever selecting an actor on the canvas.
      * @param {DrawSteelActor} actor
-     * @returns [known_languages, literate_languages] Array of Set objects for spoken, written language keys for the actor.
+     * @returns {[Set<string>, Set<string>]} Tuple of Set objects for an actor's [spoken, written] language keys.
      */
     getUserLanguages(actor) {
       let known_languages = new Set(); // set of language keys; fluency with spoken language
--- a/src/module/helpers/polyglot.mjs
+++ b/src/module/helpers/polyglot.mjs
@@ -1,0 +1,221 @@
+import DrawSteelActor from "../documents/actor.mjs";
+
+/**
+ * Provide external add-on Polyglot with default settings for all of Draw Steel's in-game languages. Polyglot LINK: https://foundryvtt.com/packages/polyglot/
+ * The DrawSteelLanguageProvider class can be later accessed at `game.polyglot.api.providers[`system.${ds.CONST.systemID}`]`, after "polyglot.ready" Hook has fired.
+ * 
+ * This provider must be handed over to Polyglot during its initialization, or the module will fallback to the Generic language provider.
+ * Most of the class is built around providing the `languages` field as soon as possible (for chat message masking?) during world load, and
+ * then later fleshing that field out with localized labels and module config states.
+ * @param {LanguageProvider} LanguageProvider - Base languages configuration class provided by the Polyglot init event.
+ */
+export function polyglotInit(LanguageProvider) {
+  class DrawSteelLanguageProvider extends LanguageProvider {
+
+    requiresReady = true; // Ask Polyglot to wait to load until "ready" has fired/babele is done loading.
+
+    /**
+     * Polyglot-compatible list of all languages in Draw Steel, with their default configurations.
+     * Called during init, before localization is ready. Polyglot unconditionally requires the language-to-font relationship be available at this stage, though,
+     * or it will fallback to Generic language provider. `rng` and `label` are fully fleshed out as part of setup later.
+     * @typedef {Object}
+     * @property {string} label - The displayed name for the language, used in the UI.
+     * @property {string} font - The typeface name used to render language'd text in the UI. This is the full, well-formatted text of the Polyglot font selection dropdown list, spaces and all.
+     * Don't assume access to core Foundry fonts! Polyglot module setting to allow Foundry fonts is off by default.
+     * @property {string} rng - Enum-like. The text scrambling logic of the language'd text.
+     * "default" (consistent scrambling algo based on input string) / "unique" (identical text will not be identically scrambled) / "none" (use actual text with the wingding-like font; readable if a "real" font is used).
+     */
+    languages = {
+      // ancestry languages
+      caelian: { // Polyglot would like the first language provided to be the system's "Common" language
+        font: "Meroitic Demotic",
+      },
+      anjali: {
+        font: "High Drowic",
+      },
+      axiomatic: {
+        font: "Miroslav Normal",
+      },
+      filliaric: {
+        font: "Kargi",
+      },
+      highKuric: {
+        font: "Maras Eye",
+      },
+      hyrallic: {
+        font: "Ar Ciela",
+      },
+      illyvric: {
+        font: "Ar Ciela",
+      },
+      kalliak: {
+        font: "Kargi",
+      },
+      kethaic: {
+        font: "Semphari",
+      },
+      khelt: {
+        font: "Barazhad",
+      },
+      khoursirian: {
+        font: "Meroitic Demotic",
+      },
+      lowKuric: {
+        font: "Ork Glyphs",
+      },
+      mindspeech: {
+        font: "Saurian",
+      },
+      protoCtholl: {
+        font: "Tengwar",
+      },
+      szetch: {
+        font: "Kargi",
+      },
+      theFirstLanguage: {
+        font: "Mage Script",
+      },
+      tholl: {
+        font: "Tengwar",
+      },
+      urollialic: {
+        font: "Skaven",
+      },
+      variac: {
+        font: "Skaven",
+      },
+      vastariax: {
+        font: "Rellanic",
+      },
+      vhoric: {
+        font: "Maras Eye",
+      },
+      voll: {
+        font: "Celestial",
+      },
+      yllyric: {
+        font: "Ar Ciela",
+      },
+      zahariax: {
+        font: "Dark Eldar",
+      },
+      zaliac: {
+        font: "Floki",
+      },
+      // Human languages. Khoursirian already covered
+      higaran: {
+        font: "Meroitic Demotic",
+      },
+      khemharic: {
+        font: "Meroitic Demotic",
+      },
+      oaxuatl: {
+        font: "Meroitic Demotic",
+      },
+      phaedran: {
+        font: "Meroitic Demotic",
+      },
+      riojan: {
+        font: "Meroitic Demotic",
+      },
+      uvalic: {
+        font: "Meroitic Demotic",
+      },
+      vaniric: {
+        font: "Meroitic Demotic",
+      },
+      vasloria: {
+        font: "Meroitic Demotic",
+      },
+      // Dead languages
+      highRhyvian: {
+        font: "Ar Ciela",
+      },
+      khamish: {
+        font: "Jungle Slang",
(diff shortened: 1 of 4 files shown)


src/module/helpers/polyglot.mjs
```diff
@@ -0,0 +1,221 @@
...
+      }
+      const languagesSetting = game.settings.get("polyglot", "Languages"); // User-set fonts and scrambling overrides for system langs
+      for (let lang in this.languages) {
+        outputLangs[lang] = {
+          label: ds.CONFIG.languages[lang].label,
+          font: languagesSetting[lang]?.font || this.languages[lang]?.font || this.defaultFont,
+          rng: languagesSetting[lang]?.rng ?? "default",
+        };
+      }
+      this.languages = outputLangs;
```
JPMeehan, line 180: Style preference to use Object.keys & reduce, but improves consistency with the rest of the codebase. I'm writing this purely on gitHub so if you do take the suggestion please test & adjust for correctness.
```suggestion
      if (this.replaceLanguages) { // if module setting to only use custom-set languages is enabled
        this.languages = {}; // remove the rpg system's own list of languages from Polyglot. (No impact to PC feature selection options.)
      }
      const languagesSetting = game.settings.get("polyglot", "Languages"); // User-set fonts and scrambling overrides for system langs
      this.languages = Object.keys(this.languages).reduce((outputLangs, lang) => {
            outputLangs[lang] = {
            label: ds.CONFIG.languages[lang].label,
            font: languagesSetting[lang]?.font || this.languages[lang]?.font || this.defaultFont,
            rng: languagesSetting[lang]?.rng ?? "default",
          };
          return outputLangs;
      }, {});
```

src/module/helpers/polyglot.mjs
```diff
@@ -0,0 +1,221 @@
...
+    /**
+     * Returns Draw Steel language keys for a specific actor document, as translated into a Polyglot-compatible object.
+     * Called during init on user's designated character, before lang list is fully loaded, to establish user's chat language options. 
+     * Called also by Director during regular play, whenever selecting an actor on the canvas.
+     * @param {DrawSteelActor} actor
+     * @returns [known_languages, literate_languages] Array of Set objects for spoken, written language keys for the actor.
+     */
+    getUserLanguages(actor) {
+      let known_languages = new Set(); // set of language keys; fluency with spoken language
+      let literate_languages = new Set(); // set of language keys; fluency with written language.
```
JPMeehan, line 192: JS standard is to use camelCase over snake_case

src/module/helpers/polyglot.mjs
```diff
@@ -0,0 +1,221 @@
...
+      }
+      this.languages = outputLangs;
+    }
+
+    /**
+     * Returns Draw Steel language keys for a specific actor document, as translated into a Polyglot-compatible object.
+     * Called during init on user's designated character, before lang list is fully loaded, to establish user's chat language options. 
+     * Called also by Director during regular play, whenever selecting an actor on the canvas.
+     * @param {DrawSteelActor} actor
+     * @returns [known_languages, literate_languages] Array of Set objects for spoken, written language keys for the actor.
```
JPMeehan, line 188: bad type annotation, if you want to specify that this is a tuple (array of fixed length) you should still be using Set<string>
```suggestion
     * @returns {[Set<string>, Set<string>]} Tuple of Set objects for spoken, written language keys for the actor.
```

src/module/helpers/polyglot.mjs
```diff
@@ -0,0 +1,221 @@
...
+      const actorLangs = Array.from(actor.system.biography?.languages);
+
+      if (actorLangs) {
+        known_languages = new Set(actorLangs);
+      }
+      
+      return [known_languages, literate_languages];
+      // REVIEW: We could mark Dead Languages as written-only with `literate_languages` here. Polyglot would limit these languages to Journals.
+      //         But Heroes p57 immediately gives an exception to Dead Langauges being written-only, with Khamish described as still having a spoken niche.
+      //         Is it desirable behavior to limit some/all Dead Languages to Journals?
```
JPMeehan, line 203: I don't think there's not a strong reason to impose that given the exceptions.

src/module/helpers/polyglot.mjs
nonstopmorningtrolley: I've implemented this without issue. :thumbsup:

src/module/helpers/polyglot.mjs
nonstopmorningtrolley: The method being overridden in Polyglot here uses these specific names-- `known_languages` and `literate_languages`. Other language implementations in their repo continue to use that format, for just this method's return values. I'll leave these names as-is for now, under the assumption that I'll be moving this class to a new Polyglot PR at some point in the near future.

If we keep this API call implementation though, I'll adjust these names ASAP. Leaving this thread unresolved for now.

nonstopmorningtrolley (commented):
> Some code feedback, but I'm unsure about this from a strategic perspective; I think there's a good argument this should just be submitted to the Polyglot repo like most other systems have. 

I hear you, @JPMeehan! This LanguageProvider class should take few changes to work within Polyglot's repo, and I'm happy to make that happen.

I'll wait for this PR to get reviewed as rejected before opening anything up in Polyglot's repo, though!

JPMeehan:
To address some separate pieces
1. I think one option for the "All PCs know Caelian" piece would be to add it to CharacterModel#_preCreate; maybe add in some property to CONFIG.DRAW_STEEL with a set of default PC languages, similar in concept to what we do for default items but less complicated
2. Is there some hook or setting that tracks user-added languages, and can that integrate into our CONFIG object? How does Polyglot handle this for dnd5e?

nonstopmorningtrolley:
As promised, @JPMeehan! Some updates on custom languages support and Polyglot-dnd5e.

### dnd5e
* Base Polyglot behavior adds custom languages to `CONFIG[systemId.toUpperCase()].languages.push({custom1: "Custom1"})`, falling back to `CONFIG[systemId.toUpperCase()].languages[custom1] = "Custom1"`.
  * This fails (null reference) for Draw Steel since the system's CONFIG key is instead rendered as `DRAW_STEEL`, with an underscore. Polyglot's attempt on this is to the non-existant `CONFIG[DRAW-STEEL].languages`, with a hyphen.
  * We can completely override where and how custom languages from Polyglot arrive in-system by overriding `LanguageProvider.addToConfig()` and `.removeFromConfig()`, which are expressly _for_ adding and removing just custom languages to a system's config. This would all be work for the Polyglot repo-- Draw Steel's repo would just need to first have a safe home for where custom languages can go and be treated legitimately by the system.
* dnd5e works fine with Polyglot's base behavior-- those additions Polyglot makes to `CONFIG[DND5E].languages` work just fine. Custom languages appear under "Other Languages" in their language picker UI, and are selectable within the character sheet as languages to become proficient in.
  * Per-actor custom language data is saved as though they were system languages: all are saved in the same array of language keys on their `actor.system.traits.language.values`. There's zero notation or extra property assignments done to explicitly flag these proficiencies as custom.
    * Localization labels on dnd5eActors are kept separately in `actor.system.traits.language.labels.languages`. The two lists are not ordered the same; relationship isn't strongly tied.
    * If Polyglot is uninstalled after a 5e character gains proficiency, the custom languages' keys remain on the actor. Despite the proficiency remaining, their UI labels fail some kind of localization pass, I think--they don't appear on the character sheet. No `.labels.languages` value is assigned for custom languages on world init in this scenario, so no UI label for the languages is available.
      * Actually, hang on, this is happening for 5e with Polyglot _installed_, too. An already-proficient custom language will reappear on the character's sheet if it the language is removed and re-added from the sheet's UI picker. Again, no impact to actual proficiency in this scenario-- PCs retain their custom language proficiencies in chat messages and journals while this visual bug is ongoing.
  * dnd5e _would_ be able to tell the difference between their own languages and the in-system ones, if ever they checked. Their CONFIG.languages object divvies in-system languages between "normal" and "exotic". With Polyglot's custom languages present, their languages object looks like: (normal { common, dwarvish, ... }, exotic { deepspeech, telepathy, ... }, Custom1, Custom2 ).
    * They don't actively _try_ to tell the difference, not that I can see. Their Polyglot lang provider's isCustomLanguage()-like checks are just "does the given language exist within the CONFIG's normal or exotic language sets?" I'm still wrapping my head around handlebars, though, and may have missed some logic in those 5e files.
    * JS _within_ the dnd5e system itself uses `.flatMap()` or an equivalent on CONFIG[DND5E].languages. They do this to get a one-dimensional array of all languages--removing their ability to tell what is and isn't custom. This approach benefits more than just Polyglot-- world scripts, adventures, and other modules are all be able to use `CONFIG[DND5E].languages` as a bit of a dumping ground for custom languages.

### Lessons for Draw Steel
Because Polyglot explicitly provides public, overrideable methods for adding custom languages to a system's configs, we have complete control over where and how custom languages from it appear and get used across the system. We're not limited by any particular Polyglot architecture, and can insert our grubbly little hooks wherever we want, muaha*ha*.

That said, some takeaways:
* Having a way to cleanly distinguish between in-system languages and custom languages would be ideal.
  * The fewer assignments outside scripts are making to the internal game state/configuration, the better~! :notes:
  * Custom languages have needs that in-system ones don't, like their unlocalizable labels.
  * We'll want to visually differentiate between human/various/extinct languages someday, in advancement form inputs or on the sheet UI. On such a day, we'll want to be able to visually differentiate from "other"/unrecognized languages, too. Otherwise, to make their custom languages function right, outside sources may attempt to dress up their additions as in-system-- something that risks side effects.
* dnd5e has already made explicit allowances for custom languages. That's their use of `.flatMap()` in their language getters has explicit checks for floating, category-less language objects in `CONFIG[DND5E].languages`. They wouldn't need that, except for custom languages' benefit. If Draw Steel wants to gracefully permit and natively handle custom languages from outside sources, then _some_ kind of explicit dumping ground and knows-what-to-expect handling for custom languages may be appropriate.

Options for the Draw Steel system that come to my really-not-familiar-with-the-project-yet mind:
* Explicitly allow custom languages from outside sources. Create a space for them, a best practice. Devs would have to remember here that custom languages exist when working on languages' and actor's languages' getters and setters.
  * Maybe having dedicated locations, such as `CONFIG[DRAW_STEEL].languages.custom` and `actor.system.biography.languages.custom`, to store both custom languages' keys and labels.
    * But that would still result in duplicate data-- language labels getting stored 2+ times, in `CONFIG` and each proficient character's actors' data, ugh.
  * We could have a new property for  language types, with a new enum-like property on languages-- as much to label "custom" languages, as to futureproof for human/various/extinct language differentiation. But that would require changing the existing simple collection of language keys stored on actors.
* We could wash our hands of it. Outside modules are responsible for whatever they get up to, Polyglot included.

For now, I'm planning on continuing development on this Polyglot support. Unless I hear from you, I'll take a stab at `CONFIG[DRAW_STEEL].languages` additions for custom languages' home.

nonstopmorningtrolley:
Thanks @JPMeehan; I needed the course correction here. I'll be trying to wrap my mind around everything yet. (And taking my foot off the talking pedal.)

I'm struggling with implementing custom languages. Selecting one during advancement is possible with the changes talked about, but they disappear from actors on browser refresh/world startup. I plan to ask for help in Polyglot circles--I figure the trouble is generic to however Foundry and these saving these embedded documents to actors work.

Will get this done yet. :muscle:

JPMeehan:
Seems like somehow you're only making temporary changes? Do you have an open PR somewhere I could look at?

PR #888, opened 2025-08-20
Update Current Stamina based on Max HP Update

ambiguousname:
In `0.7.3`, I kept experiencing issues with an actor's current stamina always being set to 20 regardless of how the max stamina was set. This adds some code to adjust the current stamina if the maximum has been changed (and the current stamina has been set to the max).

In my experimenting, also setting temporary stamina didn't update the actor's counter either. I'm happy to address that as well (or make that a separate PR).


JPMeehan:
Please open a ticket to discuss this, the potion of growth is a major complicating factor while trying to decide the best way to automate stamina handling.

PR #889, opened 2025-08-20
Added color to class folders

JPMeehan:
Color-coded the top level class folders to help them pop a bit more and blend less into each other.

image


krbz999:
We simply readjust the gradient delta. Not a problem :)

PR #910, opened 2025-08-22
Tactician L1 cleanup update

Voltan-rev1:
Added icons, corrected some ability targets (ally instead of creature, etc) re-ordered items and tried to make a standardization pass on some advancement names. Moved some abilities to other folders to hopefully aid in organization.


JPMeehan (approved):
Great work

PR #916, opened 2025-08-22
Refactored subclasses to allow multiple

### LINKED ISSUE #906: Improve Conduit Subclass Display
> After choosing your deity, pick two domains from their portfolio. The two domains you pick make up your subclass, and your choice of domains determines many of the features you’ll gain as you gain new levels.

Currently, Conduits end up with "No Subclass" in the header. It would probably be preferable to find a way to treat the domains as subclasses. One option would be to remove the singular cap on subclasses and then just turn them into subclasses, then the header would display all subclasses on a character

JPMeehan:
Adjusted logic to allow for multiple subclasses + adjusted the Conduit so their domains are now considered subclasses.

image

Closes #906


krbz999 (approved):
OK. Code looks fine. I had a quick skim through the class, and couldn't come up with an alternative other than having the Deity be the subclass or having no subclasses but instead Features granting Features, neither of which seem right.

I think it would be good to have the class determine whether multiple subclasses is allowed, however. A simple boolean property.

PR #919, opened 2025-08-22
L1 Dmg weakness ability fix [SHD+CON]

Voltan-rev1:
Noticed two abilities that could use a fix for damage weakness.
Conduit's - Corruptions Curse
Shadow's - Setup

Attempted to apply @damage.weaknesses.all as the likely resolution.


JPMeehan (approved):
For active effects, the key should start with `system.` rather than `@`

PR #923, opened 2025-08-22
[#890] Swap `stamina.value` for `stamina.spent`

### LINKED ISSUE #890: Current Stamina Should Reflect Max Stamina Changes
Following from discussion in #888, changes to max stamina should be reflected in current stamina as well. Currently, every token placed has a current stamina of 20, regardless of the max stamina.

krbz999:
Removes `stamina.value` in favor of `stamina.spent`, which has a minimum and initial value of `0`, and no maximum value.

The `spent` property is not capped by `stamina.max`, allowing for damage to overflow as before.

Token attribute bars were tested and should already work.

The base actor sheet processes `system.stamina.value` in the form data to map to `system.stamina.spent`.

The result is that if you have lost X stamina, and your maximum changes, the difference between the maximum and your current stamina is still X. This is easy to verify on an npc actor.

Closes #890.

### DIFF
--- a/src/module/applications/sheets/actor-sheet.mjs
+++ b/src/module/applications/sheets/actor-sheet.mjs
@@ -1033,6 +1033,17 @@ export default class DrawSteelActorSheet extends DSDocumentSheetMixin(sheets.Act
   }
 
   /* -------------------------------------------------- */
+
+  /** @inheritdoc */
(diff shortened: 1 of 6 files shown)


JPMeehan:
Logic checks out, main concern remains needing to get a ruling on Growth Potion

PR #936, opened 2025-08-23
Set `stamina.value` to be derived from `stamina.max` while it is undefined

ambiguousname:
Possible fix for #890.

Currently this sets `stamina.value` to `undefined` initially, and matches it to `stamina.max` if it is still set to `undefined`.

### DIFF
--- a/src/module/data/actor/base.mjs
+++ b/src/module/data/actor/base.mjs
@@ -26,7 +26,7 @@ export default class BaseActorModel extends DrawSteelSystemModel {
     const schema = {};
 
     schema.stamina = new fields.SchemaField({
-      value: new fields.NumberField({ initial: 20, nullable: false, integer: true }),
+      value: new fields.NumberField({ initial: null, nullable: true, integer: true }),
       max: new fields.NumberField({ initial: 20, nullable: false, integer: true }),
       temporary: new fields.NumberField({ initial: 0, nullable: false, integer: true }),
     });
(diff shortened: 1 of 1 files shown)


krbz999:
We've discussed looking into whether we can apply conditional logic as you do here, but not for when the value is `undefined` but when it is `null`. 

This would involve making `stamina.value` `nullable: true`, and when null it should be understood as the actor being undamaged, and their current stamina following the max stamina.

JPMeehan:
Are you still interested in working on this PR? I think overall this is the correct direction, but it's going to need some form handling.

ambiguousname:
Still interested, although it might be some time before I can get back to this; I've got a pretty busy schedule for the next two months or so.

src/module/data/actor/base.mjs
```diff
@@ -153,7 +153,12 @@ export default class BaseActorModel extends DrawSteelSystemModel {
 
     // Apply all stamina bonuses before calculating winded
     this.stamina.max += this.echelon * this.stamina.bonuses.echelon;
-    this.stamina.max += this.level * this.stamina.bonuses.level;
+
+    // If our current stamina has not been set, match it to max:
+    if (this.stamina.value === null) {
+      this.stamina.value = this.stamina.max;
+    }
```
JPMeehan: Seems like null coalescing can simplify this?
```suggestion
    this.stamina.value ??= this.stamina.max;
```

src/module/data/actor/base.mjs
```diff
@@ -153,7 +153,12 @@ export default class BaseActorModel extends DrawSteelSystemModel {
 
     // Apply all stamina bonuses before calculating winded
     this.stamina.max += this.echelon * this.stamina.bonuses.echelon;
-    this.stamina.max += this.level * this.stamina.bonuses.level;
```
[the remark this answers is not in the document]
ambiguousname: No, that's an artifact from rebasing. Just reintroduced it.

src/module/data/actor/base.mjs
ambiguousname: Done!

PR #947, opened 2025-08-24
Minor fixes for monster values

Enrahim:
Compared some key values for the monsters with values gotten from parsing steelcompendium.io monster blocks.
Found some values that were off. Confirmed with hero book, and errata.


JPMeehan:
Could you comment on #981 for assignment?

PR #955, opened 2025-08-25
Elementalist compendium

AdellRedwinters:
Updated to 2nd level


JPMeehan (approved):
Spun off a new ticket because while I don't love how Disciple of the Green is implemented, I'm not sure there's much better ways. #983

PR #958, opened 2025-08-25
Null compendium

AdellRedwinters:
Updated with 2nd level features and abilities.


JPMeehan (approved):
The deletions of the level 1 features did not get committed properly

PR #959, opened 2025-08-25
Shadow 2nd Level Features

AdellRedwinters:


JPMeehan (approved):
I threw in the fix for #980 as part of incrementing the revision number.

PR #965, opened 2025-08-25
[#841] Source Datalists

### LINKED ISSUE #841: Add datalists for Sources panel
As a content creator, I would like a data list to speed up source entry for items and NPCs.
- Book: "Heroes" and "Monsters"
- License: "Draw Steel Creators License"

Possibly make an extendible list in CONFIG

JPMeehan:
- Creates CONFIG entry for books & licenses
- Adds datalists to the source inputs
- Creates static method to parse source info from module manifests

Closes #841


src/module/data/actor/npc.mjs
```diff
@@ -92,10 +92,34 @@ export default class NPCModel extends BaseActorModel {
...
+        const defaultBook = foundry.utils.getProperty(m, "flags.draw-steel.defaultBook");
+        if (defaultBook) foundry.utils.setProperty(updates, "source.book", defaultBook);
+        const defaultLicense = foundry.utils.getProperty(m, "flags.draw-steel.defaultLicense");
+        if (defaultLicense) foundry.utils.setProperty(updates, "source.license", defaultLicense);
+      }
+    }
+
+    if (!foundry.utils.isEmpty(updates)) this.updateSource(updates);
+  }
+
```
krbz999, line 116: Useful, granted, but seems more like we just want to run a one-time script each release.

krbz999 (approved):
I pushed a commit to add some validation for module-added books and licenses, let me know what you think. The rest looks good, though I still do not agree we need (or want) `source.revision`. I cannot imagine that anyone would want to invest the effort of keeping that updated.

krbz999:
> I was planning to eliminate source.revision in a follow-up PR, but I can do that in this one if you'd prefer.

OK, sounds good. A follow-up is cleaner.

PR #968, opened 2025-08-25
Conduit 2nd Level Features and Abilities

AdellRedwinters:
Wasn't sure how to handle the ability choice at level 2 so for now all of the subclass abilities are just in the Class advancement.


JPMeehan (approved):
Minor source cleanup + fixing the unintended reversions

PR #969, opened 2025-08-26
Tactician Compendium 2nd Level

AdellRedwinters:
2nd level features and abilities for the tactician.


JPMeehan (approved):
I also took the liberty of moving the Tactician subclasses out of their folders since we decided to do class => feature => subclass for the subclass features

PR #978, opened 2025-08-29
Bugfix: Talent's Distance Augmentation

Santreim:
According to the text:

> Your ranged psionic abilities gain a +2 bonus to distance.

Changed the "Magic" Tag to "Psionic", and the bonus to +2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,9 +26,13 @@
 
 ### Changed
 
+- The skill list on the character sheet is now sorted alphabetically.
+
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


JPMeehan:
Looks good, however I'll note that you usually want to make your own branches locally to merge rather than commit directly to `develop` because you'll get conflicts when you pull back from origin due to our use of `squash`

Santreim:
Yeah I realized too late. I'll fix my fork when it comes to that, not a big deal, but thanks for the heads up!

JPMeehan (approved):

PR #985, opened 2025-08-31
Censor 3rd level features and abilities

AdellRedwinters:
7 wrath abilities.


JPMeehan (approved):
The enrichers weren't quite what I wanted format-wise; the main thing is that even though Draw Steel may use the "Bonus Damage" wording, that means you need to make an enricher for the *total* damage to make sure it's interacting correctly with immunities and weaknesses.

PR #1001, opened 2025-09-03
Implemented Characteristic Advancement

### LINKED ISSUE #707: Implement Characteristic Advancement
Initial Characteristics have two setups, model A and model B, depending purely on your choice of class

A. You start with a X of 2 and a Y of 2, and you can choose one of the following arrays for your other characteristic scores:
- 2, −1, −1
- 1, 1, −1
- 1, 0, 0

B. You start with a Z of 2, and can choose one of the following arrays for your other characteristics scores:
- 2, 2, −1, −1
- 2, 1, 1, −1
- 2, 1, 0, 0
- 1, 1, 1, 0

After Level 1, the increases consistently are paired with the rise in Echelon

Model A
- Level 4: Your X and Y scores each increase to 3.
- Level 7: Each of your characteristic scores increases by 1, to a maximum of 4.
- Level 10: Your X and Y scores each increase to 5.

Model B:
- Level 4: Your Z score increases to 3. Additionally, you can increase one of your characteristic scores by 1, to a maximum of 3.
- Level 7: Each of your characteristic scores increases by 1, to a maximum of 4.
- Level 10: Your Z score increases to 5. Additionally, you can increase one of your characteristic scores by 1, to a maximum of 5.

Note: Characteristics are pretty hard to muck with in Draw Steel, there's no treasures or anything, and we may want to be able to reference them in active effects (see #699)

JPMeehan:
Closes #707

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,8 @@
   - War Dogs Echelons 2–4
 - Added `ds.utils.updateFromCompendium`, which currently supports Active Effects, Actors, and Items. (#557)
 - New Advancement Type: Characteristic. (#707)
+  - Characteristic advances allow classes and titles to increase characteristics.
+  - These increases are not saved to the base data of the actor and are instead applied dynamically.
 - Added an "Additional" configuration to Item Grant advancements, which allows players to drop in items. (#708)
 - Added suggested books and licenses in the source input form. (#841)
 - In the Power Roll Dialog, when hovering a target's name and modifiers, the target's token will be highlighted as well.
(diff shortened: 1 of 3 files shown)


krbz999:
I like the UX, though I'd be interested to see if we can show each characteristic only once in the setup of the advancement, choosing whether it is guaranteed, an increase, or neither.

As for us being unable to rewind, did we consider performing these characteristic changes during data prep *instead* of an update to the actor? We do this with skills and languages, I am not certain there is a reason we cannot also do it with characteristics.

JPMeehan:
UI + Data Model Revision: Store data per characteristic, instead of two sets

image

Furthermore, refactored to use data prep instead of updates. Future PR will implement revised characteristic configuration as a helper app that can show characteristic advances. Expectation for 0.9 is that we will cover post-level 1 characteristic increases, but the initial characteristic assignments will need separate work. It's possible those can get handled as part of just the class model? We have some data there for `characteristics.core`.

PR #1003, opened 2025-09-04
Kobolds for Monster Compendium

AdellRedwinters:
All of the Kobolds found in the Monsters Book


JPMeehan (approved):
Shield Shield needed an effect

PR #1018, opened 2025-09-05
Basilisks Monsters and Malice Features

AdellRedwinters:
Note: The Upchuck malice feature has a Potency related damage effect that only happens on tier 2 and 3 results, but our current way of displaying damage requires me to put a zero damage entry in tier 1, so it looks a little weird atm.


JPMeehan (approved):
Point taken on the Upchuck

Issue #708, opened 2025-07-07
Implement Choose Item Advancement


JPMeehan:
Two cases
1. Ancestry point buy
2. Class/Subclass choose ability

Unlike Item Grant advancements, need to accommodate drop/search/filter arbitrary items.

It is possible that this requires building an initial version of the #130

krbz999:
I think it's important to outline how this differs from a regular Item Grant being given the ability to add an outside item.

JPMeehan:
I would be amenable to simply expanding the schema of Item Grant advancements if you think that will work best.

PR #1020, opened 2025-09-05
[#708] Implemented Item Grants taking drops

JPMeehan:
Requesting review now even though this is not complete - I would like this PR to *also* include compendium data updates, however I would first like to request validation of the data model implementation.

I decided to name the schema "expansion" because in the future this is unlikely to work purely on drag n drop and instead go through the Compendium Browser.

image

Once dropped in, they just get added to the list.

image


krbz999:
The fieldset legend says "Choose null" in some instances. Might be that this is all linked to advancements that grant all the options.

image

I reconfigured the advancement in question to be "choose 2" and things worked as expected.

JPMeehan:
What is the expected behavior for null + "additional"? I think that it would be "Choose Any", making basically everything optional (since you might drop more than you want on, and being able to check/uncheck works well enough as an "ignore" option)

krbz999:
> What is the expected behavior for null + "additional"? I think that it would be "Choose Any", making basically everything optional (since you might drop more than you want on, and being able to check/uncheck works well enough as an "ignore" option)

If it is pre-configured to be granting all items, then it should grant all items.
I am not sure this is a valid use case at all, can you explain further.

PR #1028, opened 2025-09-06
Wode Elf Monsters and Malice Features

AdellRedwinters:
All of the Wode Elves and their Malice features.


JPMeehan (approved):
Overall good, needed to adjust the Greenkeeper a bit

PR #1038, opened 2025-09-07
Radenwight and Goblin With Captain feature

AdellRedwinters:
added a "with captain" effect on the Minions for these monster groups, and also made the missing Radenwight monsters from the Monster book.

note: I noticed that on the Radenwigh Mischiever the ability modifier to damage does not apply correctly if its signature ability has both the ranged and melee keyword, even though the ability modifier is only being told to look for the Strike keyword.


JPMeehan (approved):
Worked in the logic fix for the "strike" bonuses

PR #1047, opened 2025-09-08
Ogre Compendium

AdellRedwinters:
renamed Giant's folder to Ogre, and added the remaining monsters from this section of the Monster book.


JPMeehan (approved):
Good work

PR #1052, opened 2025-09-09
[#973] Add combat group visibility

### LINKED ISSUE #973: Add a way to make a Combatant Group hidden from players in the encounter combat tracker
### User Experience

Hi, 

It would be helpful to be able to hide a combatant group in the combat tracker, either with a manual toggle like there is on the actors, or automated so that if all the actors inside the group are invisible then the group becomes invisible too.  

A trivial use case is the first encounter in the Delian Tomb, where there are two groups that are initially out of sight and unknown to the players, not showing up until round 2.  I had the goblins for those groups in the tracker while hidden, but I couldn't find a way to hide the groups so ended up having to create the groups in a short break between rounds. It wasn't a big deal, with just two groups of two monsters each, but I can imagine more complicated scenarios where it would be much harder to do that on the fly.  Being able to prepare the additional groups in advance then hide them until the reveal would be great.

cdverrett94:
Closes #973 

- `foundry.documents.CombatantGroup` sets `hidden` as a property on the class and then sets it during data prep. I've overridden `hidden` as a getter instead. I'm not sure if that's acceptable practice or not.
- Copied the CSS for hidden combatant.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -638,7 +638,8 @@
         "SquadOneCaptain": "A squad can only have one captain.",
         "MinionNoSquad": "Applying damage directly to a minion because it's not in a squad.",
         "TooManySquad": "Applying damage directly to a minion because it's in multiple squads."
-      }
+      },
+      "ToggleVisibility": "Toggle Visibility"
     },
     "Item": {
       "Tabs": {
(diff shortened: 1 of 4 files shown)


JPMeehan (approved):
Good work!

PR #1057, opened 2025-09-10
[#1054] Add damage/heal modification dialog

### LINKED ISSUE #1054: Add damage modification option to apply damage button
### User Experience

As a Director, I would like an option to apply ad-hoc damage modifications when clicking the apply damage button. This would allow application of hard to automate things as well as stuff that's not currently automated like surges.

cdverrett94:
Closes #1054 

I'm not sure if how I'm grabbing the old roll data is the best way to do it. Core doesn't save the roll data when the chat message is created, so I'm having to grab it from the ability or the message. One way to easily solve this would be to override `DamageRoll#toJSON` to include the the roll data, but I wasn't sure if that was desirable or not.

I'm creating the new damage roll as a message just so that it's clear what new damage/heal total will be post modification.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 - Added Combatant Group tracker visibility controls. (#973)
   - If all combatants in a group are hidden, the group is hidden.
   - Added context menu option to quickly hide/unhide each combatant in the group.
+- Added an Alt-Click option on the Apply Damage/Heal buttons to allow modifying damage/heal amount and damage type before applying. (#1054)
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


cdverrett94:
I'll set this is a draft for now then until specifics are decided.

cdverrett94:
@krbz999 Assuming this is even greenlit based on discussions on the issue, but thoughts on something like this instead of the keyboard modifier or the context menu? Open to whatever suggestions.

Screenshot 2025-09-10 174025

JPMeehan:
Probably, odd that mine got retargeted

cdverrett94:
Fair enough. I can either just open a new draft PR or I can wait until more discussion on the relevant issue happens to decide on how to handle keyboard modifer, context menu, or something like I presented later.

JPMeehan:
Do a draft PR to focus the discussion

Issue #780, opened 2025-08-01
Titles


JPMeehan:
- [x] 1st-Echelon
- [x] 2nd-Echelon
- [x] 3rd-Echelon
- [x] 4th-Echelon

PR #1073, opened 2025-09-12
#780 - 1st Echelon Titles

Voltan-rev1:
Added Titles for #780 for 1st Echelon.

Most of this is standard copy/paste but Titles and Abilities therein for the following may need closer review.

- Elemental Dabbler (Ability and Effects)
- Marshal (Granting of Censor ability)
- Ratcacher (Ability grant)
- Zombie Slayer (Ability grant and description. Possibly missing needed enrichment?)


JPMeehan (approved):
Main thing is you could've done more active effects. For example, a default-inactive effect works well for conditional boosts like "Close Formation"; players can toggle it on and off as they move in and out of position.

PR #1074, opened 2025-09-13
Animals Compendium and Features

AdellRedwinters:
This PR has all the pre-made animals and all of the features which have been added to the Monster Features Compendium


JPMeehan (approved):
I adjusted the naming scheme to remove the point values from the names - frankly, I expect the point values to either get shuffled off into just the description of a Journal and/or get shuffled into the flags. We probably could also ship an "Animal Builder" macro or something.

Issue #779, opened 2025-08-01
Treasures


JPMeehan:
- [x] Consumables
  - [x] 1st Echelon
  - [x] 2nd Echelon
  - [x] 3rd Echelon
  - [x] 4th Echelon
- [x] Trinkets
  - [x] 1st Echelon
  - [x] 2nd Echelon
  - [x] 3rd Echelon
  - [x] 4th Echelon
- [x] Leveled Treasures
  - [x] Armor
  - [x] Implements
  - [x] Weapons
  - [x] Other
- [ ] Artifacts

PR #1077, opened 2025-09-13
Treasures #779

NagiSungila:
- Icons added
- Missing Trinkets added
- Book information as been propogated


JPMeehan (approved):
You missed a step when you pulled my changes and didn't `npm run build:packs`, which meant that my fixes like adding enrichers to the text weren't included, so I had to redo some of that work.

PR #1084, opened 2025-09-15
Fix Beggar career granting all interpersonal skills

cdverrett94:
The Beggar career didn't have any value set for `chooseN` on the interpersonal skills.


JPMeehan (approved):
Took advantage of this PR to also make some updates to the Changelog

PR #1091, opened 2025-09-16
[#1087] Remove invalid roles from leader and solo monsters and harden template against missing roles/organizations

### LINKED ISSUE #1087: Solo monsters should not show blank entry for missing role
Solo monsters have a missing - entry in their keywords space because they don't have a role. We should just not show the "empty" slot if it's not filled in.

Image

cdverrett94:
Closes #1087 

Looks like the issue is largely because some monsters had a solo or leader role somehow despite that not being a valid role from the list. I've removed those instances. 

I also updated the template to filter out anything without a role/organization label. Will prevent against cases like this or ones where a module adds custom roles/organizations and is later removed.


JPMeehan (approved):
Good investigation

PR #1094, opened 2025-09-16
[#1081] Blank tier display improvements

### LINKED ISSUE #1081: Improve display of "blank" PREs
There are a few occasions where "blank" PREs create additional semicolons
- Effects that only apply at some tiers
- Conditional damage that only applies at some tiers

cdverrett94:
Closes #1081

Also fixes the chat cards from showing the 0 damage rolls and damage buttons.

Do you think we should add a method to the `AbilityModel` for generating a specific tiers text instead of having to map, filter, and join in multiple places? Or since its only in two spots, it's not worth it for now?


JPMeehan (approved):
Good work, nicely cleaned up Heartpiercer & Upchuck

PR #1107, opened 2025-09-17
Demons 3rd echelon monsters and malice features

AdellRedwinters:


AdellRedwinters:
When you review this, could you add a space in the monster Folder after the dash like "Demons - 3rd Echelon"? Didn't think it was worth sending a whole update for that.

PR #1112, opened 2025-09-17
Updated Targeting label handling

### LINKED ISSUE #886: Refactor ability targeting schema
The current targeting schema is insufficient
1. "Self" has become very complicated
  - Self
  - Self or Ally
  - Self and one or more allies
  - Self or creature
2. Weird specifications, such as size limits, inline with the targeting display

Having some way to combine multiple targets as well as have a custom label would be most helpful.

JPMeehan:
- Added a "Custom Label" property for ability targets, which will override the default label.
  - Also added "Enemy or Object" as a new target type.
  - Revised i18n strings for many of the "All" types, e.g. "All creatures" => "Each creature".

Closes #886

### DIFF
--- a/src/module/data/item/ability.mjs
+++ b/src/module/data/item/ability.mjs
@@ -63,6 +63,7 @@ export default class AbilityModel extends BaseItemModel {
     }, initial: "melee", required: true, blank: false });
     schema.target = new fields.SchemaField({
       type: new fields.StringField({ required: true, blank: false, initial: "self" }),
+      custom: new fields.StringField(),
       value: new fields.NumberField({ integer: true }),
     });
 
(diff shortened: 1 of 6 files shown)


src/module/data/item/ability.mjs
```diff
@@ -63,6 +63,7 @@ export default class AbilityModel extends BaseItemModel {
     }, initial: "melee", required: true, blank: false });
     schema.target = new fields.SchemaField({
       type: new fields.StringField({ required: true, blank: false, initial: "self" }),
+      custom: new fields.StringField(),
```
krbz999, line 66: Personal preference never to have string fields be `undefined` but if you feel strongly about it then eh.
```suggestion
      custom: new fields.StringField({ required: true }),
```

krbz999 (approved):

PR #1116, opened 2025-09-18
[#626] Refactor embed toggling

### LINKED ISSUE #626: Improve toggled description & embed handling
1. We should look if our shared DSDocumentSheetMixin can take up some of the work of fetching embedded documents, toggling descriptions, etc., with an eye towards methods that are reusable with Sudokus
2. We should switch from re-rendering the part to using CSS transitions & inserting the fresh html

cdverrett94:
Closes #626 

Moved the `_getEmbeddedDocument` function to the mixin. Kept it the same basically, with accounting for unowned AE's.

I think I've test different scenarios to make sure it all functions the way it did previously.

I noticed that the CSS for item-lists and effect-lists is largely similar. It might be good to add like `document-list` to the classes and then move most of the CSS to that.

In a future PR, we could probably also move the view/create/deleteDoc actions from the actor and item sheets to the mixin.

### DIFF
--- a/src/module/applications/api/document-sheet-mixin.mjs
+++ b/src/module/applications/api/document-sheet-mixin.mjs
@@ -1,7 +1,9 @@
 import constructHTMLButton from "../../utils/construct-html-button.mjs";
 
-/** @import { Constructor } from "@common/_types.mjs" */
-/** @import { Document } from "@common/abstract/_module.mjs" */
+/**
+ * @import { Constructor } from "@common/_types.mjs"
+ * @import { Document } from "@common/abstract/_module.mjs"
+ */
 
 const { HandlebarsApplicationMixin } = foundry.applications.api;
 
--- a/src/module/applications/api/document-sheet-mixin.mjs
+++ b/src/module/applications/api/document-sheet-mixin.mjs
@@ -1,6 +1,7 @@
 import constructHTMLButton from "../../utils/construct-html-button.mjs";
 
 /** @import { Constructor } from "@common/_types.mjs" */
+/** @import { Document } from "@common/abstract/_module.mjs" */
 
 const { HandlebarsApplicationMixin } = foundry.applications.api;
 
@@ -75,7 +76,7 @@ export default base => {
      * A set of the currently expanded document ids.
      * @type {Set<string>}
      */
-    _expandedDocumentEmbeds = new Set();
+    _expandedDocumentDescriptions = new Set();
 
     /* -------------------------------------------------- */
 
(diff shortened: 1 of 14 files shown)


src/module/applications/api/document-sheet-mixin.mjs
```diff
@@ -121,6 +130,67 @@ export default base => {
       return context;
     }
 
+    /* -------------------------------------------------- */
+    /*   Helper Functions                                 */
+    /* -------------------------------------------------- */
+
+    /* -------------------------------------------------- */
```
krbz999, line 137: Redundant divider

src/module/applications/api/document-sheet-mixin.mjs
```diff
@@ -121,6 +130,67 @@ export default base => {
...
+        let parent;
+        if (this.actor) {
+          parent = docRow.dataset.parentId === this.actor.id
+            ? this.actor
+            : this.actor.items.get(docRow?.dataset.parentId);
+        } else parent = this.item;
+
+        return parent.effects.get(docRow?.dataset.effectId);
+      } else return console.warn("Could not find document class");
+    }
```
krbz999, line 159: If moved to the document sheet mixin, it also needs to be generalized for any given document type. Currently this code assumes it's on an actor or item sheet.

src/module/applications/api/document-sheet-mixin.mjs
```diff
@@ -70,6 +71,14 @@ export default base => {
 
     /* -------------------------------------------------- */
 
+    /**
+     * A set of the currently expanded document ids.
+     * @type {Set<string>}
+     */
+    _expandedDocumentEmbeds = new Set();
```
krbz999, line 78: Prefer `_expandedDocumentDescriptions`.

krbz999 (changes requested):

src/module/applications/api/document-sheet-mixin.mjs
cdverrett94: Alright, I've got it generalized. Looked at how you had the `_getPseudoDocument` set up.

src/styles/system/applications/components/items.css
```diff
@@ -71,6 +71,20 @@
             }
           }
         }
+
+        .document-description {
```
JPMeehan, line 75: Needs padding on the left and right; text is going too far to the edge
image

JPMeehan (changes requested):
Good work on this refactor, just some small requests

JPMeehan (approved):

src/module/applications/api/document-sheet-mixin.mjs
```diff
@@ -121,6 +133,58 @@ export default base => {
...
+     * Fetches the embedded document representing the containing HTML element.
+     *
+     * @param {HTMLElement} target    The element subject to search.
+     * @returns {Document} The embedded document
+     */
+    _getEmbeddedDocument(target) {
+      const parentUuid = target.closest("[data-parent-uuid]").dataset.parentUuid;
+      const documentClass = target.closest("[data-document-class]").dataset.documentClass;
+      const documentId = target.closest("[data-document-id").dataset.documentId;
+      const parentDocument = fromUuidSync(parentUuid);
```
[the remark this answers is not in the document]
cdverrett94: Do you have an example of it throwing an error in a compendium? I tested actors and items in compendiums and neither was throwing an error.

As for using `documentName` over `documentClass`. Do you just want that change in this method or are you wanting me to change the html right now too? I ask because the html is on more than just the description toggling stuff.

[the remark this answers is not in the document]
cdverrett94: Are there other instances where this could be an issue or are there other document types that can be this deeply embedded `grandparent > parent > document`? Are there deeper instance? Looking at other document collections, I don't think so, but am not 100% sure.

I think I'd prefer a follow up PR just to minimize chance for breaking the stuff in this one.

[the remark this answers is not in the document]
krbz999: Two levels is fine for now.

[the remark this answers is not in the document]
JPMeehan: Yeah, I think doing cleanup in a follow-up PR is fine

PR #1125, opened 2025-09-18
[#1054] Add API for damage/heal modification

cdverrett94:
Reopening from #1057 without further changes since it auto-closed when the 0.9.0 branch was deleted.

Related to #1054 

Marking as draft since a discussion on keyboard modifiers or context menu needed to be had still. I had added a screenshot of my idea for a solution as well.

Several abilities are listed in the issue that would benefit from this.

### DIFF
--- a/src/module/data/message/ability-use.mjs
+++ b/src/module/data/message/ability-use.mjs
@@ -169,7 +169,7 @@ export default class AbilityUseModel extends BaseMessageModel {
 
     if (additionalTerms) {
       const terms = DamageRoll.parse(additionalTerms, rollData);
-      for (const term of terms) await term.evaluate(evaluationOptions);
+      for (const term of terms) if (!term._evaluated) await term.evaluate(evaluationOptions);
       newRoll.terms = newRoll.terms.concat(new foundry.dice.terms.OperatorTerm({ operator: "+" }), terms);
       newRoll.resetFormula();
       newRoll._total = newRoll._evaluateTotal();


cdverrett94:
@JPMeehan @krbz999 

Based on the [discussion in discord](https://discordapp.com/channels/732325252788387980/1345535794554601553/1420827164605026355) earlier:
- Removed the UX piece (alt click) and it's just API only for now so we can start small
- Methods moved to the base message model. 
- Expanded the dialog to allow selecting the # of damage surges.

I've got some TODOs commented, but wanted to verify this is what you were thinking. The initial request mentioned updating the chat message and rolls, but the damage roll can't be updated since some of the modifications (surges, tide of death) would be on an individual basis. Instead of creating a new message with the modified roll, I could add the new roll to the rolls array?

src/module/data/message/ability-use.mjs
```diff
@@ -147,4 +148,45 @@ export default class AbilityUseModel extends BaseMessageModel {
...
+    return newRoll;
+  }
+
+  /**
+   * Spend the speaker actor's surges.
+   * @param {number} surges The new resource value.
+   * @returns {DrawSteelActor}
+   */
+  async spendSurges(surges) {
+    if (this.parent.speakerActor?.type !== "hero") return this.parent.speakerActor;
```
JPMeehan: My gut is we should be throwing an error? I agree this should be an ability method because of the bit about retainers spending their mentor's surges

cdverrett94:
> Apologies for the review delay. This API makes sense - construct a new roll and append it to the message as a new option.

All good! I'll make the changes in the next day or 2. Finishing some work travel right now.

src/module/data/message/ability-use.mjs
JPMeehan: Yeah I think the surge spending should be on the ability model rather than the ability use message model in case we want to provide the option to spend the surges up front for say the potency adjustment

cdverrett94: In hindsight, this PR isn't really meant to be about applying surges necessarily, right? There's another issue for that. I bring this up because with moving it to the ability model, I think there's more that needs to go into the method, such as the surge use (damage or potency). Should I just strip the spendSurge stuff and that can be implemented once the complete surge use stuff gets done?

cdverrett94: Okay, I'll remove that and we can focus on the damage roll modification stuff for this PR.

src/module/data/message/ability-use.mjs
```diff
@@ -147,4 +148,31 @@ export default class AbilityUseModel extends BaseMessageModel {
...
+  /**
+   * Create a new DamageRoll based on applying modifications to a given DamageRoll.
+   * After creation, the new roll is added to the message rolls.
+   * @param {DamageRoll} roll The damage roll to modify.
+   * @param {object} modifications The modification options to apply.
+   * @param {string} [modifications.additionalTerms] Additional formula components to append to the roll.
+   * @param {string} [modifications.damageType] The damage type to use for the modified roll.
+   * @returns {DamageRoll}
+   */
+  async createModifiedDamageRoll(roll, { additionalTerms = "", damageType }) {
```
JPMeehan: Add a third parameter for `evaluationOptions`, specifically we might want to pass through `allowInteractive: false`

src/module/data/message/ability-use.mjs
```diff
@@ -147,4 +148,31 @@ export default class AbilityUseModel extends BaseMessageModel {
...
+    const rollData = ability?.getRollData() ?? this.parent.getRollData();
+
+    const formula = additionalTerms ? `${roll.total} + ${additionalTerms}` : String(roll.total);
+    const options = { ...roll.options };
+    if (damageType) options.type = damageType;
+
+    const newRoll = new DamageRoll(formula, rollData, options);
+    await newRoll.evaluate();
+
+    this.parent.update({ rolls: this.parent.rolls.concat(newRoll) });
```
JPMeehan:
```suggestion
    await this.parent.update({ rolls: this.parent.rolls.concat(newRoll) });
```

JPMeehan (commented):
Small adjustments

src/module/data/message/ability-use.mjs
```diff
@@ -147,4 +148,31 @@ export default class AbilityUseModel extends BaseMessageModel {
...
+   * @param {object} modifications The modification options to apply.
+   * @param {string} [modifications.additionalTerms] Additional formula components to append to the roll.
+   * @param {string} [modifications.damageType] The damage type to use for the modified roll.
+   * @returns {DamageRoll}
+   */
+  async createModifiedDamageRoll(roll, { additionalTerms = "", damageType }) {
+    const ability = await fromUuid(this.uuid);
+    const rollData = ability?.getRollData() ?? this.parent.getRollData();
+
+    const formula = additionalTerms ? `${roll.total} + ${additionalTerms}` : String(roll.total);
```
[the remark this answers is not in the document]
cdverrett94: I can do that, but I just used the result instead of the total from the original roll.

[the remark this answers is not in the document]
krbz999: I think seeing the dice would be nice.

PR #1126, opened 2025-09-18
War Dogs 2nd Echelon Monsters and ALL Echelon Malice Features

AdellRedwinters:
This includes Malice Features from 3rd and 4th Echelon


JPMeehan (approved):
Continuing to just normalize sort values instead of trying to get overly clever.

PR #1129, opened 2025-09-19
Consumable and Trinket Keyword update

AdellRedwinters:
quick sweep through our currently made consumables and trinkets to add missing keywords.


JPMeehan (changes requested):
Can you also remove the "bonus" description text in each of these items that Nagi added, now that we just have the keywords available?

JPMeehan (approved):

Issue #557, opened 2025-06-02
Implement "Update from Compendium" dialog


JPMeehan:
As a user, I would like to be able to cleanly update my in-world documents to match their compendium versions to account for package updates.

Implementation notes: should be accessible via a standard header button, no need to do the fancy frame adjustment

krbz999:
For items (at least), this would have an implementation specific to each item type. Swapping your "old" class item for a new one would be quite different than swapping out a trait.

PR #1132, opened 2025-09-19
[#557] Update document from Compendium

JPMeehan:
Implementation notes
- Our advancements store the choices in flags, so they won't get wiped by hard replacing `system`
- In general not touching flags is probably wise to avoid messing with module properties or even core features like alternative sheets.
- I don't see a strong reason to replace name & img properties, but could add as options in a confirmation dialog
- Could also add other configuration options for more or less properties.

This PR covers an API-only implementation. Future PRs can implement UI access.


krbz999:
An option that should be seldom used shouldnt be this accessible. I'm wondering if it should just be API only.

src/module/documents/actor.mjs
```diff
@@ -52,6 +52,80 @@ export default class DrawSteelActor extends BaseDocumentMixin(foundry.documents.
 
   /* -------------------------------------------------- */
 
+  /**
+   * Update this item based on its compendium version.
+   * @param {object}  [config]              Options to modify the update logic.
+   * @param {string}  [config.uuid]         An optional reference for a UUID to use in place of the stored compendiumSource.
+   * @param {boolean} [config.skipDialog]   Whether to skip the confirmation dialog.
+   */
+  async updateFromCompendium({ uuid, skipDialog } = {}) {
```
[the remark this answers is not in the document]
krbz999: It came to me in a dream that we could simply have this on a wiki page, as a footnote under Advancements.

src/module/documents/actor.mjs
```diff
@@ -52,6 +52,80 @@ export default class DrawSteelActor extends BaseDocumentMixin(foundry.documents.
...
+
+      if (localItem) {
+        await localItem.updateFromCompendium({ uuid: item.uuid, skipDialog: true });
+      }
+      else {
+        itemCreateData.push(game.items.fromCompendium(item));
+      }
+    }
+
+    await this.createEmbeddedDocuments("Item", itemCreateData, { keepId: true });
```
krbz999, line 124: Should there be some sort of confirmation ui.notif?

src/module/documents/item.mjs
```diff
@@ -71,6 +75,63 @@ export default class DrawSteelItem extends BaseDocumentMixin(foundry.documents.I
...
+        changes: effectData.changes,
+        description: effectData.description,
+      });
+      else createData.push(effectData);
+    }
+    for (const effect of this.effects) if (!sourceItem.effects.get(effect.id)) deleteIds.push(effect.id);
+
+    await this.updateEmbeddedDocuments("ActiveEffect", updateData);
+    await this.createEmbeddedDocuments("ActiveEffect", createData, { keepId: true });
+    await this.deleteEmbeddedDocuments("ActiveEffect", deleteIds);
```
krbz999, line 132: Good amount of overlap in the code suggests it would indeed work well to move into `ds.utils.migration`.

PR #1133, opened 2025-09-20
[#1119] Use token name in PR dialog and add target token hovering

### LINKED ISSUE #1119: Make power roll pop up use token name instead of actor name
### User Experience

When targeting multiple monsters it is hard to distinguish them in power roll pop up for edges and banes to apply edge/bane to individual monster. As shown in the image even though the goblins have different names using the auto numbering they all are just goblin warrior in the pop up thus it's hard to distinguish them. Using token name instead of actor name could help with that. 
Image

cdverrett94:
Closes #1119 

The issue only requests to use token name instead of actor name to more easily separate the targets. I figured a good improvement would be token hovering similar to the combat tracker, since a player may not always know which name goes with which token.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 - New Compendium Content:
   - Demon Echelons 2-4
 - Added suggested books and licenses in the source input form. (#841)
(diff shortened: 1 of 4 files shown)


JPMeehan (approved):
Works well, no noticeable performance issues with the mouse event.

PR #1135, opened 2025-09-20
Ancestry immunities

mrollins:


JPMeehan:
I believe the GM description for Wyrmplate also needs a fix? I may just do that when I add in the Changelog entry

mrollins:
I updated the description for the Dragon Knight features. I also made them default to active.

Issue #1114, opened 2025-09-18
Bredbeddler's `Solo Action` and other malice features have no Malice cost


KJTailor:
### What happened?

Bredbeddler's `Solo Action` and other malice features have no Malice cost

### Reproduction Steps

Draw Steel 0.8.1

### Bug Checklist

- [ ] The issue occurs while all Modules are disabled
- [ ] I am using The Forge for hosting

mrollins:
I'll take this

PR #1137, opened 2025-09-20
Add missing cost to Bredbeddle Malice abilities. (#1114)

mrollins:
Phantom bredbeddle's current stamina matches max stamina.


JPMeehan (approved):
Good work, just had to add the changelog entry

PR #1148, opened 2025-09-23
Followup HTML Cleanup

cdverrett94:
Follow up to #1116 

As part of removing some of the data attributes from the HTML, some of the JS had to change too.


cdverrett94:
Yeah thats what prompted me to open #1149. 

Do you want me to fix that in this PR or as a separate? Doing a separate one was my plan after this got merged.

JPMeehan:
Ahhh yeah in that case follow up is fine. Can also address the SortingHelpers deprecation I found

JPMeehan (approved):

PR #1154, opened 2025-09-23
[#1147] Add move action type and default move actions

### LINKED ISSUE #530: Consider implementing `Movement Actions`
While strictly not necessary, it might be nice to have the additional ability type `Movement Action`, probably at the very bottom of the abilities tab

### LINKED ISSUE #1147: Add a Move action for hero abilities
### User Experience

Image

cdverrett94:
Closes #1147 
Closes #530 

Added the move action type to the config. Since we're adding this action type, I went ahead and added default core moves (advance, disengage, and ride) as new basic abilities to be auto populated on heroes.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -813,6 +813,7 @@
           "FreeManeuver": "Free Maneuver",
           "Triggered": "Triggered Action",
           "FreeTriggered": "Free Triggered Action",
+          "Move": "Move Action",
           "None": "No Action",
           "Villain": "Villain Action"
         },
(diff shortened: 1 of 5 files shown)


lang/en.json
```diff
@@ -813,6 +813,7 @@
           "FreeManeuver": "Free Maneuver",
           "Triggered": "Triggered Action",
           "FreeTriggered": "Free Triggered Action",
+          "Move": "Move Action",
```
JPMeehan, line 816: Looks like it doesn't get the "action" label in the book
```suggestion
          "Move": "Move",
```

PR #1156, opened 2025-09-24
[#1149] Fix Active Effect sorting and address performIntegerSort deprecation warnings

### LINKED ISSUE #1149: Active Effects are not sorted in the effects list
### What happened?

We handle the sorting on drop, but the list doesn't actually reflect that sort. That part would be an easy fix.

If you try to sort AEs granted from items, then you get this error. I think this would require overriding `_onDraftStart`.

Image

### Reproduction Steps

Try sorting AEs on the actor sheet

_No response_

### Bug Checklist

### LINKED ISSUE #1155: Address SortingHelpers deprecation
Trying to sort an AE on the NPC sheet triggers a deprecation

Image

cdverrett94:
Closes #1149 
Closes #1155 

Copied the core `_onDragStart` method and just replaced the part for getting drag data.


JPMeehan:
I just want to see the code, which we will within the next 7 days. We may diverge or we may just need to make a small adjustment

cdverrett94:
Oh I didn't realize P1 was that close. I knew they were close, but didn't realize it'd be that close. Yeah, waiting makes sense then!

JPMeehan:
The milestone is closed up, just a question of tomorrow or early next week

JPMeehan:
Looks like core is expecting a conditional itemId property. My gut is that given that we're using parentId, we should just commit to being different.

cdverrett94:
I can replace parentId with itemId on grandchild AEs since it's basically the same thing. Then we can just remove the override later. Not at my computer, but I don't think parentId is being used anymore.

cdverrett94:
Ah yeah fair. I can update this later to check for documentUuid instead of itemId and effectId and then remove those data attributes from the HTML.

krbz999:
> Ah yeah fair. I can update this later to check for documentUuid instead of itemId and effectId and then remove those data attributes from the HTML.

That would be good.

cdverrett94:
Idk what happened when i rebased. i'll just reopen.

PR #1178, opened 2025-09-26
[#1173] Corrected roles for Grulqin, Orliq, and Wobalas

### LINKED ISSUE #1173: Incorrect roles for Grulqin, Orliq, and Wobalas
### What happened?

The listed role for the Grulqin is "Minion", when it should be "Brute" as per the Monsters book
The listed role for the Orliq is "Minion", when it should be "Harrier" as per the Monsters book
The listed role for the Wobalasis "Minion", when it should be "Artillery" as per the Monsters book

### Reproduction Steps

Open one of these demons in the monsters compendium

Draw Steel 0.8.1

### Bug Checklist

cdverrett94:
Closes #1173 

They had invalid roles of "minion". Added their correct roles. 

Since we had this issue with monsters having the solo role too, I did a search in the packs folder for invalid roles. Found two more that had a "leader" role which wasn't correct. Updated those as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,7 @@
     - Corrected Hrraaaaaagh! ability to be a free triggered action. (#1145)
     - Added 1 malice cost to the Hrraaaaaagh! ability per errata.
     - Corrected the description of the Destructive Path feature per errata.
+  - Corrected monster roles to the Grulqin, Orliq, and Wobalas. (#1173)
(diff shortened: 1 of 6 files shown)


JPMeehan (changes requested):
Can you also fix the img & prototype token properties?

JPMeehan (approved):
Thanks for doing this

PR #1184, opened 2025-09-27
Elementalist Fixes

AdellRedwinters:
Fixed some errors for the Elementalist abilities we've made and added custom targeting info for abilities that needed it.


JPMeehan (approved):
Good work, just went in and made changelog updates

PR #1197, opened 2025-10-02
Implement static references in PseudoDocument collections

### LINKED ISSUE #1171: Improve infrastructure of PseudoDocuments

krbz999:
Closes #1171.

This was fun, and it led to being able to remove a lot of code I wasn't entirely happy with.

The basic premise is thus:
- The collection of pseudo-documents is assigned to the root of the document and is never re-assigned.
- The collection handles altering the source data, which it holds, when a pseudo-document is added, updated, or removed entirely.
- The collection handles storing the models as usual, but will never *replace* the entries when they are updated; instead when updating a PseudoDocument it is re-initialized with the new source data. This results in the reference to the pseudo being static (just like regular Documents and embedded documents).

In other words,
```js
const item = fromUuidSync("Item.VplOyY21TDNkWsYH");
const advancement = item.pseudoCollections.Advancement.get("1XgNokBXzQVdKWN8");
await advancement.update({ "name": "NEW NAME!" }); // perform the update, which usually borked the reference
item.pseudoCollections.Advancement.get("1XgNokBXzQVdKWN8") === advancement; // TRUE!
```

There is some follow-up work to do for pseudo-documents that don't have subtypes, but we are yet to use these.

There is also some follow-up work to do for pseudo-documents that hold collections of pseudo-documents, but we are also yet to use these.

### DIFF
--- a/src/module/applications/api/pseudo-document-sheet.mjs
+++ b/src/module/applications/api/pseudo-document-sheet.mjs
@@ -17,8 +17,7 @@ export default class PseudoDocumentSheet extends HandlebarsApplicationMixin(Appl
    */
   constructor(options) {
     super(options);
-    this.#pseudoUuid = options.document.uuid;
-    this.#document = options.document.document;
+    this.#pseudoDocument = options.document;
   }
 
   /* -------------------------------------------------- */
@@ -65,7 +64,7 @@ export default class PseudoDocumentSheet extends HandlebarsApplicationMixin(Appl
 
   /**
    * Registered sheets. A map of documents to a map of pseudo-document uuids and their sheets.
-   * @type {Map<foundry.abstract.Document, Map<string, PseudoDocumentSheet>>}
+   * @type {Map<Document, Map<string, PseudoDocumentSheet>>}
    */
   static #sheets = new Map();
 
@@ -98,10 +97,10 @@ export default class PseudoDocumentSheet extends HandlebarsApplicationMixin(Appl
   /* -------------------------------------------------- */
 
   /**
-   * Stored uuid of this pseudo document.
-   * @type {string}
+   * The pseudo document.
+   * @type {PseudoDocument}
    */
-  #pseudoUuid;
+  #pseudoDocument;
 
   /* -------------------------------------------------- */
 
@@ -110,15 +109,7 @@ export default class PseudoDocumentSheet extends HandlebarsApplicationMixin(Appl
    * @type {TPseudo|null}
    */
   get pseudoDocument() {
-    let relative = this.document;
-    const uuidParts = this.#pseudoUuid.replace(relative.uuid, "").slice(1).split(".");
-    for (let i = 0; i < uuidParts.length; i += 2) {
-      const dname = uuidParts[i];
-      const id = uuidParts[i + 1];
-      relative = relative?.getEmbeddedDocument(dname, id);
-      if (!relative) return null;
-    }
-    return relative;
+    return this.#pseudoDocument;
   }
 
   /* -------------------------------------------------- */
@@ -127,16 +118,8 @@ export default class PseudoDocumentSheet extends HandlebarsApplicationMixin(Appl
    * The parent document.
    * @type {Document}
    */
-  #document;
-
-  /* -------------------------------------------------- */
-
-  /**
-   * The parent document.
-   * @type {Document}
-   */
   get document() {
-    return this.#document;
+    return this.#pseudoDocument.document;
   }
 
   /* -------------------------------------------------- */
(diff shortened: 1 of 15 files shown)


src/module/applications/api/pseudo-document-sheet.mjs
```diff
@@ -110,33 +109,17 @@ export default class PseudoDocumentSheet extends HandlebarsApplicationMixin(Appl
...
-    let relative = this.document;
-    const uuidParts = this.#pseudoUuid.replace(relative.uuid, "").slice(1).split(".");
-    for (let i = 0; i < uuidParts.length; i += 2) {
-      const dname = uuidParts[i];
-      const id = uuidParts[i + 1];
-      relative = relative?.getEmbeddedDocument(dname, id);
-      if (!relative) return null;
-    }
-    return relative;
+    return this.#pseudoDocument;
```
krbz999, line 112: Huzzah, less jank!

src/module/applications/apps/power-roll-dialog.mjs
```diff
@@ -95,7 +95,7 @@ export default class PowerRollDialog extends RollDialog {
     context.damageOptions = null;
     for (const tier of PowerRoll.TIER_NAMES) {
 
-      const effect = context.ability.system.power.effects.getByType("damage").find(e => e.damage[tier].types.size > 1);
+      const effect = context.ability.system.power.effects.documentsByType.damage.find(e => e.damage[tier].types.size > 1);
```
krbz999, line 98: The `getByType` method, initially copied from dnd5e, has been yote as well, in favor of the now general `Collection#documentsByType` getter which is found on all EmbeddedCollections throughout the platform now.

No deprecation or shim has been put in place for `getByType`.

src/module/data/pseudo-documents/pseudo-document.mjs
```diff
@@ -100,13 +100,8 @@ export default class PseudoDocument extends foundry.abstract.DataModel {
...
   get fieldPath() {
-    const fp = this.schema.fieldPath;
-    let path = fp.slice(0, fp.lastIndexOf("element") - 1);
-
-    if (this.parent instanceof PseudoDocument) {
-      path = [this.parent.fieldPath, this.parent.id, path].join(".");
-    }
-
+    let path = this.parent.constructor.metadata.embedded[this.documentName];
+    if (this.parent instanceof PseudoDocument) path = [this.parent.fieldPath, this.parent.id, path].join(".");
```
krbz999, line 104: I found a bug here though I did not test whether it existed on the main branches. It prevented updates, but regardless the new edition is simpler.

src/module/utils/model-collection.mjs
```diff
@@ -1,29 +1,58 @@
...
  */
 export default class ModelCollection extends foundry.utils.Collection {
+  constructor(documentName, document, data) {
+    super();
+    const name = CONFIG[document.documentName].dataModels[document._source.type].metadata.embedded[documentName];
+    Object.defineProperties(this, {
+      name: { value: name, writable: false },
+      _source: { value: data, writable: false },
+      documentClass: { value: ModelCollection.documentClasses[documentName], writable: false },
+    });
```
krbz999, line 24: If you compare the `ActiveEffect` collection to the `Advancement` collection on eg a class item, you'll notice only few differences, main being the `name` being long-winded for the pseudo-collections (see my comment on the PR).

In dnd5e, they assign the collection name differently, eg `advancements` and `activities`. We don't have a collection name for `PowerRollEffect`s.

src/module/utils/model-collection.mjs
```diff
@@ -38,50 +67,86 @@ export default class ModelCollection extends foundry.utils.Collection {
...
   /**
-   * Fetch an array of data models of a certain type.
-   * @param {string} type     The subtype of the data models.
-   * @returns {Model[]}   The data models of this type.
+   * The subtypes sorted by type.
+   * @type {Record<string, Model[]>}
    */
-  getByType(type) {
-    return Array.from(this.#types.get(type) ?? []).map(key => this.get(key));
+  get documentsByType() {
```
krbz999, line 83: The previous iteration updated `#types` with each addition or removal. Like with embedded collections, this getter now instead handles the assignment of `#types` (now `#documentsByType`) and other methods are responsible for `null`'ing it so it can be updated.

src/module/utils/model-collection.mjs
```diff
@@ -38,50 +67,86 @@ export default class ModelCollection extends foundry.utils.Collection {
...
+    if (this.#documentsByType) return this.#documentsByType;
+    const types = Object.fromEntries(Object.keys(this.documentClass.documentConfig).map(t => [t, []]));
+    for (const doc of this.values()) types[doc._source.type ?? "base"]?.push(doc);
```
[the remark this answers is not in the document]
krbz999: No, cus `groupBy` won't create empty arrays - we would then no longer be able to assume that, say, `item.system.advancements.documentsByType.trait` exists.

PR #1215, opened 2025-10-10
Audited implementation of StringField

JPMeehan:
- Addressed inability to reset the "Additional" field on Item Grants to blank
- General consistency pass

### DIFF
--- a/src/module/data/message/ability-use.mjs
+++ b/src/module/data/message/ability-use.mjs
@@ -28,7 +28,7 @@ export default class AbilityUseModel extends BaseMessageModel {
   static defineSchema() {
     const schema = super.defineSchema();
     // All ability use messages MUST have a uuid pointing to the relevant document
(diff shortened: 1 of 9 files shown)


src/module/data/message/ability-use.mjs
```diff
@@ -28,7 +28,7 @@ export default class AbilityUseModel extends BaseMessageModel {
   static defineSchema() {
     const schema = super.defineSchema();
     // All ability use messages MUST have a uuid pointing to the relevant document
-    schema.uuid = new fields.StringField({ required: true, nullable: false, blank: false });
+    schema.uuid = new fields.DocumentUUIDField({ nullable: false, type: "Item" });
```
krbz999: do we need a migration here?

JPMeehan: Should not, we've only been feeding it UUIDs. I honestly couldn't figure out why it wasn't one to begin with

PR #1236, opened 2025-10-17
[#907] Implemented PseudoDocument Drag and Drop

### LINKED ISSUE #907: Add Drag and Drop support for Pseudodocuments
As a user, I would like to be able to drag and drop Power Roll Effects and Advancements to
1. Change the Order
2. Copy from one item to another

JPMeehan:
Need to follow up with handling "Advancements dropped on an actor where the advancement should be run"

Closes #907

### DIFF
--- a/src/module/data/pseudo-documents/pseudo-document.mjs
+++ b/src/module/data/pseudo-documents/pseudo-document.mjs
@@ -239,26 +239,18 @@ export default class PseudoDocument extends foundry.abstract.DataModel {
 
   /**
    * A helper function to handle obtaining the relevant PseudoDocument from dropped data provided via a DataTransfer event.
-   * The dropped data could have:
-   * 1. A data object explicitly provided
-   * 2. A UUID.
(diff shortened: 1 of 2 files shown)


src/module/data/pseudo-documents/pseudo-document.mjs
```diff
@@ -233,6 +235,38 @@ export default class PseudoDocument extends foundry.abstract.DataModel {
...
+   *
+   * @param {object} data           The data object extracted from a DataTransfer event.
+   * @returns {Promise<PseudoDocument>}   The resolved PseudoDocument.
+   * @throws If a Document could not be retrieved from the provided data.
+   */
+  static async fromDropData(data) {
+    let pseudo = null;
+
+    // Case 1 - Data explicitly provided
+    if (data.data) pseudo = new this(data.data);
```
krbz999: I understand we align with core here, but it doesn't seem necessary. There will always be a `uuid`.

PR #1238, opened 2025-10-18
Use item image when creating craft treasure project

cdverrett94:
Previously, it would use the default item icon. Now it'll use the item's img for the project.
Screenshot 2025-10-18 075526


cdverrett94:
I always forget about the changelog!

PR #1240, opened 2025-10-18
Conduit levels 6 to 8

mrollins:


JPMeehan (changes requested):
As with the previous 4 & 5 we may need to restructure the actual advancements part in a future PR due to the double-subclass oddity.

JPMeehan (approved):

PR #1244, opened 2025-10-19
[#823] Implemented Vertical Ellipses

### LINKED ISSUE #823: Consider switching right-click context menus for vertical ellipsis
Right-click context menus are less discoverable than vertical ellipsis buttons. While they do have a bigger hitbox, it would also allow better distinction between Item and AE operations on the actor sheets.

### LINKED ISSUE #928: Improve handling of round-duration effects
Many personal buffs have standard foundry timed durations such as a round or three. It would be helpful if toggling them as "enabled" refreshed their start time to the current one.

JPMeehan:
Overhauled actor & item sheet context menus. Major improvements for effects especially.

image

Closes #823

Closes #928

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,7 @@
   - This is a change for applied power roll effects that also applies to the new Apply enricher.
 - In the Power Roll Dialog, changed the target's name from the actor name to the token name. (#1119)
 - Heroes now have a helper app for adjusting characteristics. (#1206)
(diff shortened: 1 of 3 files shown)


JPMeehan:
The right click is still around and has a hit area for the full row, and just clicking on the row for stuff like items is tied to using them.

JPMeehan:
Oh I misread yeah ++ height and width is probably justified

PR #1245, opened 2025-10-19
Improved docs asset usage

JPMeehan:
Currently just using the /develop branch as our reference point, open to considering if it should be main instead.


JPMeehan:
Which cycles back to I think the current implementation works; /develop is a sufficient source of truth for image assets, and this will correctly parse the necessary URL paths for each.

PR #1246, opened 2025-10-19
Monster Updates 2

AdellRedwinters:
Addressed some errors and added Apply Effects for Devils, Draconians, Dragons, and Dwarves.

Deathcount is another feature that would benefit from allowing some sort of counter tracking for certain abilities, similar to Rage on werewolves. Stacking works for now but having 10 of the same icon on a token does get pretty busy and hard to navigate when the number is meant to go up and down constantly throughout the battle.

Meteor Dragon: I opted out of making an active effect for Annihilated on the Meteor Dragon as it is isolated to just that villain action and is resolved within the villain action, it has no impact on the battle before or after so tracking it on the pc seems like a waste of time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@
   - Demon Echelons 2–4
   - Devils
   - Draconians
+  - Dragons
   - Elves (Shadow)
   - Giants
   - Hobgoblins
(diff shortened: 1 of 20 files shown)


JPMeehan (approved):
Looks good, did my usual touch-ups to ensure sources & prototype token information was to our expectations

PR #1247, opened 2025-10-19
Small cleanup

JPMeehan:
- Ensured usage of glyphs on the ability sheet
- Typing fixes
- Added content link highlighting to item sheet

### DIFF
--- a/src/styles/system/applications/applications.css
+++ b/src/styles/system/applications/applications.css
@@ -28,4 +28,7 @@ dl.power-roll-display {
     font-weight: 500;
     line-height: 0.9;
   }
+  dd > p, dt > p {
+    margin: 0;
+  }
 }


JPMeehan:
applications.css appears to somehow be old enough it's in CRLF which is why it looks like a total replacement

src/styles/system/applications/applications.css
```diff
@@ -1,19 +1,31 @@
...
+.application.draw-steel .section-header {
+  display: flex;
+  justify-content: space-between;
+}
+
+/**
+ * General use CSS
+ **/
+dl.power-roll-display {
+  dt.tier1, dt.tier2, dt.tier3 {
```
krbz999: Do we need the distinction between tiers?

JPMeehan: Not really that was just me trying to be forward looking with the classes

PR #1248, opened 2025-10-19
Troubadour Features and Abilities for 4 and 5

alinuxfan:
Added Troubadour level 4 and 5 features and abilities


JPMeehan (approved):
Good clean work. Hear ye Hear Ye probably needs a styling revisit to use the Draw Steel glyphs but that's a future problem after we do some CSS work to figure out those styles.

PR #1249, opened 2025-10-19
[#744] Disable interaction with rollable elements on actor sheet for non-owners

metrzero:
Disabled rollable elements in:
* Characteristics
* Projects
* Abilities

Updated Free Strike on NPC sheets to display the attack value without a button.

Tested all changes and everything still works for owners as expected and disabled for non-owners.


metrzero:
Looks like I wasn't updated before opening the PR. Feel free to reject if you'd like me to re-submit.

PR #1250, opened 2025-10-20
Shadow level 8, 9 and 10

Stolassz:
Added:
Level 8: College Features and class abilities  (11 insight) Level 9: College Abillities (11 insight) and class features Level 10: Features

Change of the level 7 class feature “Careful Observation Improvement” to “Careful Observation Improvement I,” so as not to conflict with the level 9 feature of the same name, here named “Careful Observation Improvement II.”


JPMeehan (approved):
Good clean work

PR #1265, opened 2025-10-21
[#26] Batch token coloration

### LINKED ISSUE #26: Support color markers for tokens
When running multiple squads of minions, the Bestiary suggests differentiating them using colors, textures, or icons. This system should implement a few such markers as part of the built-in token Status Effects.

I am thinking of something similar to the [Token Color Marker](https://github.com/Gundancer/foundryvtt-token-color-marker) module. If this system is structured like the Simple Worldbuilding System, then maybe this package will work out-of-the-box.

JPMeehan:
An actual use of Dialog.wait! This PR implements a context menu option for combatant groups that will apply a tint or token ring color adjustment to all tokens in the combatant group.

image

image

Closes #26


JPMeehan:
Swatches make a lot of sense, not sure if the color picker natively supports that at all

JPMeehan:
I'll stick it in forms then?

PR #1268, opened 2025-10-22
[#704] Add condition immunities property to actors

metrzero:
Add condition immunities property to actors and use it to filter status effects when preparing derived data

### DIFF
--- a/src/module/documents/active-effect.mjs
+++ b/src/module/documents/active-effect.mjs
@@ -45,7 +45,7 @@ export default class DrawSteelActiveEffect extends foundry.documents.ActiveEffec
     if (immuneList.size) {
       // Warn the user with a list of condition names
       const formatter = game.i18n.getListFormatter({ type: "unit" });
-      const formattedConditions = formatter.format(immuneList.map(id => game.i18n.localize(ds.CONFIG.conditions[id]?.name)));
+      const formattedConditions = formatter.format(immuneList.map(id => game.i18n.localize(ds.CONFIG.conditions[id]?.name ?? id)));
 
       ui.notifications.warn("DRAW_STEEL.ActiveEffect.ImmunityWarning", { localize: true, format: { conditions: formattedConditions } });
     }


JPMeehan:
I don't know what you did in the commit history but I merged develop into the branch and despite being behind 14 commits it only made the most recent PR updates

image

JPMeehan (changes requested):
I think I see the problem

1. You can `return false` in _preCreate to block the creation of a document
2. `options.parent` will also have the parent information to check against

krbz999:
I think I will request we simplify the PR here for the original ask of a single property to hold the status ids which the actor is immune to.

I recognize that there are edge cases beyond just immunity, but that is able to be covered in a follow-up PR should it ever become relevant.

src/module/documents/active-effect.mjs
```diff
@@ -34,6 +34,82 @@ export default class DrawSteelActiveEffect extends foundry.documents.ActiveEffec
...
+  /* -------------------------------------------------- */
+
+  /** @inheritdoc */
+  _onUpdate(changed, options, userId) {
+    super._onUpdate(changed, options, userId);
+
+    if ((game.userId === userId) && (this.parent.documentName === "Actor")) {
+      // Undefined or false on disabled means we should perform the check
+      if (this.active && (changed.disabled != true)) {
+        this.#checkAndRemoveImmuneStatuses();
```
krbz999: Making data prep async is a huge no-no. This should never be done.

src/module/data/actor/base.mjs
```diff
@@ -64,6 +64,11 @@ export default class BaseActorModel extends DrawSteelSystemModel {
       weaknesses: damageTypes(requiredInteger, { all: true }),
     });
 
+    schema.conditions = new fields.SchemaField({
+      immunities: new fields.SetField(setOptions()),
+      ignored: new fields.SetField(setOptions()),
+    });
```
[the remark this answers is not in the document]
metrzero: I see two scenarios:
1. Standard condition immunity, which I believe should result in removing conditions to avoid confusion in the UI. Without this behavior, tokens show as "Slowed" but they register as full speed, or "Bleeding" which is still manually managed for triggering/rolling damage, will be assigned and appear on tokens that are listed as immune.
3. Temporary immunity, which should not remove existing conditions or prevent their assignment.

Some examples of that are:
Wellness Tonic
image

Absolute Zero ability
image

Only two examples, but they already exist. If you do still decide to simplify down to one, I think we should keep the condition prevention version as immunity, for the above reasons.

src/module/documents/active-effect.mjs
```diff
@@ -34,6 +34,21 @@ export default class DrawSteelActiveEffect extends foundry.documents.ActiveEffec
...
   /* -------------------------------------------------- */
 
+  /** @inheritdoc */
+  async _preCreate(data, options, user) {
+    super._preCreate(data, options, user);
+
+    // Check if actor has immunity to any statuses being applied by this effect
+    // If so, return false to prevent creation
+    for (const statusId of this.statuses) {
+      if (this.parent.system.conditions.immunities.has(statusId)) {
```
JPMeehan: 1. Assumes `this.parent` is defined (won't be true in v14)
2. Assumes `system` has `conditions` (isn't true if the parent is an item, isn't true if it's a custom data model subtype)

You should be searching for an actor parent, and then if it exists you can optional access for `system.condition.immunities`, then if *that* exists you can use that set.

src/module/documents/active-effect.mjs
```diff
@@ -34,6 +34,22 @@ export default class DrawSteelActiveEffect extends foundry.documents.ActiveEffec
 
   /* -------------------------------------------------- */
 
+  /** @inheritdoc */
+  async _preCreate(data, options, user) {
+    const allowed = await super._preCreate(data, options, user);
+    if (allowed === false) return false;
```
krbz999: Subjective preference to shorten this ever so slightly:
```suggestion
    if ((await super._preCreate(data, options, user)) === false) return false;
```

JPMeehan:
> Looks like the branch will also need a rebase, as some unrelated commits have snuck in.

We're going to squish anyways

JPMeehan (approved):
Yeah, this fits expectations

PR #1281, opened 2025-10-23
Shadow and Talent Advancements

Stolassz:
Inclusion of features and abilities in class and subclass advancements


JPMeehan (approved):
Made some adjustments to normalize the advancement names

PR #1290, opened 2025-10-25
Implemented item grant advancement requirement checks

JPMeehan:
This PR represents the logic to implement requirements in the Item Grant Configuration Dialog. The follow-up work would be to move the Conduit's domain advancements *out* of the domains and into the main class, with the chosen domains "unlocking" features and abilities to be chosen.

image

This PR adds a number of fields to existing item types, some of which should be filled out by batch updates. The advantage of this approach is those fields can serve other purposes; 
- abilities listing their linked classes can improve our Heroic Resource display for unowned items
- Expansion content on Patreon has shown Perks that have class requirements
- General CB requirements

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -780,11 +780,11 @@
             },
             "dsid": {
               "label": "DSID",
-              "hint": "The Draw Steel ID of other items (e.g. classes, subclasses) required by this item. Any may be fulfilled."
+              "hint": "The Draw Steel ID of other items (e.g. classes, subclasses) required by this item. Any may be fulfilled. Only needed if this is an option in advancements."
             },
             "level": {
               "label": "Level",
-              "hint": "The minimum level required to choose this item."
+              "hint": "The minimum level required to choose this item. Only needed if this an option in advancements."
             }
           }
         }


krbz999:
Too difficult to review without something to test with.

src/module/data/item/ability.mjs
```diff
@@ -45,6 +45,16 @@ export default class AbilityModel extends BaseItemModel {
     // Items don't have descriptions
     delete schema.description;
 
+    // Can be expanded over time for automation
+    schema.prerequisites = new fields.SchemaField({
```
krbz999: Like discussed, do we want a boolean property here?

krbz999:
Testing this now, here are the issues I am encountering:

Create a character and add enough XP to be able to raise them to level 2. Add the conduit class, pick the advancements, and for subclasses pick Creation and Death.

This works correctly and allows you to pick 1 (of several, though all but 2 are disabled) subclass features. Now level up to level 2; in the '2nd-Level Domain Feature' advancement, you are able to pick the same ability you chose at level 1.

This issue also happens at levels 4-5.

src/module/data/item/ability.mjs
JPMeehan: DSID
> The Draw Steel ID of other items (e.g. classes, subclasses) required by this item. Any may be fulfilled. Only needed if this is an option in advancements.

Level
> The minimum level required to choose this item. Only needed if this an option in advancements.

PR #1292, opened 2025-10-25
Censor Level 9 & 10. Features and Abilities. Also Conduit L9&10 Domain Abilities

alinuxfan:
Adds:
Censor Level 9 and 10 Class and Order Features and Abilities
Conduit Level 9 Domain Abilities and Level 10 Class Features. 
NOTE: Conduit Abilities are created but not yet linked to the Class Sheet.


JPMeehan (approved):
Only real adjustment was putting the holy damage enricher in the description of the Banished effect

PR #1303, opened 2025-10-27
AJAX

AdellRedwinters:


JPMeehan (approved):
Something I realized during this review is that the solos needed their `system.combat.turns` property set; Ajax being a special boy gets 3, but the rest of the solos in the game needed 2. I addressed that as part of this QA pass

PR #1306, opened 2025-10-28
[#714] Add unflankable condition

### LINKED ISSUE #714: Implement `unflankable` status
A non-hud status would be useful for active effects that need to make an actor immune to flanking. Also requires adjusting the getter on DrawSteelToken.

metrzero:
Closes #714

Also added logic to prevent dead actors from contributing to flanking.


metrzero:
So where did we land on approach? Keep it as-is in the schema, or move elsewhere and compute it during data prep? I'm not sure I'm clear on where the AE would target to indicate this property for the latter case.

JPMeehan:
I do think non-db only would be correct. Fair question on the polarity; `system.conditions.flankable = true` is probably the best way to handle this; anything missing that property is probably an object or something else not using the flanking logic. Then AEs would set to false with `system.conditions.flankable | OVERRIDE | false` or the like.

PR #1309, opened 2025-10-29
[#203] Add surprised condition

### LINKED ISSUE #203: Add Surprised as a condition
> A surprised creature can’t take triggered or free triggered actions and ability power rolls against them gain an edge.

metrzero:
Closes #203

- Added Surprised condition to rules journal
- Targeting a surprised actor grants an edge
- Added surprised as a condition

### DIFF
--- a/src/module/config.mjs
+++ b/src/module/config.mjs
@@ -167,14 +167,15 @@ preLocalize("healingTypes", { key: "label" });
 
 /**
  * @typedef DrawSteelCondition
- * @property {string} name
- * @property {string} img
- * @property {string} rule
- * @property {boolean} [sheet]
- * @property {boolean} [targeted]
- * @property {number} [maxSources]
- * @property {number} [defaultSpeed]
- * @property {Record<string, Set<string>>} [restrictions]
+ * @property {string} name            The i18n name of the condition. Localized as part of effect creation.
+ * @property {string} img             An SVG representing the condition.
+ * @property {string} rule            JournalEntryPage UUID Reference for the condition.
+ * @property {boolean} [sheet=true]   Show on the actor sheet? An explicit false is needed to not show.
+ * @property {boolean} [targeted]     Ask for a targeting prompt when this condition is applied.
+ * @property {number} [maxSources]    How many sources can be valid for this condition?
+ * @property {number} [defaultSpeed]  Used by slowed to infer the default speed to reduce to.
+ * @property {Record<string, Set<string>>} [restrictions] Restrictions on ability usage imposed by this condition.
+ *                                    The system currently supports `dsid` and `type` based restrictions.
  */
 
 /**


JPMeehan:
Trying to decide if I like the inclusion of surprised on the character sheet.

image

metrzero:
Agreed that (for PCs at least) there is unlikely enough use to warrant taking up this valuable real estate.

For my own games, I plan to handle this with a macro that assigns the status to selected tokens, but what's the plan for more general use?

krbz999:
If it is as important as the rest of the statuses displayed in the same area, it should go there.

JPMeehan:
So, Surprised... isn't, which is why it got missed until now. For example, it's not listed on the official character sheet
image


It's not listed with the others on p. 77; instead, it's listed on p. 266 as part of the initiative rules.

metrzero:
I'll plan to remove it from the actor sheet but keep it on the HUD.

metrzero:
> Also worth comparing the icon choice to dnd5e; there's a ton of options on https://game-icons.net/

Some more literal choices from game-icons:
https://game-icons.net/1x1/cathelineau/dread.html
https://game-icons.net/1x1/darkzaitzev/distraction.html
https://game-icons.net/1x1/lorc/surprised.html

Some examples generated using different fonts in the dnd5e style:
1. surp-1 surp-1-white
1. surp-2 surp-2
1. surp-3 surp-3

Thoughts?

assets/icons/surprised.svg
krbz999: Should be added to the license file as well.

src/module/config.mjs
```diff
@@ -173,6 +173,7 @@ preLocalize("healingTypes", { key: "label" });
  * @property {boolean} [targeted]
  * @property {number} [maxSources]
  * @property {number} [defaultSpeed]
+ * @property {boolean} [hud_only]
```
krbz999: We dont typically use underscores in properties, just `hud` or perhaps `sheet: false` would be better.

assets/icons/surprised.svg
metrzero: Ack. I generated this. I'm happy to release it to CC. Thoughts on where to indicate that?

src/module/applications/sheets/actor-sheet.mjs
```diff
@@ -443,8 +443,8 @@ export default class DrawSteelActorSheet extends DSDocumentSheet {
     /** @type {Record<string, StatusInfo>} */
     const statusInfo = {};
     for (const status of CONFIG.statusEffects) {
-      // Only display if it would show in the token HUD *and* it has an assigned _id
-      if ((!status._id) || !ActiveEffect.implementation.validHud(status, this.actor)) continue;
+      // Only display if it would show in the token HUD, is marked for sheet display, *and* it has an assigned _id
+      if ((!status._id) || (!status.sheet) || !ActiveEffect.implementation.validHud(status, this.actor)) continue;
```
krbz999: I would flip it in the config object, and here specifically check for `status.sheet === false`, but I do not feel strongly either way. The property missing can be equal to `false`.

metrzero: Great suggestion. Thanks.

JPMeehan (approved):
Looks good, I added in some type annotations to DrawSteelCondition because the typedef has gotten pretty complicated and it's good to know the intent before we rework whenever we do v14 prep

PR #1311, opened 2025-10-29
3rd Echelon Treasures

Stolassz:
Added:
3rd Echelon Consumables
3rd Echelon Trinkets


JPMeehan (approved):
Biggest adjustment was switching to our DL power roll effect instead of the table for Timesplitter

PR #1318, opened 2025-10-29
[#1270] Update PRE potencies to use DS Glyphs font

### LINKED ISSUE #1270: Leverage glyphs in PRE potencies
The {{potency}} substitute we use should leverage the glyphs in their formatting, including differentiating between abilities that are on an actor or not.

metrzero:
Closes #1270 

This version takes a text replacement approach with the idea that it keeps formatting separate from other data structure construction logic.

For reference, the current output:
image

### DIFF
--- a/src/module/data/pseudo-documents/power-roll-effects/base-power-roll-effect.mjs
+++ b/src/module/data/pseudo-documents/power-roll-effects/base-power-roll-effect.mjs
@@ -207,7 +207,7 @@ export default class BasePowerRollEffect extends TypedPseudoDocument {
     });
     // Since the resulting glyph string may contain HTML characters, we need to escape it properly
     const potencyGlyphs = Handlebars.escapeExpression(this.constructor.translatePotencyGlyphs(potencyString));
-    return `<div class="potency">${potencyGlyphs}</div>`;
+    return `<span class="potency">${potencyGlyphs}</span>`;
   }
 
   /* -------------------------------------------------- */
--- a/src/styles/system/applications/applications.css
+++ b/src/styles/system/applications/applications.css
@@ -28,8 +28,7 @@ dl.power-roll-display {
     font-weight: 500;
     line-height: 0.9;
   }
-  div.potency {
-    display: inline-block;
+  span.potency {
     font-family: "Draw Steel Glyphs";
     font-size: var(--font-size-16);
     text-transform: lowercase;


src/module/data/pseudo-documents/power-roll-effects/base-power-roll-effect.mjs
```diff
@@ -166,7 +207,9 @@ export default class BasePowerRollEffect extends TypedPseudoDocument {
       characteristic: ds.CONFIG.characteristics[tierValue.potency.characteristic]?.rollKey ?? "",
       value: potencyValue,
     });
-    return potencyString;
+    // Since the resulting glyph string may contain HTML characters, we need to escape it properly
+    const potencyGlyphs = Handlebars.escapeExpression(this.#translatePotencyGlyphs(potencyString));
```
JPMeehan: We *probably* now need to add these properties to HTML cleaning in `system.json` since it's now gonna get parsed as HTML on the client-side.

JPMeehan (commented):
Overall I like the approach and think it's relatively clean

src/module/data/pseudo-documents/power-roll-effects/base-power-roll-effect.mjs
```diff
@@ -152,12 +152,53 @@ export default class BasePowerRollEffect extends TypedPseudoDocument {
...
+        strength: "@potency.average",
+        glyph: "v",
+      },
+      {
+        strength: "@potency.strong",
+        glyph: "s",
+      },
+    ];
+
+    let glyphValue = closedValue;
```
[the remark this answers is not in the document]
metrzero: Funny, I originally had it that way and then felt it didn’t match the existing style. I’ll change it back.

src/module/data/pseudo-documents/power-roll-effects/base-power-roll-effect.mjs
JPMeehan: I'm getting HTML parsing here

image

image

metrzero:
Shoot! Thank you!

src/module/data/pseudo-documents/power-roll-effects/base-power-roll-effect.mjs
metrzero: I added escaping of the `display` string which would be a component of the string that eventually made it inside the `{{{ }}}` render. I did not add display to htmlFields as that seems to be for _intentional_ HTML fields, but if that's necessary it's an easy addition.

JPMeehan (approved):
I think I'm happy now with the implementation

Issue #1120, opened 2025-09-18
Disciple of Earth stamina bonus incorrect


zarick1342:
### What happened?

I created a level 2 dwarf (spark off your skin) earth elementalist who ended up with 24 max stamina, when they should have 42 (18 base + 6 level + 6 enchant of permanance + 6 earth elementalist + 6 spark off your skin).

This is a lot of random HP bonuses so I tried to track it down and it seems like the issue is that Disciple of Earth (the L2 elementalist feature) has `system.stamina.bonuses` add 6, when it seems like it should be a specific key of system.stamina.bonuses (I don't know which one since it's a one time bonus).

### Reproduction Steps

Create a level 2 earth elementalist who has any other special bonuses to HP, which will be overridden.

### What core and system version are you reporting this for?

FVTT 13.348, Draw Steel 0.8.1

### Bug Checklist

- [x] The issue occurs while all Modules are disabled
- [ ] I am using The Forge for hosting

PR #1330, opened 2025-11-01
[#1120] Fix Disciple of Earth Stamina Effect. system.stamina.max

alinuxfan:
#1120 Fix


JPMeehan (changes requested):
Neither of this is correct. This should just be `system.stamina.max | ADD | 6`; it's a steady +3 per level, but we don't have scaling for that so it's just going to be have to the value at level 2 when the elementalist gains the feature

PR #1332, opened 2025-11-01
Fix issues in `createModifiedDamageRoll`

cdverrett94:
It wasn't taking into account untyped damage or updating the flavor on damage type change.


JPMeehan (approved):
Good clean fix

PR #1334, opened 2025-11-03
[#1321] Implement forced movement bonuses

### LINKED ISSUE #1321: Allow Ability Modifiers to adjust the damage of forced movement
Bracers of Strife (Heroes p. 325)
> You gain ... +1 bonus to the distance you push any target with any weapon ability.

This should allow specifying push, pull, or slide separately. Unlike damage, there does not need to be a specification for a given tier result. An effect that increases all three can have multiple change lines, there does not need to be an "all forced movement" key.

metrzero:
Closes #1321 

I segmented out the logic to collect the forced movement bonuses, but otherwise kept everything as manipulating the `toText` output on the forced movement PRE given movement is not automated in any way.

### DIFF
--- a/src/docs/Active-Effects.md
+++ b/src/docs/Active-Effects.md
@@ -229,8 +229,12 @@ Ability modifiers support a far more limited set of keys, but still use the same
 |`forced.push`|Give a bonus to push movement distance|
 |`forced.pull`|Give a bonus to pull movement distance|
 |`forced.slide`|Give a bonus to slide movement distance|
+|`keyword`|Add a keyword to an ability|
 |`potency`|Give a bonus to all potency values|
 
-**Behavior:**
-- Type-specific bonuses (push/pull/slide) apply only to their matching movement types
-- Forced movement bonuses apply across all tiers; there is no way to target bonuses per tier
+**Notes:**
+- For `forced...`:
+  - Type-specific bonuses (push/pull/slide) apply only to their matching movement types
+  - Forced movement bonuses apply across all tiers; there is no way to target bonuses per tier
+- For `keyword`, use the "Add" change mode with values like `magic` or `strike`
+
--- a/src/module/data/item/ability.mjs
+++ b/src/module/data/item/ability.mjs
@@ -143,6 +143,15 @@ export default class AbilityModel extends BaseItemModel {
    * @protected
    */
   _applyAbilityBonuses() {
+    // Apply keyword modifiers first to ensure later effects operate on the modified set
+    for (const bonus of (this.actor.system._abilityBonuses ?? [])) {
+      if (bonus.key !== "keyword") continue;
+      if (bonus.mode !== CONST.ACTIVE_EFFECT_MODES.ADD) continue;
+      if (!bonus.filters.keywords.isSubsetOf(this.keywords)) continue;
+
+      this.keywords.add(bonus.value);
+    }
+
     for (const bonus of (this.actor.system._abilityBonuses ?? [])) {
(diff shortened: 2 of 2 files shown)


src/module/data/pseudo-documents/power-roll-effects/forced-movement-effect.mjs
```diff
@@ -102,26 +100,90 @@ export default class ForcedMovementPowerRollEffect extends BasePowerRollEffect {
...
+      if (!bonus.key.startsWith(prefix)) continue;
+      if (!bonus.filters.keywords.isSubsetOf(this.parent?.keywords)) continue;
+
+      const key = bonus.key.substring(prefix.length);
+      // Bonus change objects are stored as strings, convert to Number
+      bonuses[key] = Number(bonus.value);
+    }
+
+    return bonuses;
+  }
```
krbz999:
```suggestion
  /**
   * Collects the bonuses for forced movement effects.
   * Searches the actor's ability bonuses for keys "forced.[movementType]".
   * 
   * @returns {Record<string, number>} An object mapping movement types to their bonus values.
   */
  #collectForcedMovementBonuses() {
    const bonuses = {};
    const abilityBonuses = this.actor?.system._abilityBonuses ?? [];

    const prefix = "forced.";
    for (const bonus of abilityBonuses) {
      if (!bonus.key.startsWith(prefix)) continue;
      if (!bonus.filters.keywords.isSubsetOf(this.parent.keywords)) continue;

      const key = bonus.key.substring(prefix.length);
      // Bonus change objects are stored as strings, convert to Number
      bonuses[key] = Number(bonus.value);
    }

    return bonuses;
  }
```

krbz999 (commented):
Quick drive-by review prior to any testing.

src/module/data/pseudo-documents/power-roll-effects/forced-movement-effect.mjs
```diff
@@ -102,26 +100,90 @@ export default class ForcedMovementPowerRollEffect extends BasePowerRollEffect {
...
+      const bonus = bonuses[movementType] || 0;
+      const finalDistance = Number(baseDistance) + bonus;
+
+      if (!groups.has(finalDistance)) {
+        groups.set(finalDistance, []);
+      }
+      groups.get(finalDistance).push(movementType);
+
+      return groups;
+    }, new Map());
```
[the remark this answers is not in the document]
metrzero: TIL

src/module/data/pseudo-documents/power-roll-effects/forced-movement-effect.mjs
```diff
@@ -102,26 +100,90 @@ export default class ForcedMovementPowerRollEffect extends BasePowerRollEffect {
...
+  }
+
+  /* -------------------------------------------------- */
+
   /**
    * @param {1 | 2 | 3} tier
    * @inheritdoc
    */
   toText(tier) {
+    const bonuses = this.#collectForcedMovementBonuses();
```
JPMeehan: What's the logic of doing this here rather than in AbilityModel _applyAbilityBonuses when we're already looping through the bonuses?

metrzero: I went back and forth on this, and eventually landed here. My (loosely held) view was that since forced movement has no other mechanical interaction beyond how this text is displayed, I only fetch it here, rather than stashing it on the object.

I kept it separated out so it's easy to change. Lmk if you would like it stashed on the object during `_applyAbilityBonuses` instead.

PR #1337, opened 2025-11-03
[#1336] Add modifiable divisor for recovery values

### LINKED ISSUE #1336: Reserves of Strength Feature changes Recovery Value (half of Stamina)
### User Experience

It would be nice to have this feature auto-calculate the recovery value instead of manually adjusting

alinuxfan:
#1336 to set default to 1/3 but allow changes.


JPMeehan (approved):
Looks good, added in changelog + ts types pieces

PR #1342, opened 2025-11-04
Advancement drop improvements

JPMeehan:
- Dropping an advancement on another item (to make a copy) will check against the item's type restrictions
- ~~Dropping an item on an actor will attempt to run that advancement if the actor is of an appropriate level~~


JPMeehan:
> > * Dropping an item on an actor will attempt to run that advancement if the actor is of an appropriate level
> 
> Users may want to create the advancements, configure them slightly, and then re-run them. I don't think it's worth the complexity to run the advancement immediately.

Currently, we don't have a way to "re-run" advancements that don't have choices. Should that be a follow-up PR here?

krbz999:
> Currently, we don't have a way to "re-run" advancements that don't have choices. Should that be a follow-up PR here?

Hm, good point, I forgot about that. It will be good to have but if we have a fallback rn then current behavior (as per your latest commits) are fine for 0.9.

PR #1346, opened 2025-11-05
[#1170] Journal css

christopherhopper:
Added font support, themed window and content for journals to support MCDM-style asides and inverse-callout boxes. Support for light and dark mode.

### DIFF
--- a/src/styles/system/applications/components/embeds.css
+++ b/src/styles/system/applications/components/embeds.css
@@ -35,86 +35,82 @@ document-embed, .content-embed {
     }
   }
 
-  .roll-table-embed {
-    table-layout: fixed;
+  /* Embeds inside a Draw Steel application */
+  .draw-steel & {
+    .roll-table-embed {
+      table-layout: fixed;
 
 
-    thead {
-      border: none;
-      border-color: transparent;
-    }
+      thead {
+        border: none;
+        border-color: transparent;
+      }
 
-    th,
-    td {
-      width: auto;
-    }
+      th,
+      td {
+        width: auto;
+      }
 
-    th:first-child,
-    td:first-child {
-      width: 5em;
-      worder-wrap: break-word;
+      th:first-child,
+      td:first-child {
+        width: 5em;
+        word-wrap: break-word;
 
-      button {
-        margin: 0 auto;
+        button {
+          margin: 0 auto;
+        }
       }
     }
-  }
 
-  /* all embedded captions*/
-  /*draws gradient behind*/
-  caption {
-    text-align: center;
-    position: relative;
-    border-bottom: 2px solid #866d4b;
-    line-height: 0.75em;
-    padding-top: 1em;
-    width: 100%;
-    border-image: linear-gradient(to right, #ffffff03, #866d4b99, #ffffff03) 1;
-    border-image-slice: 1;
+    /* all embedded captions*/
+    /*draws gradient behind*/
+    caption {
+      text-align: center;
+      position: relative;
+      border-bottom: 2px solid #866d4b;
+      line-height: 0.75em;
+      padding-top: 1em;
+      width: 100%;
+      border-image: linear-gradient(to right, #ffffff03, #866d4b99, #ffffff03) 1;
+      border-image-slice: 1;
 
-    /*adds white block over gradient, behind text*/
-    strong {
-      background: var(--draw-steel-c-white);
-      padding: 0.5em 1em;
-      margin-top: -0.75em;
-      position: absolute;
-      transform: translate(-50%, 0%);
-    }
+      /*adds white block over gradient, behind text*/
+      strong {
+        background: var(--draw-steel-c-white);
+        padding: 0.5em 1em;
+        margin-top: -0.75em;
+        position: absolute;
+        transform: translate(-50%, 0%);
+      }
 
-    /*adds diamonds*/
-    strong::before {
-      content: "®";
-      font-family: Draw Steel Glyphs;
-      color: #866d4b;
-      position: absolute;
-      left: -0.5em;
-    }
+      /*adds diamonds*/
+      strong::before {
+        content: "®";
+        font-family: Draw Steel Glyphs;
+        color: #866d4b;
+        position: absolute;
+        left: -0.5em;
+      }
 
-    strong::after {
-      content: "®";
-      font-family: Draw Steel Glyphs;
-      color: #866d4b;
-      position: absolute;
-      right: -0.5em;
-    }
+      strong::after {
+        content: "®";
+        font-family: Draw Steel Glyphs;
+        color: #866d4b;
+        position: absolute;
+        right: -0.5em;
+      }
 
-    cite {
-      position: absolute;
-      top: 1.25em;
-      right: 0;
-    }
+      cite {
+        position: absolute;
+        top: 1.25em;
+        right: 0;
+      }
 
-    cite::before {
-      content: '';
-      clear: both;
-      margin-top: 3em;
-    }
-  }
-
-  .roll-table-sheet {
-    thead {
-      border: none;
-      border-color: transparent;
+      cite::before {
+        content: '';
+        clear: both;
+        margin-top: 3em;
+      }
     }
   }
 }


src/styles/system/applications/sheets/journal-entry-sheet.css
```diff
@@ -0,0 +1,233 @@
...
+  font-style: italic;
+  color: #866d4b;
+}
+/* fancy fieldset*/
+ .polygonal-fieldset {
+  position:relative;
+      border: 1px solid #333; /* Standard border for fallback and visual guide */
+      border-radius: 15px; /* Defines the size of the corner */
+      corner-shape: bevel; /* Creates a straight, angled corner */
+      color: background: var(--draw-steel-c-white);
```
JPMeehan: VSCode is marking an error here?

src/styles/system/applications/sheets/journal-entry-sheet.css
```diff
@@ -0,0 +1,233 @@
...
+        left: -1em;
+        width: calc(100% + 20px);
+        text-align: center;
+      }
+    }
+    .polygonal-fieldset-inverse {
+      position:relative;
+      border: 1px solid #866d4b; /* Standard border for fallback and visual guide */
+      border-radius: 15px; /* Defines the size of the corner */
+      corner-shape: bevel; /* Creates a straight, angled corner */
```
JPMeehan: Interesting that my vscode is calling this an unknown property

src/styles/system/applications/sheets/journal-entry-sheet.css
JPMeehan: The error is for the color: background

src/styles/system/applications/sheets/journal-entry-sheet.css
krbz999: Doesn't look like Firefox supports "corner-shape".

https://caniuse.com/mdn-css_properties_corner-shape

src/styles/elements/typography.css
```diff
@@ -1 +1,2 @@
 @import url("https://fonts.googleapis.com/css2?family=Roboto:wght@400;700&display=swap");
+@import url("https://fonts.googleapis.com/css2?family=Atkinson+Hyperlegible:ital,wght@0,400;0,700;1,400;1,700&family=Newsreader:ital,opsz,wght@0,6..72,200..800;1,6..72,200..800&display=swap");
```
krbz999: More of a question than part of the review since I'm not familiar with using this. Is this a lookup to an external site? If so, how does that work if not connected to the web?

src/styles/system/applications/sheets/journal-entry-sheet.css
```diff
@@ -0,0 +1,295 @@
+/* CSS for DrawSteelJournalEntrySheet */
+
+/*all*/
+.draw-steel.journal-sheet .window-content{
```
krbz999:
```suggestion
.draw-steel.journal-sheet .window-content {
```

src/styles/system/applications/sheets/journal-entry-sheet.css
```diff
@@ -0,0 +1,317 @@
+/* CSS for DrawSteelJournalEntrySheet */
+
+/*all*/
+.draw-steel.journal-sheet .window-content {
+  backdrop-filter: none;
+  background: var(--draw-steel-c-white);
+}
+.draw-steel .content-link,
+.inline-roll {
```
krbz999: Not name-spaced.

src/styles/system/applications/sheets/journal-entry-sheet.css
```diff
@@ -0,0 +1,317 @@
...
+  }
+  .toc li.page.active:last-of-type {
+    border-bottom-color: transparent;
+    border-image: linear-gradient(to right, var(--draw-steel-c-laserline), transparent) 1;
+  }
+}
+a.button:hover,
+button:hover {
+  background: var(--draw-steel-c-laserline);
+}
```
krbz999: Not name-spaced.

src/styles/system/applications/applications.css
```diff
@@ -39,3 +39,54 @@ dl.power-roll-display {
     margin: 0;
   }
 }
+
+/*laser lines on captions*/
+caption {
```
krbz999: Not name-spaced.

src/styles/variables/colors.css
```diff
@@ -10,6 +10,11 @@
   --draw-steel-c-failure: red;
   --draw-steel-c-item-header-bg: #bebbab;
   --draw-steel-c-item-alternating-bg: rgba(0, 0, 0, .05);
+  --draw-steel-c-laserline: #856C4B;
+  --color-shadow-primary: #ccccccaa;
+  --button-focus-outline-color: var(--draw-steel-c-dark);
+  --button-hover-background-color: var(--draw-steel-c-dark);
+  /*--input-text-color:var(--draw-steel-c-black);*/
```
krbz999: Afraid this is not gonna work. Right now, this would bleed over to every part of the platform due to `:root` above.

src/styles/variables/colors.css
```diff
@@ -22,4 +27,8 @@
   --draw-steel-c-groove: #11121f;
   --draw-steel-c-item-header-bg: #28232f;
   --draw-steel-c-item-alternating-bg: rgba(255, 255, 255, .05);
+  --draw-steel-c-laserline: #856C4B;
+  --color-shadow-primary: #ccccccaa;
+  --button-focus-outline-color: var(--draw-steel-c-laserline);
+  --button-hover-background-color: var(--draw-steel-c-laserline);
```
krbz999: Same comment here.

krbz999 (changes requested):
See my comments and questions. More issues I found in testing I will write here:

Light mode behaves strangely with some entirely invisible(?) background image when you resize it. I found this is also happening in dark mode.

Image

Image

Related to the above, light mode has an invisible header title when resized to be tall enough.

Image

Barely visible edit button.

Image

Barely visible roll button.

Image

I think I pointed this out but we have a rectangle here that was, I think, meant to have cut corners.

Image

Image

Low contrast header:

Image

I take it we were not able to resolve the lazer-lines? Is it required to be `H1 > strong`?

Image

The contrast in colors in the ToC (the horizontal lines which here are red and black) is very low. These are meant to show you which pages are currently in view.

Image

Hovering over the 'Add Page' button looks undesirable. The other two buttons have the same coloration, and in addition this hover effect applies to disabled buttons as well. Difficult to see, but the Next Page button here is actually disabled.

Image

In dark mode, the `THEAD` element in a table is not visibly distinct from the background.

Image

src/styles/system/applications/sheets/journal-entry-sheet.css
```diff
@@ -0,0 +1,457 @@
...
+    /* size of the cut */
+    width: 100%;
+    display: inline-block;
+    /*aspect-ratio: 1.5;*/
+    min-height: 80px;
+    position: relative;
+    z-index: 0;
+    padding: 20px;
+
+    .charm {
```
JPMeehan: Should this be more direct with the use of `>`?

src/styles/system/applications/applications.css
```diff
@@ -39,3 +39,54 @@ dl.power-roll-display {
...
+  }
+  cite::before {
+    content: '';
+    clear: both;
+    margin-top: 3em;
+  }
+}
+
+
+.system-draw-steel .roll-table-sheet {
```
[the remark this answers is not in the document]
krbz999: Still unresolved.

src/styles/variables/colors.css
```diff
@@ -1,25 +1,59 @@
...
   --draw-steel-c-tan: #b5b3a4;
   --draw-steel-c-white: #000000;
-  --draw-steel-c-black: #ffffff;
+  --draw-steel-c-black: whitesmoke;
   --draw-steel-c-groove: #11121f;
   --draw-steel-c-item-header-bg: #28232f;
   --draw-steel-c-item-alternating-bg: rgba(255, 255, 255, .05);
+  --draw-steel-c-color-shadow: #ccccccaa;
+  --draw-steel-c-laserline: #856C4B;
+  --color-shadow-primary: #ccccccaa;
```
krbz999: Also here.

krbz999 (dismissed):
It is shaping up, though a lot of my comments from the previous review still seem unresolved.

Please start reading reviews in full and responding to questions.

Now, we've discussed moving this to 0.10 so there is less pressure, sure, but here are my thoughts nonetheless for this round.

It's all looking a lot better, and with just some adjustments I think this could be ready for prime time.

Embedded tables are the big one, I think. This is what it looks like (with 4 different configurations of `cite` and `caption`):

Image

Do you want us to reconfigure the markup of embedded tables? We can do that. Just say the word.

I also feel like this has grown a bit unwieldy. I think we should bite the bullet and refactor this:
- create subdirectories for individual components that live on journal entries (asides, hints, etc)
- one file just for the general styling of a journal (colors, background image, and so forth).

Does this make sense?

src/styles/system/applications/sheets/journal-entry-sheet.css
```diff
@@ -0,0 +1,566 @@
...
+.draw-steel.journal-sheet .journal-entry,
+.draw-steel .journal-entry-page {
+  .hanging-indent {
+    padding-left: 1em;
+    /* Indents the entire block by 2em */
+    text-indent: -1em;
+    /* Pulls the first line back by 2em */
+  }
+
+  aside {
```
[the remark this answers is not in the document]
christopherhopper: changed to aside class="boxed" and dropped the top/bottom charms in favor of a 9-slice border

src/styles/system/applications/components/embeds.css
```diff
@@ -1,5 +1,6 @@
 /* Sheets do not include the document-embed wrapper but do use a `.content-embed` div class */
-document-embed, .content-embed {
+.system-draw-steel document-embed,
+.system-draw-steel .content-embed {
```
krbz999: The `system-draw-steel` class is redundant cus it's in the root.

Did you mean `.draw-steel`, which is the class applied to individual floating windows?

JPMeehan: I pushed a change to this by splitting some stuff up and then having a wrapper of `.draw-steel &` for the roll table embed pieces.

PR #1354, opened 2025-11-05
Null and Tactician Review

Stolassz:
Added:

- /gain and /apply enrichers to some abilities and features
- Other adjustments


JPMeehan (approved):
I've been adjusting the tables I find to properly use <thead>, if you're not comfortable with HTML editing I can keep doing that in reviews

PR #1355, opened 2025-11-05
[#1319] Add enricher options for additional resource types

### LINKED ISSUE #1319: Expand /gain enricher options
Other things the gain enricher could cover
- Epic resources
- Renown
- Wealth
- Victories

We should consider if the latter 3 should be split into an /award enricher, its unclear if that's meaningfully distinct (we don't have esoteric XP splitting rules like dnd5e)

metrzero:
Closes #1319 

- Added a central mapping of resource type to localization string to reuse throughout
- Broke out format string to accommodate per-resource variation in localization
- Updated docs

I kept the switch statement for modifying the token as that seemed most likely to need custom handling in the future, but it'd be simple to move to a similar mapping approach as for the format strings if that's preferred.

### DIFF
--- a/src/module/applications/ux/enrichers/roll.mjs
+++ b/src/module/applications/ux/enrichers/roll.mjs
@@ -299,12 +299,11 @@ function enrichGain(parsedConfig, label, options) {
   for (const c of parsedConfig) {
     const formulaParts = [];
     if (c.formula) formulaParts.push(c.formula);
-    c.type = c.type?.replaceAll("/", "|").split("|") ?? [];
     for (const value of c.values) {
       const normalizedValue = value.toLowerCase();
       // If the normalized value is present in the lookup object, add it to the config type
       if (allGainKeys.some((key) => key === normalizedValue)) {
-        c.type.push(normalizedValue);
+        c.type = normalizedValue;
       } else {
         formulaParts.push(value);
       }
@@ -315,7 +314,7 @@ function enrichGain(parsedConfig, label, options) {
     );
     if (c.formula) {
       linkConfig.formula = c.formula;
-      linkConfig.gainType = c.type[0]; // Require type to be specified
+      linkConfig.gainType = c.type; // Require type to be specified
       break; // Only use first formula
     }
   }


JPMeehan (dismissed):
This is really clean, and I think merits inclusion in 0.9 as it generally fits with the gain enricher work already done without really making requirements of the content work.

src/module/applications/ux/enrichers/roll.mjs
```diff
@@ -29,6 +29,22 @@ export const pattern = new RegExp(`\\[\\[/(?<type>${rollTypes.join("|")})(?<conf
 
 /* -------------------------------------------------- */
 
+/**
+ * Resource name to localization key mappings for gain enricher.
+ * Maps resource type identifiers to their i18n localization keys.
+ */
+const RESOURCE_LOCALIZATION_MAP = {
```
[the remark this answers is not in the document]
metrzero: I'm game. My original version had this also include update path, but I wasn't confident the token modification logic would remain consistent enough over time to standardize the call to `modifyTokenAttribute`. But more than happy to put that back and collapse down to a lookup rather than a switch.

src/module/applications/ux/enrichers/roll.mjs
```diff
@@ -29,6 +29,56 @@ export const pattern = new RegExp(`\\[\\[/(?<type>${rollTypes.join("|")})(?<conf
...
+ * @typedef {string} resource - The full path, relative to the actor, used to update the resource
+ * @typedef {string} resourceFormatString - Final key in the i18n path used for localization, relative to DRAW_STEEL.EDITOR.Enrichers.Gain.{MessageTitle|FormatString} (Default: "Default")
+ * @typedef {string} aliasFor - Value that linkConfig.gainType should be rewritten to before creating a link (Optional)
+ */
+const GAIN_RESOURCE_LOOKUP = {
+  epic: {
+    label: "DRAW_STEEL.Actor.hero.FIELDS.hero.epic.value.label",
+    resource: "hero.epic.value",
+  },
+  hr: {
```
JPMeehan: I think aliases can be their own object without the rest of the duplicate info

src/module/applications/ux/enrichers/roll.mjs
```diff
@@ -250,14 +293,17 @@ async function rollDamageHeal(link, event) {
 function enrichGain(parsedConfig, label, options) {
   const linkConfig = { type: "gain", formula: null, gainType: null };
 
+  const allGainKeys = Object.keys(GAIN_RESOURCE_LOOKUP).concat(Object.keys(GAIN_RESOURCE_ALIASES));
+
   // Parse the formula and type from configuration
   for (const c of parsedConfig) {
     const formulaParts = [];
     if (c.formula) formulaParts.push(c.formula);
     c.type = c.type?.replaceAll("/", "|").split("|") ?? [];
```
JPMeehan: Realizing there's some other bad copy here, `gain` enrichers aren't trying to support multiple types like damage enrichers have to. Also impacts a few lines later, should be just `gainType = c.type` not `c.type[0]`.

To elaborate: `[[/damage 1d3 type=fire/cold]]` is valid, `[[/gain 1d3 type=wealth/victory]]` is not

JPMeehan (approved):
Very happy with this

PR #1370, opened 2025-11-10
Fixes for updateFromCompendium helper

JPMeehan:
- Preserves certain properties through updates
- Switches from merge to hard-override of `system` for effects, allowing things like removing sudokus.

### DIFF
--- a/src/module/utils/update-from-compendium.mjs
+++ b/src/module/utils/update-from-compendium.mjs
@@ -32,7 +32,22 @@ export default async function updateFromCompendium(doc, options = {}) {
     if (!proceed) return;
   }
 
-  if (!options.embedOnly) await doc.update(compendiumUpdateData(compendiumDocument));
+  if (!options.embedOnly) {
+    let savedProps = {};
+    switch (doc.type) {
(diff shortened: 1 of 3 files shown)


src/module/utils/update-from-compendium.mjs
```diff
@@ -67,7 +67,13 @@ function compendiumUpdateData(doc) {
...
     case "Actor":
     case "Item":
-      return { _id: doc.id, system: documentData.system };
+      // Preserve project point usage
+      if (doc.type === "career") delete documentData.system.projectPoints;
+      // Preserve class level
+      else if (doc.type === "class") delete documentData.system.level;
+      // Preserve current project completion status
+      else if (doc.type === "project") delete documentData.system.points;
+      return { _id: doc.id, "==system": documentData.system };
```
krbz999: Wouldn't this end up setting pp/level/points to the initial value regardless if using `==`?

JPMeehan: You are correct. I moved this to a separate spot that does work.

PR #1373, opened 2025-11-11
Small Undead Update

Dragodoth:
Updated several Undead monsters.

Ghost - added Presence to potency for Spirited Away
Rotting Zombie - added {{potency}} to tier 3
Skeleton, Ghoul, Zombie - added @potency instead of hard coded values so they can interact with active effects.


JPMeehan (approved):
I'm probably going to use this PR to add in some final changelog entries and the like, but this LGTM

Issue #306, opened 2025-03-15
Include manual ancestry subtype field in NPC character sheet meta data filler


KJTailor:
If I would want to state that a humanoid enemy is for example a devil or Hakaan, there is no way to do so. Or that a hag is a fey/hag.

![Image](https://github.com/user-attachments/assets/75313c83-46d6-4358-87f8-65d8873d8337)

Please include

krbz999:
Would multiple keywords not suffice?

KJTailor:
Technically yes, but that would require someone to go through all the NPC key words and implement them, and that list would probably get really long, really quickly.

Unless of course, implementing a text typing field would be more hassle than it's worth

JPMeehan:
Yeah, this was one of the questions I wasn't sure about when I first drafted this section up. It's probably worth having an "extra keywords" field that gets added onto the main keywords? Not super sure on the *exact* way to handle this.

KJTailor:
That sounds like an idea. "extra keywords" field that allows to type in the special key word and that then can be displayed either together with the other key word, or behind the other keyword in parenthesis?
Happy to provide a screenshot from the 5e system to show how it looks like there, if that would help?

krbz999:
A free-form field, I think, we should generally try to shy away from. There has to be some form of structure to it, and it would ideally be able to adhere to i18n. But the question is then how to make it easy enough to add new keywords when needed.

JPMeehan:
> That sounds like an idea. "extra keywords" field that allows to type in the special key word and that then can be displayed either together with the other key word, or behind the other keyword in parenthesis? Happy to provide a screenshot from the 5e system to show how it looks like there, if that would help?

both Zhell and I have contributed to 5e, we're very familiar. The issue is figuring out what is both accessible and can enforce consistency

KJTailor:
Good to know, :) pros at work
While I get the consistency issue, people are already going crazy with homebrewing, so I feel like, a little leeway to allow for such thing, could go a long way?

PR #1377, opened 2025-11-12
[#306] Initial implementation for custom NPC keywords

metrzero:
This version does not yet incorporate manifest-based keywords, but I wanted to get a draft posted to start collecting feedback.

### DIFF
--- a/src/module/applications/apps/npc-keywords-config.mjs
+++ b/src/module/applications/apps/npc-keywords-config.mjs
@@ -1,0 +1,231 @@
+import { systemID, systemPath } from "../../constants.mjs";
+
+const { HandlebarsApplicationMixin, Application } = foundry.applications.api;
+
+/**
+ * Configuration menu for managing NPC keywords.
+ */
+export default class NPCKeywordsConfig extends HandlebarsApplicationMixin(Application) {
+  /**
+   * Local state for keywords being edited.
+   * @type {Set<string>}
+   */
+  #keywords = new Set();
+
+  /* -------------------------------------------------- */
+
+  /** @inheritdoc */
+  static DEFAULT_OPTIONS = {
+    id: "npc-keywords-config",
+    classes: ["draw-steel", "npc-keywords-config"],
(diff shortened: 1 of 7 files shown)


src/module/applications/apps/npc-keywords-config.mjs
```diff
@@ -0,0 +1,231 @@
...
+
+    // Close the application
+    this.close();
+  }
+}
+
+/*
+ * Populate initial keywords from settings into ds.CONFIG on ready
+ */
+Hooks.once("ready", async function () {
```
JPMeehan: Since we already have a ready hook in the root `draw-steel` method, I would prefer you setup some static function from this application that handles it, probably after migrations?

src/module/applications/apps/npc-keywords-config.mjs
```diff
@@ -0,0 +1,231 @@
+import { systemID, systemPath } from "../../constants.mjs";
+
+const { HandlebarsApplicationMixin, Application } = foundry.applications.api;
+
+/**
+ * Configuration menu for managing NPC keywords.
+ */
+export default class NPCKeywordsConfig extends HandlebarsApplicationMixin(Application) {
```
JPMeehan: Is there a reason this isn't just extending DSApplication?

src/module/applications/apps/npc-keywords-config.mjs
```diff
@@ -0,0 +1,231 @@
...
+    window: {
+      title: "DRAW_STEEL.Setting.NPCKeywords.Label",
+      icon: "fa-solid fa-tags",
+    },
+    position: {
+      width: 500,
+      height: "auto",
+    },
+    form: {
+      handler: NPCKeywordsConfig.prototype._onSubmit,
```
JPMeehan: I think it's likely you'll need to replace the submission handler, but it should be `static` and hard private with `#`, then use the `@this` annotation so it understands the calling context as the instance.

PR #1386, opened 2025-11-15
Leveled Treasures for Compendium

kenpoh01:
Added the leveled treasures in separate folders in the rewards compendium.


JPMeehan (commented):
We also ask for the page numbers, I know those aren't in the import data but do you mind filling those in? It's the source info menu if you edit in-app, but you can also probably just edit the json directly if you're comfortable with that.

kenpoh01:
I’ll do that today.  Had it done last night but notepad++ requires a save
before closing a window (with no warning!!!) it seems.
Also, Executioners Blade hadn’t parsed properly so that will be done too.
Ivan

kenpoh01:
Let me know if there's anything else that needs to change.

JPMeehan (approved):
Will need someone else to go in and implement enrichers & effects, but that can be follow up work.

PR #1388, opened 2025-11-15
fix: size of monster is now modified correctly

Mithror:
Was creating a custom animal with the Imposer trait and saw that the size was not properly updated. Seemed like an easy fix, so here it is!


JPMeehan:
Good catch!

JPMeehan (approved):

PR #1395, opened 2025-11-17
[#1067] Basic implementation of edge and bane tracking

### LINKED ISSUE #1067: Implement Bane & Edge tracking
There are a number of general edge and bane effects that the system can track without worrying about odd conditionals

- Bane/Edge to hit a target (e.g. "ability rolls made against them gain an edge")
- Bane/Edge on certain abilities (e.g. "Gain an edge on strikes")
- Bane/Edge on tests with a characteristic (e.g. "Gain an edge on Might tests")
- Bane/Edge on tests with a skill (e.g. "Gain an edge on on tests that use the Handle Animals skill")

This does not cover GM-adjudication effects like "edge on tests to hide or sneak" or "edge against demons, devils, undead, and creatures not native to your current world".

These values do not need to be configurable as part of the Hero/NPC sheet and can only exist as AE paths.

metrzero:
Closes #1067 

I currently append " (+1 Edge)" to the end of skills. This is probably more appropriately done with a new format string to give translators options on where to present the additional modifier. I'll make that change if the rest of this is OK.

The UI currently misbehaves when there are more than 2 global edges or banes, as the dropdown cannot represent them, and so it reverts to 0. I think it will be necessary to move away from a dropdown to some other control, but was asked to hold off making any UI changes beyond the labeling.

### DIFF
--- a/src/module/applications/apps/power-roll-dialog.mjs
+++ b/src/module/applications/apps/power-roll-dialog.mjs
@@ -149,7 +149,7 @@ export default class PowerRollDialog extends RollDialog {
     const pr = this.constructor.EdgePluralFormatter;
 
     // If there are skill modifiers, alter the label to include (+1 Edge) or (+2 Edges), etc.
-    context.skillOptions = Array.from(context.skills).reduce((accumulator, value) => {
+    context.skillOptions = Array.from(context.skills).map(value => {
       const { group } = list[value];
       let { label } = list[value];
       if (value in skillModifiers) {
@@ -168,10 +168,8 @@ export default class PowerRollDialog extends RollDialog {
         const formatter = game.i18n.getListFormatter("narrow");
         label += ` (${formatter.format(modifiers)})`;
       }
-      accumulator.push({ label: label, group: groups[group].label, value });
-      return accumulator;
-    }, []);
-    return;
+      return { label: label, group: groups[group].label, value };
+    });
   }
 
   /* -------------------------------------------------- */
--- a/src/module/data/item/ability.mjs
+++ b/src/module/data/item/ability.mjs
@@ -231,8 +231,6 @@ export default class AbilityModel extends BaseItemModel {
           case "power.roll.edges":
             this.power.roll.edges = this.power.roll.edges ?? 0 + (Number(bonus.value) || 0);
             break;
-          default:
-            break;
         }
       }
     }


src/module/applications/apps/power-roll-dialog.mjs
```diff
@@ -132,11 +132,30 @@ export default class PowerRollDialog extends RollDialog {
...
-      const { label, group } = list[value];
-      accumulator.push({ label, group: groups[group].label, value });
+      const { group } = list[value];
+      let { label } = list[value];
+      if (value in skillModifiers) {
+        const modifiers = [];
+
+        const edges = skillModifiers[value].edges;
+        const edgeName = edges === 1 ? "Edge" : "Edges";
+        if (edges > 0) modifiers.push(game.i18n.format("DRAW_STEEL.ROLL.Power.Modifier.Label", { number: `+${edges}`, mod: edgeName }));
```
JPMeehan: For this and the names, I'm pretty sure there are more formal ways to leverage i18n pluralization rules

src/module/data/actor/base.mjs
```diff
@@ -396,6 +396,38 @@ export default class BaseActorModel extends DrawSteelSystemModel {
    * @returns {Promise<DrawSteelChatMessage | null>}
    */
   async rollCharacteristic(characteristic, options = {}) {
+    const skillModifiers = {};
+
+    // Extract applicable modifier bonuses
+    for (const bonus of this._abilityBonuses) {
```
JPMeehan: I don't think I'd be expecting these to be Ability bonuses, just normal AEs modifying actor data

src/module/applications/apps/power-roll-dialog.mjs
```diff
@@ -132,11 +132,34 @@ export default class PowerRollDialog extends RollDialog {
    */
   _prepareSkillOptions(context) {
     const { list, groups } = ds.CONFIG.skills;
+    const skillModifiers = context.skillModifiers;
+
+    const pr = new Intl.PluralRules(game.i18n.lang, { type: "cardinal" });
```
JPMeehan: I wonder if we can cache this, e.g. make it a static property

src/module/data/item/ability.mjs
```diff
@@ -524,6 +524,23 @@ export default class AbilityModel extends BaseItemModel {
 
     // Restrained conditions check
     if (this.actor.statuses.has("restrained")) options.modifiers.banes += 1;
+
+    // Apply active effect modifiers from abilityModifier effects
+    for (const bonus of (this.actor.system._abilityBonuses ?? [])) {
+      if (!bonus.key.startsWith("power.")) continue;
+      if (!bonus.filters.keywords.isSubsetOf(this.keywords)) continue;
```
JPMeehan: Shouldn't the looping be handled in _applyAbilityBonuses where we're already looping and checking filters? Then on line 428 you can initialize banes & edges to `this.power.roll.banes` etc.

src/module/applications/apps/power-roll-dialog.mjs
```diff
@@ -132,11 +141,37 @@ export default class PowerRollDialog extends RollDialog {
    */
   _prepareSkillOptions(context) {
     const { list, groups } = ds.CONFIG.skills;
+    const skillModifiers = context.skillModifiers;
+
+    if (!this.constructor.EdgePluralFormatter) {
+      this.constructor.EdgePluralFormatter = new Intl.PluralRules(game.i18n.lang, { type: "cardinal" });
+    }
+    const pr = this.constructor.EdgePluralFormatter;
```
JPMeehan: Null coalescing here should be sufficient?
```suggestion
    const pr = this.constructor.EdgePluralFormatter ??= new Intl.PluralRules(game.i18n.lang, { type: "cardinal" });
```

src/module/applications/apps/power-roll-dialog.mjs
```diff
@@ -31,6 +31,15 @@ export default class PowerRollDialog extends RollDialog {
 
   /* -------------------------------------------------- */
 
+  /**
+   * Formatter to select the correct plural form for "Edge".
+   * Must be initialized before first use to ensure language is available.
+   * @type {Intl.PluralRules | null}
+   */
+  static EdgePluralFormatter = null;
```
krbz999: I think we can make this a getter. Or perhaps we want to extend `game.i18n`? Perhaps that's insane.

JPMeehan: v14 looks like they're just making local cached versions of formatters (e.g. Active Effect duration nonsense) which is why that was my suggestion.

src/module/data/actor/base.mjs
```diff
@@ -140,6 +140,13 @@ export default class BaseActorModel extends DrawSteelSystemModel {
       // Can consider removing in v14 after phases are introduced
       multiplier: 1,
     });
+
+    Object.values(this.characteristics).forEach((chr) => {
```
JPMeehan: Stylistically prefer `for of` loops
```suggestion
    for (const chr of Object.values(this.characteristics)) {
```

src/module/data/item/ability.mjs
```diff
@@ -201,6 +201,19 @@ export default class AbilityModel extends BaseItemModel {
...
+      if (bonus.key.startsWith("power.")) {
+        switch (bonus.key) {
+          case "power.roll.banes":
+            this.power.roll.banes = this.power.roll.banes ?? 0 + (Number(bonus.value) || 0);
+            break;
+          case "power.roll.edges":
+            this.power.roll.edges = this.power.roll.edges ?? 0 + (Number(bonus.value) || 0);
+            break;
+          default:
+            break;
```
JPMeehan: Can just leave a blank default case off
```suggestion
```

JPMeehan (commented):
Ultimately a bunch of stylistic notes & nitpicks, I think this PR is close to done.

metrzero:
I'm out of town for a few more days, but should be able to get to it this coming weekend.

metrzero:
I should be able to get to it today/tomorrow. Thanks for the reminder.

JPMeehan (approved):
Great, thanks for tackling this, it was a big project

PR #1402, opened 2025-11-22
More table formatting 2

kenpoh01:


JPMeehan (changes requested):
There's a lot of problems with this PR, I'll try to stitch them up but
- lots of duplicate creations of existing actors & malice features
- the Snaring Line malice ability can't be used by minions

JPMeehan (approved):
OK, fixed up now.

PR #1403, opened 2025-11-25
Add message partial rendering

krbz999:
- Defines a new data model subtype, `MessagePart`.
- Defines a new chat message subtype, `standard`.
- Each part defines the hbs template it uses.
- Each part is rendered in sequence.
- A header part is rendered as the very first thing (containing speaker, time, delete button).
- Whether a message is a roll is handled by the parts.
- Whether a message is visible is handled by the parts.

### DIFF
--- a/src/module/data/message/standard.mjs
+++ b/src/module/data/message/standard.mjs
@@ -60,10 +60,11 @@ export default class StandardModel extends foundry.abstract.TypeDataModel {
       actor: this.parent.speakerActor,
       user: game.user,
       rollData: this.parent.getRollData?.(),
+      isWhisper: this.parent.whisper.length,
+      whisperTo: this.parent.whisper.map(u => game.users.get(u)?.name).filterJoin(", "),
     };
 
-    const element = document.createElement("LI");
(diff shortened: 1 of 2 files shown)


src/module/data/message/parts/base.mjs
```diff
@@ -0,0 +1,124 @@
+const { ArrayField, JSONField, StringField } = foundry.data.fields;
+
+export default class MessagePart extends foundry.abstract.DataModel {
```
JPMeehan: class description

src/module/data/message/standard.mjs
```diff
@@ -0,0 +1,90 @@
+import MessagePart from "./parts/base.mjs";
+
+const { ArrayField, TypedSchemaField } = foundry.data.fields;
+
+export default class StandardModel extends foundry.abstract.TypeDataModel {
```
JPMeehan: Class description

JPMeehan (commented):
The overall framework makes a lot of sense, the devil will be migrating our stuff over to use it. I think this needs to target a feature branch.

PR #1408, opened 2025-11-26
Placement API

JPMeehan:
Copied (& adjusted) the placement API from dnd5e for use in summoning. Does not look to be meaningfully breaking with v14 although I think we'll eventually want to adjust the controls to allow adjusting elevation.

### DIFF
--- a/src/module/documents/token.mjs
+++ b/src/module/documents/token.mjs
@@ -41,7 +41,7 @@ export default class DrawSteelTokenDocument extends foundry.documents.TokenDocum
 
   /**
    * If the token's movementAction is invalid, force it to null (default).
-   * @returns {Promise<boolean>} True if the refresh has caused a change in movementAction, otherwise false.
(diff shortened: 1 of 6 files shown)


src/module/documents/token.mjs
```diff
@@ -40,8 +40,8 @@ export default class DrawSteelTokenDocument extends foundry.documents.TokenDocum
   /* -------------------------------------------------- */
 
   /**
-   * If the token's movementAction is invalid, force it to null (default)
-   * @returns {Promise<boolean>} true if the refresh has caused a change in movementAction, otherwise false
+   * If the token's movementAction is invalid, force it to null (default).
+   * @returns {Promise<boolean>} True if the refresh has caused a change in movementAction, otherwise false.
```
krbz999:
```suggestion
   * @returns {Promise<boolean>} Whether the refresh has caused a change in movementAction.
```

Issue #945, opened 2025-08-24
Create Reference System


JPMeehan:
As a user, I would like to be able to hover common keywords to display a tooltip explaining them.

Implementation notes
- Need some kind of enricher for this, dnd5e's usage of `&Reference` might be good to replicate
- The Glossary should definitely be included
- We should also consider if there are any un-enriched fields that would still like this kind of feature (likely not)

krbz999:
Syntax suggestions:
- `[[reference]]`
- `[[/reference]]`
- `[[refer]]`
- `[[/refer]]`

JPMeehan:
Considerations:

1. Should we have a rule-type page for separate full description vs. tooltip? e.g. for Agility, the glossary has "A characteristic that represents a creature’s coordination and nimbleness. p. 3" which references a slightly longer entry:
> Agility (A) represents coordination and nimbleness. A creature’s capacity to backflip out of danger, shoot a crossbow, dodge an explosion, or pluck keys from a guard’s belt is determined by Agility.

Plenty of other entries have much longer full descriptions, like Victories (quarter page on p. 7) vs. the one sentence summary at the end of the glossary

2. Many of the glossary entries reference content we model as items, not journal pages, e.g. Wode Elf has "A fey ancestry in tune with magical forests. p. 32". Similarly, classes have entries both for themselves and their heroic resource. Should we build those into the items themselves and rely on indexing?

JPMeehan:
Another good candidate would be transcribing the Art Descriptions from the Art Description Glossary into a journal as pages. 

https://files.mcdmproductions.com/DrawSteel/DrawSteelArtDescriptionGuide.pdf

As part of this, we should make sure to invest in having easy ways to carry the alt text of any image assets throughout the application, matching the highest levels of web a11y standards.

Issue #760, opened 2025-08-01
Implement Reference Enricher


JPMeehan:
The introduction has a substantive "Glossary Index" that would be good to allow easy reference to for tooltip purposes. It is theoretically possible that just setting up all of these entries as i18n strings and allowing inline use of `data-tooltip` would be preferable.

PR #1410, opened 2025-11-27
[#945, #760] Implement tooltip manager and base reference system

krbz999:
- Subclasses `TooltipManager` to allow for enriched tooltips. The target document must implement `richTooltip` in either the document class or the system model.
- Adds a `reference` page subtype which implements `richTooltip`.
- Adjusts content links for `JournalEntryPage` to allow for rich tooltips (if the page is of the given type).
- Adds a `[[reference]]` enricher that can be used to manually create a reference to a page, registered by id in `ds.CONFIG.references`.
- Adds a handlebars helper `ds-tooltip` which creates the markup for the placeholder tooltip that the mutation observer observes to replace with enriched content.

Below examples are chat messages with `[[reference myFirstReference]]` and a content link, respectively.

image

image

TODO:
- [ ] Implement custom sheet for `reference` pages, which currently uses just the default ProseMirror sheet.
- [ ] Consider how to implement alt text for images (add `richTooltip` to the `image` page type?).
- [ ] Add additional properties to the `reference` page subtype to support both short and long text.
- [ ] Use custom template for the enriched tooltip of a `reference` page (currently uses just the embed).
- [ ] Adjust visual style, and tagName, of created element from the enricher.
- [ ] Populate `ds.CONFIG.references` during or after `init` with references from subconfigs (eg conditions).

### DIFF
--- a/src/module/applications/ux/enrichers/reference.mjs
+++ b/src/module/applications/ux/enrichers/reference.mjs
@@ -20,22 +20,21 @@ export const pattern = /\[\[reference (?<config>[^\]]+)]](?:{(?<label>[^}]+)})?/
  * @type {TextEditorEnricher}
  */
 export async function enricher(match, options) {
-  let { config, label: fallback } = match.groups;
-  config = parseConfig(config);
+  const { config, label: fallback } = match.groups;
+  const parsedConfig = parseConfig(config);
 
-  if (!("id" in config)) {
-    const id = config.values.find(k => k in ds.CONFIG.references);
-    if (!id) return null;
-    config.id = id;
+  for (const value of parsedConfig.values) {
+    if (value in ds.CONFIG.references) parsedConfig.id ??= value;
   }
 
-  const uuid = ds.CONFIG.references[config.id];
+  const uuid = ds.CONFIG.references[parsedConfig.id];
+  if (!uuid) return null;
   const page = await fromUuid(uuid);
   if (!page || (page.type !== "reference")) return null;
 
   const dataset = {
-    referenceId: config.id,
-    tooltipHtml: game.tooltip.constructor.constructHTML({ uuid }),
+    referenceId: parsedConfig.id,
+    tooltipHtml: CONFIG.ux.TooltipManager.constructHTML({ uuid }),
   };
   const label = fallback?.trim() || page.name;
   return createLink(label, dataset);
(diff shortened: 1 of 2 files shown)


src/module/applications/ux/enrichers/reference.mjs
```diff
@@ -0,0 +1,50 @@
...
+ */
+export async function enricher(match, options) {
+  let { config, label: fallback } = match.groups;
+  config = parseConfig(config);
+
+  if (!("id" in config)) {
+    const id = config.values.find(k => k in ds.CONFIG.references);
+    if (!id) return null;
+    config.id = id;
+  }
```
JPMeehan: We usually loop through `config.values` and then do nullish assignment, since it's more resilient to adding new arguments. e.g.

```js
for (const value of config.values) {
  const normalizedValue = value.toLowerCase();
  if (normalizedValue in ds.CONFIG.references) config.id ??= normalizedValue;
}
```

We may want to demand exact references rather than the lower case normalizedValue, depending on if we want to demand that all of our CONFIG.references keys are lowercase or camel case.

src/module/helpers/interaction/tooltip-manager.mjs
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Helper method to construct tooltip HTML.
+   * @param {object} options
+   * @returns {string}
+   */
+  static constructHTML(options) {
```
JPMeehan: What's the advantage of this being static if we're going to have a reliable reference in `game.tooltip`?

src/module/documents/journal-entry-page.mjs
```diff
@@ -0,0 +1,12 @@
+/**
+ * A document subclass adding system-specific behavior and registered in CONFIG.Combatant.documentClass.
```
JPMeehan:
```suggestion
 * A document subclass adding system-specific behavior and registered in CONFIG.JournalEntry.documentClass.
```

src/module/config.mjs
```diff
@@ -1893,3 +1893,12 @@ export const sourceInfo = {
 };
 preLocalize("sourceInfo.books", { keys: ["label", "title"] });
 preLocalize("sourceInfo.licenses", { keys: ["label"] });
+
+/* -------------------------------------------------- */
+
+/**
+ * Record of journal entry pages of type `reference` that help display tooltips
+ * throughout the system.
+ * @type {Record<string, string>}
```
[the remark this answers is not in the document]
JPMeehan: I mean, we can also say that `[[reference conduit]]` or whatever isnt supported, and that's just text in the Glossary page. But I think that your structure does seem to make it pretty easy. It does raise the question of if those would need to be registered or if the indexing/registry I think we're going to develop will handle it. (I'll make a ticket for the registry)

Separately, I'm still not sure about what format we're expecting. Are we going to strictly require people to know these camelCase keys of valid references?

PR #1411, opened 2025-11-27
[#1412] Handle non-base-type AEs gracefully when determining context menu options

### LINKED ISSUE #1412: Context menu conditionals throw errors for non-"base" active effect types
For instance, if a module adds a new AE subtype, the context menu on an actor or item sheet will fail due to an unguarded reference to `system.end.type`

roth-michael:
Closes #1412
Elsewhere (like in `HeroModel#takeRespite`) this is handled with something like:
```js
if (!(effect.system instanceof BaseEffectModel)) continue;
```
Could do similar here, if that's desired, i.e.
```diff
  const effect = this._getEmbeddedDocument(target);
- return (effect.documentName === "ActiveEffect") && (effect.system.end.type === "save");
+ return (effect.system instanceof BaseEffectModel) && (effect.system.end.type === "save");
```
and
```diff
  const effect = this._getEmbeddedDocument(target);
- return effect.system.end.type === "save";
+ return (effect.system instanceof BaseEffectModel) && (effect.system.end.type === "save");
```

### DIFF
--- a/src/module/applications/sheets/actor-sheet.mjs
+++ b/src/module/applications/sheets/actor-sheet.mjs
@@ -656,7 +656,7 @@ export default class DrawSteelActorSheet extends DSDocumentSheet {
         icon: "<i class=\"fa-solid fa-fw fa-dice-d10\"></i>",
         condition: (target) => {
           const effect = this._getEmbeddedDocument(target);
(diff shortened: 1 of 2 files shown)


krbz999:
Is there a relevant issue for this? If not, please open one for the purpose of a changelog and assignment.

PR #1416, opened 2025-11-27
[#1064] Added XP advancement options

### LINKED ISSUE #1064: Add setting for configuring heroic advancement speed
### User Experience

As a Director, I would like the option to change the advancement speed from normal speed (every 16 xp) to double speed (every 8 xp) or half speed (every 32 xp). This would reflect on the character sheets and show the level up available button at the new thresholds.

cdverrett94:
Closes #1064

- Adds `ds.CONFIG.hero.xp_tracks` with normal, double, and half speed tracks set up. Modules could in theory add custom tracks.
- Added a setting to select an XP track from the options in the config.
- `ds.CONFIG.hero.xp_track` is now a getter and grabs the chosen track in the settings. This was done so that all existing code references to this property still work.

### DIFF
--- a/src/module/config.mjs
+++ b/src/module/config.mjs
@@ -867,7 +867,7 @@ export const hero = {
   get xp_track () {
     foundry.utils.logCompatibilityWarning("ds.CONFIG.hero.xp_track is deprecated. To get the currently configured "
       + "xp track use ds.CONFIG.hero.xpTrack instead. Setting an xp track "
-      + "has moved to an object in ds.CONFIG.hero.xpTracks.", { since: "0.10.0", until: "0.12.0" });
+      + "has moved to an object in ds.CONFIG.hero.xpTracks.", { since: "0.10.0", until: "0.12.0", once: true });
     return ds.CONFIG.hero.xpTrack;
   },
   /**


src/module/config.mjs
```diff
@@ -833,10 +834,33 @@ export const hero = {
...
+      label: "DRAW_STEEL.Setting.XPAdvancement.HalfSpeed",
+      track: [0, 32, 64, 96, 128, 160, 192, 224, 256, 288],
+    },
+  },
+  /**
+   * The chosen XP advancement option from the settings.
    * @type {number[]}
    */
-  xp_track: [0, 16, 32, 48, 64, 80, 96, 112, 128, 144],
+  get xp_track() {
```
krbz999: Wouldn't mind seeing this changed, too, to not use an underscore. Unsure if a shim and deprecation is warranted.

cdverrett94: I'll change this one. I can add a shim/deprecation too. I imagine the biggest use case for using this property would have been overriding the track. Should I add a shim for setting the value too or just leave it as a getter?

cdverrett94:
I'll resolve the conflicts when i get home later

Issue #1419, opened 2025-11-28
Build Compendium Document Registry


JPMeehan:
As a developer, I would like to be able to lookup key, cached information about some of our more prominent document subtypes like classes, for use in things like opening up the advancement system.

We'll start with building a registry of classes, with the keys being their DSID and the registry entries having their UUID, name, and heroic resource name.

JPMeehan:
Basic use case: allow abilities to say what heroic resource cost they have.

Advanced use case: properly open up class advancements to not need a hard coded list of ability/feature/subclass options.

PR #1422, opened 2025-11-28
Implemented first pass of DrawSteelRegistry

JPMeehan:
Basic outline for #1419 

Classes & subclasses obviously need registry for ability mapping. Kits and Perks are part of our Item Grant drop choices, so having a pre-compiled list seems helpful.

Current performance is a hair over a second (~1200ms), which seems well within acceptable parameters. The use of type__in massively helps here by having the server filter out irrelevant document types.

### DIFF
--- a/src/module/helpers/registry.mjs
+++ b/src/module/helpers/registry.mjs
@@ -41,41 +41,37 @@ export default class DrawSteelRegistry {
 
       const indices = docs.filter(idx => registryTypes.has(idx.type));
 
+      // Use "world" and "system" for world & system compendiums, elsewise the module ID
+      const packageId = pack.metadata.packageType !== "module" ? pack.metadata.packageType : pack.metadata.packageName;
+
       for (const idx of indices) {
         const dsid = idx.system._dsid || DrawSteelItem.generateDSID(idx.name);
+
+        const key = `${packageId}:${dsid}`;
+
         /** @type {RegistryEntry} */
         const registryEntry = {
+          key,
           dsid,
           name: idx.name,
           uuid: idx.uuid,
         };
+
         switch (idx.type) {
           case "class":
             registryEntry.primary = idx.system.primary;
-            if (this.class.has(dsid)) {
-              console.warn(`Replacing ${idx.type} registry entry for ${dsid}`);
-            }
-            this.class.set(dsid, registryEntry);
+            this.class.set(key, registryEntry);
             break;
           case "subclass":
             registryEntry.classLink = idx.system.classLink;
-            if (this.subclass.has(dsid)) {
-              console.warn(`Replacing ${idx.type} registry entry for ${dsid}`);
-            }
-            this.subclass.set(dsid, registryEntry);
+            this.subclass.set(key, registryEntry);
             break;
           case "perk":
             registryEntry.perkType = idx.system.perkType;
-            if (this.perk.has(dsid)) {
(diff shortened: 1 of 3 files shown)


src/module/helpers/registry.mjs
```diff
@@ -0,0 +1,95 @@
...
+
+  subclass = new DSRegistryCollection();
+
+  /* -------------------------------------------------- */
+
+  perk = new DSRegistryCollection();
+
+  /* -------------------------------------------------- */
+
+  kit = new DSRegistryCollection();
```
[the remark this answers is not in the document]
JPMeehan: Fixed

JPMeehan:
This is ready for full & proper review

src/module/helpers/registry.mjs
```diff
@@ -0,0 +1,123 @@
...
+            this.kit.set(dsid, registryEntry);
+            break;
+        }
+      }
+    }
+  }
+
+  /* -------------------------------------------------- */
+  /*  Registries                                        */
+  /* -------------------------------------------------- */
```
krbz999: The private properties are missing annotations, and the getters are missing `@type`.

src/module/helpers/registry.mjs
```diff
@@ -0,0 +1,123 @@
...
+        module: 1,
+        world: 2,
+      };
+
+      return typeMap[a.metadata.packageType] - typeMap[b.metadata.packageType];
+    });
+
+    for (const pack of itemPacks) {
+      // Fancy performant fetch
+      const docs = await pack.getDocuments({ type__in: ["class", "subclass", "perk", "kit"] });
```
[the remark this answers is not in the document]
JPMeehan: OK, we'll have to manually construct the DSIDs if `system._dsid` is empty. Should we refactor to a helper function there?

src/module/helpers/registry.mjs
```diff
@@ -0,0 +1,124 @@
...
+    const registryTypes = new Set(["class", "subclass", "perk", "kit"]);
+
+    for (const pack of itemPacks) {
+      // Need to re-call `getIndex` for `system._dsid` to be populated
+      const docs = await pack.getIndex();
+
+      const indices = docs.filter(idx => registryTypes.has(idx.type));
+
+      for (const idx of indices) {
+        const dsid = idx.system._dsid ?? idx.name.replaceAll(/(\w+)([\\|/])(\w+)/g, "$1-$3").slugify({ strict: true });
```
krbz999: The dsid creator here would be better as a static function on our Item class.

src/module/helpers/registry.mjs
```diff
@@ -0,0 +1,124 @@
...
+        const registryEntry = {
+          dsid,
+          name: idx.name,
+          uuid: idx.uuid,
+        };
+        switch (idx.type) {
+          case "class":
+            registryEntry.primary = idx.system.primary;
+            if (this.class.has(dsid)) {
+              console.warn(`Replacing ${idx.type} registry entry for ${dsid}`);
```
krbz999: Do we not want to be able to register two classes with the same dsid? If this registry should be used for an eventual Compendium Browser, and a user has implemented a class that replaces a core class, I would expect them to have the same dsid and both be available.

krbz999 (changes requested):

krbz999 (approved):

PR #1435, opened 2025-11-29
[#1396] Add resource gain PRE

### LINKED ISSUE #1396: Support gaining resources in Power Roll effect
### User Experience

As an example, the Tactician's Battle Cry grants a varying number of surges based on tier:
Image

The Conduit's Orison of Victory ability is similar.

This could be achieved by adding enricher support to effect text, or a new PRE type for gaining resources.

cdverrett94:
Closes #1396 

Technically the book only has abilities that grant surges via power roll, but supporting heroic/epic resources made sense and didn't really add complexity.

If this gets approved, I will probably also look to tackle the healing PRE.

### DIFF
--- a/src/module/data/pseudo-documents/power-roll-effects/gain-resource-effect.mjs
+++ b/src/module/data/pseudo-documents/power-roll-effects/gain-resource-effect.mjs
@@ -49,8 +49,9 @@ export default class GainResourcePowerRollEffect extends BasePowerRollEffect {
 
   /* -------------------------------------------------- */
 
-  /** @type {Intl.PluralRules} */
-  static pluralRules;
+  /** The i18n plural rules to apply.
+   * @type {Intl.PluralRules | null} */
+  static pluralRules = null;
 
   /* -------------------------------------------------- */
 


src/module/data/pseudo-documents/power-roll-effects/resource-gain-effect.mjs
```diff
@@ -0,0 +1,161 @@
...
+
+  /**
+   * Helper method to get the localization key for either singular or plural resources.
+   * @param {1|2|3} n     The tier.
+   * @returns {string}    The default value.
+   */
+  #getResourceKey(n) {
+    const tierAmount = this.resource[`tier${n}`].amount;
+    const resourceType = this.resource[`tier${n}`].type;
+    const pluralRules = new Intl.PluralRules(game.i18n.lang, { type: "cardinal" });
```
JPMeehan: I think we can cache this

JPMeehan (approved):
Looks clean and good; I'm definitely realizing though that we probably want to just bite the bullet and handle proper i18n pluralization rules, so maybe we just do it as follow up work in #1436

PR #1440, opened 2025-11-30
[#1424] Implement custom image page subtype and data model

### LINKED ISSUE #1424: Consider allowing for image pages to be used in references

JPMeehan:
Did a few things
0. Some cleanup (made metadata read-only, added _types file)
1. Registered a data model to the core-provided `image` JEP type.
2. Added `system.artDescription` which is the new default for its art text
3. Registered new image sheet that shows that property as a textarea
4. Filed core tickets to ask them to better support alt text https://github.com/foundryvtt/foundryvtt/issues/13602 and https://github.com/foundryvtt/foundryvtt/issues/13642

Demonstration:

image

Closes #1424

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@
   - Added `forced.pull`, `forced.push`, and `forced.slide` to adjust the distance of forced movement abilities. (#1321)
   - Added `keyword` to add keywords to abilities. (#1322)
   - Added `potency` to adjust the potency of abilities. (#1350)
+- Added `system.artDescription` as a valid property for Image journal entry pages, providing rich alt text. (#1424)
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


krbz999:
I'm a bit confused. If you filed core tickets, then we don't need this.

JPMeehan:
I think that while supporting improved tooltips here is good, this is ultimately *not* about references; this is just a11y stuff. Nothing we need to build references for are primarily images.

src/module/data/journal-entry-pages/image.mjs
```diff
@@ -0,0 +1,35 @@
...
+  /** @override */
+  static defineSchema() {
+    const schema = super.defineSchema();
+
+    const fields = foundry.data.fields;
+
+    schema.artDescription = new fields.StringField();
+
+    return schema;
+  }
```
krbz999: Would like us to consider this design.
```suggestion
  /** @override */
  static defineSchema() {
    return Object.assign(super.defineSchema(), {
      artDescription: new foundry.data.fields.StringField({ required: true }),
    });
  }
```

PR #1442, opened 2025-11-30
[#1428] Implemented reference links for large swathes of CONFIG

JPMeehan:
- Added `setup` hook to populate `ds.CONFIG.references`
- Added a *lot* more `reference` properties based on what shows up in the Heroes book glossary + the opening section of Monsters

Considerations
- We have likely key collisions for `none` in quite a few places, plus likely others.  Having some kind of `category` property (thus turning the references from a pure string: string mapping into a string: object mapping) would help users disambiguate.
- Not *all* of our internal keys are reasonable to guess at for reference purposes.

### DIFF
--- a/draw-steel.mjs
+++ b/draw-steel.mjs
@@ -228,7 +228,6 @@ Hooks.once("setup", () => {
   for (const path of referenceObjects) {
     const config = foundry.utils.getProperty(ds.CONFIG, path);
     for (const [key, { reference }] of Object.entries(config)) {
-      if (key in ds.CONFIG.references) console.warn(key, path);
       if (reference) ds.CONFIG.references[reference.identifier ?? key] = reference.uuid;
     }
   }


JPMeehan:
This needs strategic discussion & possible refactoring of the targeted code areas.

draw-steel.mjs
```diff
@@ -199,6 +199,36 @@ Hooks.once("i18nInit", () => {
...
+    "equipment.categories",
+    "equipment.armor",
+    "equipment.weapon",
+    "projects.types",
+  ];
+
+  for (const path of referenceObjects) {
+    const config = foundry.utils.getProperty(ds.CONFIG, path);
+    for (const [key, { reference }] of Object.entries(config)) {
+      if (key in ds.CONFIG.references) console.warn(key, path);
```
krbz999: Seems irrelevant. As does having the large `references` object pre-made with blank strings.

JPMeehan: Those blank strings will get filled in a follow-up PR, I just wanted to lay out what all I expected to be included. I've removed this console.warn call

PR #1463, opened 2025-12-05
Add language sorting to the language picker

DocSchlock:
QOL change to add sorting to the language picker options on the actor sheets, ordering them A-Z based on the label.

No clue if this is the best way to go about it, but it does work.


JPMeehan:
Better way would be to pass `sort=true` to the formInput/formGroup where it's used, as that handles both options and groups.

DocSchlock:
> Better way would be to pass `sort=true` to the formInput/formGroup where it's used, as that handles both options and groups.

You weren't kidding, that is a wayyyyy better way. I changed the PR to use that instead.

PR #1492, opened 2025-12-15
Fix to Glowing Eyes ability

kenpoh01:
Folder created in compendium for Granted Abilities and advancement updated.


JPMeehan (approved):
Looks good, thanks for making these fixes.

Issue #1444, opened 2025-12-01
Implement Message Parts


JPMeehan:
The introduction of Message Parts in #1403 is a very powerful tool but has implications for our current roll and message logic. Fortunately, it seems as though we can compartmentalize the work and refactor in parallel with minimal risk of merge conflicts.

PR #1509, opened 2025-12-23
[#1444] Implemented Chat Message Parts, Test Enricher

JPMeehan:
- Massive refactor of system to leverage chat message parts
- Ability use and Saving Throw converted to new system
- Implemented Tests as new feature
- Project Roll messages not converted yet; left to follow-up PR
- Some incidental roll pipeline changes
  - The "base roll" gets stored in the message as before
  - Target-specific roll copies go into parts
  - DSN no longer sees the copies, removed explicit DSN overrides
  - Changed return type of PowerRoll.prompt

image
image
image


src/module/data/pseudo-documents/message-parts/ability-result.mjs
```diff
@@ -0,0 +1,208 @@
...
+    const heroActors = ds.utils.tokensToActors().filter((a) => a.type === "hero");
+    for (const actor of heroActors) {
+      await actor.modifyTokenAttribute(path, Number(amount), true, false);
+    }
+  }
+
+  /* -------------------------------------------------- */
+
+  /**
+     * Create a new DamageRoll based on applying modifications to a given DamageRoll.
```
krbz999: Borked indentation.

src/module/data/pseudo-documents/message-parts/roll.mjs
```diff
@@ -0,0 +1,38 @@
...
+ */
+export default class RollPart extends BaseMessagePart {
+  /** @inheritdoc */
+  static get TYPE() {
+    return "roll";
+  }
+
+  /* -------------------------------------------------- */
+
+  static ACTIONS = {
```
krbz999: comment

src/module/data/pseudo-documents/message-parts/roll.mjs
```diff
@@ -0,0 +1,38 @@
...
+  /** @inheritdoc */
+  static TEMPLATE = systemPath("templates/sidebar/chat/parts/roll.hbs");
+
+  /* -------------------------------------------------- */
+
+  /** @inheritdoc */
+  async _prepareContext(context) {
+    await super._prepareContext(context);
+
+    context.ctx.buttons = [];
```
krbz999:
```suggestion
    context.ctx.buttons = this.rolls.map((roll, i) => {
      if (roll instanceof DamageRoll) return roll.toRollButton(i);
      return null;
    }).filter(_ => _);
```

src/module/data/pseudo-documents/message-parts/saving-throw.mjs
```diff
@@ -0,0 +1,96 @@
...
+
+  /** @inheritdoc */
+  static TEMPLATE = systemPath("templates/sidebar/chat/parts/saving-throw.hbs");
+
+  /* -------------------------------------------------- */
+
+  /** @inheritdoc */
+  static defineSchema() {
+    return Object.assign(super.defineSchema(), {
+      effectUuid: new DocumentUUIDField({ nullable: false, type: "ActiveEffect" }),
```
krbz999: Do we have instances of wanting multiple effects?

src/module/data/pseudo-documents/message-parts/test.mjs
```diff
@@ -0,0 +1,97 @@
+import BaseMessagePart from "./base-message-part.mjs";
+import { systemPath } from "../../../constants.mjs";
+import enrichHTML from "../../../utils/enrich-html.mjs";
+
+const { SchemaField, HTMLField } = foundry.data.fields;
+
+/**
+ * A part containing a Test roll and.
```
krbz999: Don't hold me in suspense like this.

src/module/data/pseudo-documents/message-parts/test.mjs
```diff
@@ -0,0 +1,97 @@
...
+  /* -------------------------------------------------- */
+
+  /**
+   * Reroll this test while expending a hero token.
+   *
+   * @this TestPart
+   * @param {PointerEvent} event   The originating click event.
+   * @param {HTMLElement} target   The capturing HTML element which defined a [data-action].
+   */
+  static async #heroReroll(event, target) {
```
krbz999: Similar here, a public API method.

src/module/config.mjs
```diff
@@ -1476,6 +1476,40 @@ preLocalize("Advancement", { key: "label" });
 
 /* -------------------------------------------------- */
 
+export const MessagePart = {
```
krbz999: comment.

templates/sidebar/chat/parts/content.hbs
```diff
@@ -0,0 +1,8 @@
+{{#if part.flavor}}
+<div class="message-part-flavor">
+  <em>{{part.flavor}}</em>
```
krbz999: Let's leave the italics to css.

templates/sidebar/chat/parts/roll.hbs
```diff
@@ -0,0 +1,13 @@
+{{#if part.flavor}}
+<div class="message-part-flavor">
+  <em>{{part.flavor}}</em>
```
krbz999: Let's leave the italics to css.

templates/sidebar/chat/parts/test.hbs
```diff
@@ -0,0 +1,18 @@
+{{#if part.flavor}}
+<div>
+  <em>{{part.flavor}}</em>
```
krbz999: Let's leave the italics to css.
Also looks like a missing css class here.

krbz999 (changes requested):
It looks good, and though my testing was somewhat limited, it all seems to work as well with few issues. I have a good amount of comments wrt the code, but nothing that would be a major change.

However, I would have appreciated that this was multiple PRs; I think it could easily have been one for the implementation, one for the migration, and one for the test enricher.

Lastly, we will need to play the dice sound when a roll is appended. By default, core only does this when a message is created - I have implemented similar in Ryuutama which you can use for reference but it amounts to checking how many rolls exist (and are visible) before and after and then emitting the sound effect.

src/module/data/pseudo-documents/message-parts/test.mjs
```diff
@@ -0,0 +1,97 @@
...
+
+    schema.results = new SchemaField({
+      tier1: new HTMLField(),
+      tier2: new HTMLField(),
+      tier3: new HTMLField(),
+      critical: new HTMLField(),
+    });
+
+    return schema;
+  }
```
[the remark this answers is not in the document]
JPMeehan: Well, the result table of a *test* could come from anywhere (this is the table option of the enricher). For example, you could have some adventure journal that just inline in the adventure text has results for opening a door or something similar. The enricher, being in the same html block as the table, grabs the tier results and then feeds them to the part to store & track.

JPMeehan:
For the sounds, I added a bit to ChatMessage#_onUpdate and filed a core ticket https://github.com/foundryvtt/foundryvtt/issues/13720

JPMeehan:
OK, all code pieces are revised, just need to improve the documentation on the table option

src/docs/Enrichers.md
```diff
@@ -72,6 +72,50 @@ The `[[/gain]]` enricher allows you to grant heroic resources (like Ferocity, Fo
...
+    <dd>
+        <p>Tier 2 Result Text.</p>
+    </dd>
+    <dt class="tier3">
+        <p>#</p>
+    </dt>
+    <dd>
+        <p>Tier 3 Result Text.</p>
+    </dd>
+</dl>
```
krbz999: OK, now that this has been fully explained, I can see how to use it and what the use case is.

I think this should get pulled out of this PR and that we should discuss internally whether there isn't a better option here.

src/module/data/pseudo-documents/power-roll-effects/applied-effect.mjs
```diff
@@ -169,6 +169,50 @@ export default class AppliedPowerRollEffect extends BasePowerRollEffect {
...
   /* -------------------------------------------------- */
 
+  /**
+   * Apply an effect to one or more actors.
+   * @param {string} tierKey    The tier of effect to apply.
+   * @param {string} effectId   A status effect or AE id.
+   * @param {object} [options]
+   * @param {Iterable<DrawSteelActor>} [options.targets] Defaults to all selected actors.
+   */
+  async applyEffect(tierKey, effectId, options = {}) {
```
krbz999: Needs an explicit error if you do not own the targets.

src/module/data/pseudo-documents/power-roll-effects/gain-resource-effect.mjs
```diff
@@ -160,11 +163,43 @@ export default class GainResourcePowerRollEffect extends BasePowerRollEffect {
...
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Give a hero some resources.
+   * @param {string} tierKey
+   * @param {object} [options]
+   * @param {Iterable<DrawSteelActor>} [options.targets] Defaults to all selected hero actors.
+   */
+  async applyGain(tierKey, options) {
```
krbz999: Needs an explicit error if you're not owner of the targets.

krbz999 (changes requested):

src/module/data/pseudo-documents/message-parts/test.mjs
```diff
@@ -0,0 +1,113 @@
...
+      if (token === false) return false;
+
+      newRoll.options = { ...newRoll.options, flavor: game.i18n.localize("DRAW_STEEL.ChatMessage.PARTS.test.HeroTokenReroll.flavor") };
+    }
+
+    const rolls = this.rolls.concat(newRoll);
+
+    await this.update({ rolls }, { notify: true, ds: {
+      dsn: { [this.id]: [rolls.length - 1] },
+    } });
```
krbz999: Why is this here and not in `_preUpdate`?

JPMeehan: As in diff the array length to adjust the options? It's possible, although maybe we don't always want to re-roll the dice & finding that difference/change is certainly going to be messier.

JPMeehan:
Removed all `resultHTML` construction in the Test part for now, will fill in with follow-up PRs to AbilityModel & creation of a new JEP type for the result tables.

krbz999 (approved):

PR #1510, opened 2025-12-23
[#1462] Add treasure embed with option for including project info

### LINKED ISSUE #1462: Implement toEmbed option for Treasures to include project details
As a content creator, I would like to nicely display the project details for treasures in their embed if passed an appropriate option like `includeProject=true` for use in journals.

cdverrett94:
Closes #1462 

By default, it will include the project details. Updated the treasure details sheet to use the embed in play mode like other embeds.

Screenshots of with and without project info.
treasure-wo-project
treasure-w-project


JPMeehan:
I think we don't want it on the equipment tab but the details tab is good.

cdverrett94:
Alright, here we go! Reversed the default.

equipment-wo-project

cdverrett94:
Conflict resolved.

Issue #864, opened 2025-08-17
Add advancement property for "repick on respite"


JPMeehan:
As a user, I would like to be provided a list of advancements I can swap out during a Respite, such as my choice of kit or kit-like feature.

Implementation note: Take inspiration from the unpicked skill & language dialogs.

JPMeehan:
This should be a proper app and I don't think that's a priority right now.

cdverrett94:
On this one, going through the book, it seems like there are two different ways you can repick on respite, either as a respite activity or just whenever you finish a respite. So I'm thinking of adding a new `StringField` for `repickOnRespite` where the limited choices are `null`, `"activity"`, or `"finish"`. This way we can differentiate to the user if they can just do it for free during a respite or if its takes their one respite activity. 

Then similar to the `_unfilledTraits`, have a `_respiteAdvancements` internal property that gets set during data prep to have a convenient way to pull those later.

JPMeehan:
This got done, #1136 represents the follow-up work

PR #1522, opened 2025-12-31
Add option for repick on respite to advancements

cdverrett94:
Related to #864 

Added the schema field to advancements for repick respites. This can be `null`, `"activity"`,  or `"finish"`. During item data prep, `_respiteAdvancements` gets populated with a different set for `activity` and `finish`.

I updated the Conduit's prayer and ward to be set to repick as a respite activity to have something to test with.

Future PRs would add the UI/UX to actually show the user their pickable advancements when they take a respite.

### DIFF
--- a/src/module/data/item/advancement.mjs
+++ b/src/module/data/item/advancement.mjs
@@ -54,6 +54,8 @@ export default class AdvancementModel extends BaseItemModel {
 
     const level = this.actor.system.level;
     for (const advancement of this.advancements) {
+      if (!advancement.levels.some(l => l <= level)) continue;
+
       // Populate _respiteAdvancements
       if (advancement.repick.respite) {
         respiteAdvancements[advancement.repick.respite] ??= new Set();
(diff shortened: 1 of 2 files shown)


src/module/data/item/advancement.mjs
```diff
@@ -52,6 +54,13 @@ export default class AdvancementModel extends BaseItemModel {
...
     const level = this.actor.system.level;
     for (const advancement of this.advancements) {
+      // Populate _respiteAdvancements
+      if (advancement.repick.respite) {
+        respiteAdvancements[advancement.repick.respite] ??= new Set();
+        respiteAdvancements[advancement.repick.respite].add(advancement.getRelativeUUID(this.actor));
+      }
+
+      // Populate _unfilledTraits
       if (!advancement.levels.some(l => l <= level)) continue;
```
JPMeehan: Shouldn't you be after the level check? Don't need to add in advancements above the level of the character in case there's some late-in-the-game bonus.

JPMeehan (approved):
LGTM

PR #1527, opened 2026-01-04
[#1520] Add AE targetable path for canFlank

### LINKED ISSUE #1520: Can Flank status
### User Experience

As a Director, I would like some kind of "can flank" (or perhaps "cannot flank") status so that I can implement monster features such as the Crawling Claw's "Disorganized" trait (Delian Tomb Encounters pg. 35) which prevents the monster from flanking with its allies.

cdverrett94:
Closes #1520 

Also changed some stuff based on the initial discussion in discord.

### DIFF
--- a/src/module/canvas/placeables/token.mjs
+++ b/src/module/canvas/placeables/token.mjs
@@ -33,9 +33,9 @@ export default class DrawSteelToken extends foundry.canvas.placeables.Token {
     // Defeated/Dead actors cannot flank
     if (this.actor.statuses.has(CONFIG.specialStatusEffects.DEFEATED)) return false;
     // Check if active effects have modified the ability to flank
-    if (!this.actor.system.statuses.canFlank) return false;
+    if (!this.actor.system.statuses?.canFlank) return false;
     // Checking if active effects have restricted triggered abilities
-    return !this.actor.system.restrictions.type.has("triggered");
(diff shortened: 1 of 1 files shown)


src/module/canvas/placeables/token.mjs
```diff
@@ -28,9 +28,12 @@ export default class DrawSteelToken extends foundry.canvas.placeables.Token {
...
   get canFlank() {
-    if (!this.actor) return true;
-    // Dead actors cannot flank
-    if (this.actor.statuses.has("dead")) return false;
+    // For orphaned tokens, assume they cannot flank
+    if (!this.actor) return false;
+    // Defeated/Dead actors cannot flank
+    if (this.actor.statuses.has(CONFIG.specialStatusEffects.DEFEATED)) return false;
+    // Check if active effects have modified the ability to flank
+    if (!this.actor.system.statuses.canFlank) return false;
```
krbz999: An actor subtype might not implement `system.statuses`.


```suggestion
    if (!this.actor.system.statuses?.canFlank) return false;
```

krbz999: If you could change the return value below as well to accomodate that, that would be good.

cdverrett94:
> LGTM. The term `canFlank` feels off to me but I can't say I feel strongly about it.

I agree on the term 'canFlank`, but I also cant think of a better single word that would fit.

PR #1536, opened 2026-01-12
Advancement Logic Refactor

JPMeehan:
- Broke the current advancement logic into three new classes, `AdvancementChain`, `AdvancementNode`, and `AdvancementLeaf`
- One chain to manage the whole process, with an arbitrary number of nodes, and leaves underneath them. A leaf is equivalent to the previous `choice`, and is constructed by the advancement so we don't have to hard-code type-specific logic in these classes.
- It's possible that the AdvancementChain logic should just be absorbed by the configuration dialog but I'm not strictly sure that's a huge upside. I'm also open to adjusting the names; leaf doesn't *quite* convey its optional nature as either an endpoint for traits or a bridge for other nodes in the case of an Item Grant.
- This is not intended to include any new functionality, it's just paving the way for future upgrades

### DIFF
--- a/src/module/data/pseudo-documents/advancements/item-grant-advancement.mjs
+++ b/src/module/data/pseudo-documents/advancements/item-grant-advancement.mjs
@@ -149,6 +149,7 @@ export default class ItemGrantAdvancement extends BaseAdvancement {
   async createLeaves(node) {
     const promises = [];
     for (const { uuid } of this.pool) {
+      // TODO: Optimize DB calls
       /** @type {DrawSteelItem} */
       const item = await fromUuid(uuid);
       if (!item) continue;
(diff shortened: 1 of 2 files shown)


src/module/data/pseudo-documents/advancements/item-grant-advancement.mjs
```diff
@@ -142,6 +145,23 @@ export default class ItemGrantAdvancement extends BaseAdvancement {
...
+    const promises = [];
+    for (const { uuid } of this.pool) {
+      /** @type {DrawSteelItem} */
+      const item = await fromUuid(uuid);
+      if (!item) continue;
+      const leaf = node.choices[item.uuid] = new AdvancementLeaf(node, item.uuid, item.toAnchor().outerHTML, { item });
+      if (!item.supportsAdvancements) continue;
+
+      promises.push(...node.chain.createNodes(item, { parentLeaf: leaf }));
+    }
```
krbz999: This doesnt need to hold up this PR but I would like to see an improvement here to avoid calling `fromUuid`.

PR #1537, opened 2026-01-13
[#945] Remaining reference pages

### LINKED ISSUE #945: Create Reference System

JPMeehan:
Closes #945


JPMeehan:
Last two pushes did the rest of the pages, and then also removed the folders from the compendium because there are so few journals it wasn't worth it.

PR #1542, opened 2026-01-16
Update icon element class

krbz999:
Updates the `DSIconElement` class:
- Store only raw html, no elements.
- Replace strings such as `__1__` with random ids; this allows for using `mask` and `clipPath` in the SVGs which requires having an id; these are then replaced when the element is inserted, removing the possibility of duplicate ids across the DOM.

Example:
```svg
<mask id="__1__">
  <rect width="100%" height="100%" fill="black" />
</mask>
<g mask="url(#__1__)" fill="purple">
  <circle cx="120" cy="75" r="50" />
</g>
```


JPMeehan (approved):
LGTM

PR #1545, opened 2026-01-17
[#1448] Convert project rolls to use message parts

### LINKED ISSUE #1448: Migrate Project Rolls to Message Parts

cdverrett94:
Closes #1448 

I used `TestPart` as a guide for making the `ProjectPart`, so hopefully this looks good.

I also found a couple of bugs with how the project events were handled so fixed those. Notably, even when the setting was for `roll`, milestones could still trigger an event.

### DIFF
--- a/src/module/data/pseudo-documents/message-parts/project.mjs
+++ b/src/module/data/pseudo-documents/message-parts/project.mjs
@@ -135,7 +135,7 @@ export default class ProjectPart extends BaseMessagePart {
   /**
    * Make the breakthrough roll, updating the project with the new points, and adding the roll to the project part.
    */
-  async rollBreakthrough(target, event) {
+  async rollBreakthrough() {
     const project = this.project;
     if (!project) return;
 


cdverrett94:
Good catch! I've removed it.

cdverrett94:
I just realized there is a `ContentPart`. That might be better to use for the events text instead of just appending to the bottom of the `ProjectPart`. Might look better for the display.

src/module/data/pseudo-documents/message-parts/project.mjs
```diff
@@ -85,15 +86,29 @@ export default class ProjectPart extends BaseMessagePart {
...
+   */
+  async updateEventText() {
+    const eventSetting = game.settings.get(systemID, "projectEvents");
+    if ((eventSetting === "none") || !this.events) return;
+
+    let eventText = "";
+    if (eventSetting === "roll") eventText = game.i18n.localize("DRAW_STEEL.Item.project.Events.RollTriggered");
+    else if (eventSetting === "milestone") eventText = game.i18n.format("DRAW_STEEL.Item.project.Events.MilestoneTriggered", { events: this.events });
+
+    const eventPart = this.parent.parts.find(p => p instanceof ContentPart);
```
JPMeehan: Can use `documentsByType` to leverage cached access

JPMeehan:
> Okay, I updated the event text to use the message content and a content part.

Are we concerned about multiple events in the same message, since the content part is functionally unique?

cdverrett94:
> > Okay, I updated the event text to use the message content and a content part.
> 
> Are we concerned about multiple events in the same message, since the content part is functionally unique?


Honestly? I think me accounting for multiple events was overkill in the first place. Triggering multiple events in one respite is insanely statistically improbable. You'd need to get like like 3 or more breakthroughs in a row.

JPMeehan:
Re-reading the rules in the book, yeah one event is perfectly fine; need at least 200 points for a project to *consider* two events, which means a minimum of 70 points/event. And the "roll for event" is just a d6 for the whole party.

src/module/data/pseudo-documents/message-parts/project.mjs
```diff
@@ -85,15 +86,29 @@ export default class ProjectPart extends BaseMessagePart {
...
+    if (eventPart && (eventSetting === "milestone")) {
+      await this.message.update({ content: eventText });
+    } else {
+      await this.message.update({ content: eventText });
+      ContentPart.create({
+        type: "content",
+      }, { parent: this.message });
```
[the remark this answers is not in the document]
cdverrett94: Let me know if the way I've done this is the best way to handle it.

src/module/data/pseudo-documents/message-parts/project.mjs
```diff
@@ -0,0 +1,175 @@
...
+
+  /** @inheritdoc */
+  static TEMPLATE = systemPath("templates/sidebar/chat/parts/project.hbs");
+
+  /* -------------------------------------------------- */
+
+  /** @inheritdoc */
+  static defineSchema() {
+    return Object.assign(super.defineSchema(), {
+      projectUuid: new DocumentUUIDField({ nullable: false, type: "Item" }),
```
krbz999: Need to validate that this is not in a pack, otherwise the `project` getter can throw an error.

src/module/data/pseudo-documents/message-parts/project.mjs
```diff
@@ -0,0 +1,175 @@
...
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Fetches the project from the UUID. Can return null if the project no longer exists.
+   * @returns {DrawSteelActiveEffect | null}
+   */
+  get project() {
+    return fromUuidSync(this.projectUuid);
+  }
```
krbz999: Not an effect.

PR #1549, opened 2026-01-18
Implemented glyphs in ability metadata

JPMeehan:
Honestly we could probably consider dropping the metadata section for abilities in the actor sheet, its redundant with the item row.

image
image
image


JPMeehan:
I realized that we had redundant coverage of embeds and decided to coalesce in just the `components/embeds.css` file instead of having the separate `data` folder structure.

krbz999 (approved):

PR #1550, opened 2026-01-18
[#342] Reactive Test Abilities

### LINKED ISSUE #342: Allow "Inverted" abilities that require a test from targeted creatures
Is that something doable? If the ability is used, it presents a button in the chat window, like the damage apply, and the target must click that button, to then roll a save with the required characteristic,

JPMeehan:
image

image

Closes #342

### DIFF
--- a/src/module/data/actor/creature.mjs
+++ b/src/module/data/actor/creature.mjs
@@ -80,7 +80,7 @@ export default class CreatureModel extends BaseActorModel {
   /* -------------------------------------------------- */
 
   /**
-   * Prompt the user for what types.
+   * Perform a power roll using a characteristic.
    * @param {string} characteristic   The characteristic to roll.
    * @param {object} [options]        Options to modify the characteristic roll.
    * @param {Array<"test" | "ability">} [options.types] Valid roll types for the characteristic.


src/module/data/pseudo-documents/message-parts/test.mjs
```diff
@@ -39,22 +46,23 @@ export default class TestPart extends BaseMessagePart {
 
   /* -------------------------------------------------- */
 
+  /**
+   * Finds the most recent test roll (useful for managing rerolls).
+   */
+  get latestTest() {
+    return this.rolls.filter(r => r instanceof ds.rolls.PowerRoll).at(-1);
+  }
```
krbz999: Well now I wonder if we should add time stamps to the rolls.

src/module/data/pseudo-documents/message-parts/test.mjs
```diff
@@ -98,12 +135,108 @@ export default class TestPart extends BaseMessagePart {
...
+    }
+
+    const damageRoll = pre.toDamageRoll(tier, { damageSelection });
+    await damageRoll.evaluate();
+
+    const options = { notify: true };
+
+    if (!damageRoll.isDeterministic) options.ds = { dsn: { [this.id]: [this.rolls.length] } };
+
+    await this.update({ rolls: this.rolls.concat(damageRoll) }, options);
```
krbz999: Any use case for this method to call a public method? I would think so.

PR #1553, opened 2026-01-19
[#1269] Add potency enricher

### LINKED ISSUE #1269: Implement potency enricher for ability effects
Some abilities reference potencies in their Effect line. We should have a specialized enricher that formats with our glyphs and reflects if the item has a parent or not (has a text or numerical value)

cdverrett94:
Closes #1269 

Adds the potency enricher. The `BasePowerRollEffect` previously had a `translatePotencyGlyphs`. I removed it and made a version that could be reusable for both the enricher and ability display.

This doesn't handle the compendium updates. I can do that either in this PR or in a separate. I just wanted to get approval for the formatting on the enricher before I did that.

### DIFF
--- a/src/module/applications/ux/enrichers/potency.mjs
+++ b/src/module/applications/ux/enrichers/potency.mjs
@@ -19,6 +19,14 @@ export const pattern = new RegExp("\\[\\[(?<type>potency)(?<config> .*?)?]](?!])
 /* -------------------------------------------------- */
 
 /**
+ * A cached list of valid rollKeys for characteristics.
+ * @type {Set<string>}
+ */
+const rollKeys = new Set();
+
+/* -------------------------------------------------- */
+
+/**
  * Enricher function.
  * @type {TextEditorEnricher}
  */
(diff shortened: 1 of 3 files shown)


src/module/applications/ux/enrichers/potency.mjs
```diff
@@ -0,0 +1,70 @@
...
+export async function enricher(match, options) {
+  let { config, label: fallback } = match.groups;
+
+  /** @type {ParsedConfig} */
+  const parsedConfig = parseConfig(config);
+  parsedConfig._input = match[0];
+
+  const characteristics = ds.CONFIG.characteristics;
+  const rollKeys = new Set(Object.values(characteristics).map(c => c.rollKey.toLowerCase()));
+  const potencyStrengths = new Set(["weak", "average", "strong"]);
```
JPMeehan: 1. Can probably cache this? Like, don't need to always reconstruct
2. I feel like this should be part of a CONST or something.

JPMeehan (commented):
First pass looks good, but I want to do a deeper review later.

PR #1554, opened 2026-01-20
[#1140] Add AE keys for treasure bonuses to stamina and damage.

### LINKED ISSUE #1140: Prevent treasure bonuses to stamina and roll damage from stacking.
### User Experience

According to the rules, Abilities can only benefit from one weapon or implement at a time (implying you may be able to switch which you want at the time of using an ability?) but more importantly:
> If two treasures give a creature a bonus to their Stamina or a bonus to the rolled damage of their abilities, only the higher bonus applies unless the treasure’s description notes otherwise.

So preferably we have a default rule where it doesn't stack and takes the highest result, but a means for homebrew and future treasures to ignore that rule.

cdverrett94:
Closes #1140 

The compendium updates haven't been done yet. I can go through and do those at a later time.

It did require a bit more to do the treasure damage bonuses than I posed in #1140. `AbilityModel#_applyAbilityBonuses` runs after actor prep, so I had to had a `preparePostAbilityPrepData` to PREs to run after the ability bonuses ran.

Because I was adding the stuff for applying treasure bonuses to each tier, I also added a field for `damage.bonuses.value` for allowing applying a value to each tier without having to have a change for all three tiers with the same value.

I also incidentally found a bug in the `Formulafield#_applyChangeUpgrade/Downgrade` that I fixed.


JPMeehan (approved):
Good catch on the complexity of the damage bonuses in the current environment.

Issue #1502, opened 2025-12-17
Create TOC view for Journal compendiums 


JPMeehan:
Take inspiration from dnd5e and SWADE

PR #1557, opened 2026-01-21
[#1502] Table of Contents

JPMeehan:
image
image

### DIFF
--- a/src/module/applications/sidebar/apps/table-of-contents.mjs
+++ b/src/module/applications/sidebar/apps/table-of-contents.mjs
@@ -64,6 +64,7 @@ export default class DrawSteelCompendiumTOC extends foundry.applications.sidebar
 
   /**
    * Valid entry types, as represented by `flags.draw-steel.type`.
+   * @remarks Left as type literal rather than a Record for typing purposes.
    */
   static ENTRY_TYPES = {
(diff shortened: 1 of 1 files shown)


krbz999 (commented):
Config menu:
- submit on change, show all relevant inputs
- get rid of sub-sub config for "special"
- larger fieldsets, dont need to combine inputs into fewer form-groups
- make use of rootId
- no need to allow for it being resizable, look into whether the fieldsets can be collapsed/expanaded
- add icon to header

ToC app:
- maybe make columns (slightly) wider min width
- margin above and to left of "Contents"?
- it seems to use "Interface" theme setting instead of "Application"

src/module/applications/sidebar/apps/table-of-contents.mjs
```diff
@@ -0,0 +1,237 @@
...
+        Object.assign(context.header, {
+          title: tocFlags.title || page?.name,
+          content: page?.text.content,
+          cssClass: "",
+        });
+        continue;
+      }
+
+      const data = {
+        type, tocFlags,
```
krbz999:
```suggestion
        tocFlags, type,
```

krbz999 (approved):

PR #1571, opened 2026-01-23
[#1426] Customize Reference Tooltip

### LINKED ISSUE #1426: Use a custom hbs template for the enriched tooltip of `reference` pages

JPMeehan:
Took some inspiration from our journal CSS. Tooltips appear to always be dark mode.

image

Closes #1426


krbz999:
Tooltips are indeed always in dark-mode, but this can be adjusted if we want, and quite easily since we extend TooltipManager.

Issue #950, opened 2025-08-24
Conduit advancement is giving both Domain Features at level 1


Lindharin:
### What happened?

When creating a new Conduit character, the advancement process is giving both domain features at level 1.  Per the Heroes book on page 97, you are supposed to "Choose one of your domains. You gain a domain feature for that
domain, as shown on the 1st-Level Conduit Domain Features table. Additionally, you gain a skill from the chosen domain, selected from the skill group indicated on the table."  I think you're only supposed to get one at 1st level, and the second one at 2nd level.

<img width="867" height="1129" alt="Image" src="https://github.com/user-attachments/assets/c0ed0ab6-2924-4d6c-ba1b-b9e1a7a8ea28" />

### Reproduction Steps

Create a new character, add the Conduit class, choose 2 domains - I chose Fate and War.  It tries to give both the Oracular Visions and Sanctified Weapons features with corresponding skills.  

### Bug Checklist

- [x] The issue occurs while all Modules are disabled
- [ ] I am using The Forge for hosting

JPMeehan:
Relevant item grant texts. Another option is we have a straight up "optional" flag or something that allows people to opt out of whichever one they're *not* picking; then we can keep the choices all on the subclasses.

Level 1
> Choose one of your domains. You gain a domain feature for that domain, as shown on the 1st-Level Conduit Domain Features table. Additionally, you gain a skill from the chosen domain, selected from the skill group indicated on the table.

Level 2
> You gain the 1st-level domain feature and ability to choose a skill for the domain you selected at 1st level but whose domain feature you didn’t take at that level.
> Choose one of your domains. You gain a heroic ability from that domain, as shown on the 2nd-Level Conduit Domain Abilities table.

Level 4
> Choose one of your domains. You gain a domain feature for that domain, as shown on the 4th-Level Conduit Domain Features table.

Level 5
> You gain the 4th-level domain feature for the domain whose feature you didn’t select at that level.

Level 6
> Choose one of your domains. You gain a heroic ability from that domain, as shown on the 6th-Level Conduit Domain Abilities table.

Level 7
> Choose one of your domains. You gain a domain feature for that domain, as shown on the 7th-Level Conduit Domain Features table.

Level 8
> You gain the 7th-level domain feature for the domain whose feature you didn’t select at that level.

Level 9
> Choose one of your domains. You gain a heroic ability from that domain, as shown on the 9th-Level Conduit Domain Abilities table.

PR #1581, opened 2026-01-25
[#950] Conduit Advancements

JPMeehan:
- Moved prerequisites from TitleModel up to FeatureModel
  - Added level & DSID prerequisite fields. 
  - DSID is assumed to be OR, since it represents classes or subclasses. Can re-evaluate structure pending Shark's conditional logic implementation.
  - Level is not currently used but will be useful for broadening selection (e.g. non-core options for abilities)
- Conduit advancements are now on the class instead of subclass, leverage prerequisites to filter domain-specific options.
- Various typing and documentation fixes that I forgot to put into 0.9 that I had found on the old 950 branch.

### DIFF
--- a/src/module/applications/apps/advancement/item-grant-configuration-dialog.mjs
+++ b/src/module/applications/apps/advancement/item-grant-configuration-dialog.mjs
@@ -101,7 +101,7 @@ export default class ItemGrantConfigurationDialog extends DSApplication {
     if (!this.#fulfilledDSID) {
       this.#fulfilledDSID = new Set(this.node.chain.actorSubclasses.map(i => i.dsid));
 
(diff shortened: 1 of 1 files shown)


krbz999 (approved):
I am going to approve this, with the caveat that we need to implement some follow-up work:
- Changing the subclass during advancement should clear irrelevant choices (such as the subclass features from a subclass that is no longer picked).
- You should not be able to pick the same feature at level 1 and 2, or 4 and 5 (etc). Currently the choice of an already owned feature is still presented.

My thinking with us refactoring advancements was that we would be able to more easily scan the nodes to see if they were still "connected", and if not, remove them.

Issue #1594, opened 2026-02-04
Improve rollCharacteristic to support other modules (e.g. Monk's TokenBar)


Mithror:
### User Experience

I would like to create some automated traps which would require the players to make rolls (e.g. Agility tests). In order to get this to work some changes need to happen to the way rollCharacteristic returns rolls. In general there should be an option to return rolls not as a message, but as normal roll as well. This also prevents modules like Dice So Nice from showing a roll as a double.

I have already prepared a pull request for this change with minor (tested) changes that keep backwards compatibility. It was a bit unclear what the purpose of the "evaluation" option was supposed to do, so I sorta tried to use this as the vehicle for this change. The power roll prompt always get "evaluate" as option and I've kept that, but that means some code can probably go. Or perhaps there's a reason for it, in which case I could look at implementing the desired behavior of that parameter.

PR #1595, opened 2026-02-04
feat: use evaluation option for rollCharacteristic

Mithror:
evaluation was previously set to "evaluate", provided to the prompt and never used again. Instead evaluation is now set to either the provided value or "message" if no value was provided. The prompt receives a hard-coded "evaluate" as before. When evaluation is set to "message", then the roll will be returned as a message, otherwise just as a roll.

Associated PR: #1594

This is a change required to support Monk's Tokenbar which I'm also preparing to submit once this PR has been accepted. You can see the changes here: https://github.com/Mithror/monks-tokenbar/commit/8b41cd0389f15bfcfe98c11fbba932eb5f3e403e

There are some improvements to the existing codebase that could be done. Specifically in the prompt and the code linked to the evaluation parameter, but as it was not clear what the intended functionality is supposed to be, I've let it as is. If instructed, I can clean up that part and only leave the code for the "evaluate" path.

Output of npm lint:

```
> draw-steel@1.0.0 lint
> eslint draw-steel.mjs src/module/**/* templates/**/*


/home/gael/git/draw-steel/src/module/applications/sidebar/apps/_types.d.ts
   8:3  warning  Sentences must end with a period  @jsdoc/require-description-complete-sentence
  26:3  warning  Sentences must end with a period  @jsdoc/require-description-complete-sentence
  36:3  warning  Sentences must end with a period  @jsdoc/require-description-complete-sentence
  38:3  warning  Sentences must end with a period  @jsdoc/require-description-complete-sentence

✖ 4 problems (0 errors, 4 warnings)
  0 errors and 4 warnings potentially fixable with the `--fix` option.
```
If necessary I can have this issues fixed, but as I did not touch these files I did not want to clutter the PR with unrelated changes.

### DIFF
--- a/src/module/data/actor/creature.mjs
+++ b/src/module/data/actor/creature.mjs
@@ -116,7 +116,6 @@ export default class CreatureModel extends BaseActorModel {
     const skills = this.hero?.skills ?? null;
     const skillModifiers = this.hero?.skillModifiers ?? null;
 
-    const evaluation = "evaluate";
     const formula = `2d10 + @${ds.CONFIG.characteristics[characteristic].rollKey}`;
     const data = this.parent.getRollData();
     const modifiers = {
@@ -127,9 +126,14 @@ export default class CreatureModel extends BaseActorModel {
 
     const doc = await fromUuid(options.resultSource);
 
+    // evaluation was previously set to "evaluate", provided to the prompt and never used again. 
+    // Instead evaluation is now set to either the provided value or "message" if no value was
+    // provided. The prompt receives a hard-coded "evaluate" as before. When evaluation is set
+    // to "message", then the roll will be returned as a message, otherwise just as a roll.
+    const evaluation = options.evaluation ?? "message";
     const promptValue = await PowerRoll.prompt({
       type,
-      evaluation,
+      evaluation: "evaluate",
       formula,
       data,
       modifiers,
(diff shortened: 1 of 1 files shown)


krbz999:
Is the main issue that this module requires returning a Roll?

Mithror:
> I think a simpler refactor would be to make it so if the power roll dialog did not receive targets, the modifiers selected during it get applied to our "base roll", which means that the single roll in the message root is fully correct with those edges and banes.

Forgive my ignorance, I'm new to the whole ecosystem of foundry, but what exactly do you mean with "if the power roll dialog did not receive targets". Just as an fyi: If the roll is initiated through Monk's TokenBar then the player who makes the roll still gets the dialog. For them it's as if they clicked on their ability.

> The other thing to consider is to just read the roll from the system.parts which is where all that actual roll data gets used. The base roll is primarily passed for DSN purposes.

Ah, interesting, I tried this by changing evaluation option to `"message"` (so it returns a message again) and then adding this line in Monk's TokenBar draw steel system support:
```js
                if (roll instanceof ChatMessage) {
                    let msg = roll;
                    roll = roll.system.parts.find(part => part.type === "test")?.rolls[0]
                    msg.delete();
                }
                else {
                    roll = Array.isArray(rolls) && rolls.length ? rolls[rolls.length-1] : rolls;
                }
```
That properly show the number with the modifiers. A downside is that keeps showing the message of course, so now I have to manually delete it myself as to not clutter the chat too much. So this technically works. If that's your preferred way of going forward, I can live with that.

If that's the case, perhaps I could just clean up the code with regards to the evaluation option? Or if you want to actually use this for something, I could look into the implementation?

Issue #1580, opened 2026-01-25
Implement Lookup enricher


JPMeehan:
A large variety of features like Revitalizing Ritual currently crudely implement their referenced values as rolls. We should implement the lookup enricher wherever possible, especially with fallback values, e.g. `equal to [[lookup @level]]{your level}`

cdverrett94:
So I'm starting to go through and find good opportunities to use this.  I've found an example that I think requires some changes to how the evaluate option of the lookup is handled.

There's some examples where its like do something for some rounds equal to might with a minumum of 1 round. The best lookup would be `[[lookup formula="max(@M, 1) evaluate=true]]{your might (minimum 1 round)}`, but evaluate just evaulates the formula even if the item is unowned. I would expect the fallback text to show if the formula is using actor data and there's no actor. 

Thoughts?

cdverrett94:
Here's my thought on a solution. Basically, pre replace formula data. If there's still roll data, use the fallback, otherwise continue evaluating the formula.

```
if (parsedConfig.evaluate) {
    const replacedFormula = Roll.replaceFormulaData(parsedConfig.formula, data);
    value = (replacedFormula.includes("@")) ? fallback : ds.utils.evaluateFormula(replacedFormula, data, { contextName: "lookup" });
}
```

JPMeehan:
Yeah that looks good to me

Issue #1050, opened 2025-09-08
Enrichers with long text don't wrap which causes issues in chat


cdverrett94:
### What happened?

Roll enrichers aren't wrapping. We have CSS in our files mimicking the core behavior which also doesn't wrap. 

<img width="297" height="387" alt="Image" src="https://github.com/user-attachments/assets/b2e66b28-8519-4cc6-a45a-ff702790c072" />

### Reproduction Steps

Share the Censor's Edict of Stillness to chat.

### Bug Checklist

- [x] The issue occurs while all Modules are disabled
- [ ] I am using The Forge for hosting

PR #1603, opened 2026-02-08
[#1580] Use lookup enricher

cdverrett94:
Related to #1580.  Not sure if I got all places where this would be useful.

Per discussion in that thread, I already made it so that if the roll data isn't present when using evaluate, it'll use the fallback instead.

I didn't use lookups where a damage formula would already take that value into account.

I also changed some damage enricher instances to make the label text shorter to mitigate #1050 and to make some of it more consistent in what's enriched.

NOTE: It may be good to add a non-enumerable `characteristic.highest` getter on the actor model and then to add that to the roll data. There's a quite a few places where to get the highest characteristic I'm setting the formula as `max(@M, @A, @R, @I, @P)` which is really cumbersome.


JPMeehan (approved):
Great work!

PR #1636, opened 2026-02-13
Fix/pr 1626

### LINKED ISSUE #1626: Concussive Slam forced movement
### Which compendiums is this in?

Ancestries & Backgrounds

### Documents

ability_Concussive_Slam_FRPORMdCeEONySxL.json

### What changes need to be made?

Time Raider granted ability Concussive Slam has forced movement only on tier 3 however the compendium item has also unnecessary:
"forced": {
            "tier1": {
              "distance": ""
            },
which messes with the Token Action HUD.

Mithror:
Fixes #1626

The issue mentions that only tier 3 has a push, but there is also a push at tier 2. Remove the push from tier 1.

Also fixed an issue with the push distance being displayed incorrectly.

And there was also a change in casing for some items, which causes pack/unpack to detect it as a change.


JPMeehan:
> I removed the tier 1 push from the compendium, but there is still an entry in the json file, don't know if that is the issue the were having? I just made it such that the text no longer displays the push. I suspect there's always a tier1 entry in a forced movement entry in the json?

Yeah, empty tiers are an expected part of this.

PR #1655, opened 2026-02-18
removed certain styles from root, tooltip

### LINKED ISSUE #1654: Fix light mode style bleed into tooltips
Image

Image

christopherhopper:


src/styles/variables/colors.css
```diff
@@ -1,4 +1,4 @@
-:root,
+/*:root,*/
```
krbz999, line 1: Will we be adding it back later or can we just remove it entirely?

christopherhopper:
I suspect we'll need it back for the chat and other dialog options, but
don't know enough about Foundry to say for certain.

JPMeehan (approved):
Tested, looks good.

PR #1662, opened 2026-02-20
Adjusted lookup tooltip

### LINKED ISSUE #1660: Lookup enricher creates "undefined" tooltip when no fallback
Image

Should probably just either not have a tooltip or only display formula or something.

JPMeehan:
Closes #1660

### DIFF
--- a/src/module/applications/ux/enrichers/lookup.mjs
+++ b/src/module/applications/ux/enrichers/lookup.mjs
@@ -66,7 +66,8 @@ export async function enricher(match, options) {
   if (!value && (options.documents === false)) return null;
   else if (!value) span.classList.add("not-found");
   span.innerText = value ?? parsedConfig.formula;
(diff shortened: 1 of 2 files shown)


src/module/applications/ux/enrichers/lookup.mjs
```diff
@@ -66,7 +66,8 @@ export async function enricher(match, options) {
   if (!value && (options.documents === false)) return null;
   else if (!value) span.classList.add("not-found");
   span.innerText = value ?? parsedConfig.formula;
-  span.dataset.tooltip = (value === fallback) ? parsedConfig.formula : fallback;
+  console.log(fallback, parsedConfig);
```
krbz999: Gezundheit

krbz999 (approved):

src/module/applications/ux/enrichers/lookup.mjs
JPMeehan: This is why we do review!

PR #1667, opened 2026-02-21
Add environmental hazard DTOs

cdverrett94:
Added the environmental hazard DTOs


JPMeehan (approved):
LGTM, obviously need to spend some time in v14 figuring out how to make these in particular work with regions.

PR #1673, opened 2026-02-23
Implement mvp party actor

krbz999:
- Adds `party` actor.
- ~~Deprecates `canvas.tokens.performTokenPlacement` in favor of `canvas.tokens.placeToken` to be in line with `canvas.regions.placeRegion`.~~
- Fixed issues with `canvas.tokens.#getTokenData` which threw errors when items or effect were empty.

### Notes

- I wasn't really able to find a good spot for using `canvas.tokens.#getTokenData` and don't think it's relevant for this anyhow. We are also not using it anywhere in the system at present.
- This is a proof of concept for now while we discuss whether we want to add it to 0.11.

### TODO
- [x] Implement basic party sheet.
- [x] Extend the Actor sidebar and implement `game.actors.party` as an easy access getter.
- [x] Add setting to assign "primary party" with context menu options in the sidebar.
- [x] Disallow creating (any?) items and effects on the party actor.

To test:

Create party actor and hero actor.

```js
const party = await Actor.implementation.create({ type: "party", name: "Party" });
const hero = await Actor.implementation.create({ type: "hero", name: "Hero" });
await party.system.addMembers([hero]);
await party.system.placeMembers();
```

### DIFF
--- a/src/module/data/actor/party.mjs
+++ b/src/module/data/actor/party.mjs
@@ -1,0 +1,133 @@
+/**
+ * @import DrawSteelActor from "../../documents/actor.mjs";
+ * @import DrawSteelTokenDocument from "../../documents/token.mjs";
+ */
+
+const { HTMLField, SchemaField, TypedObjectField } = foundry.data.fields;
+
+export default class PartyModel extends foundry.abstract.TypeDataModel {
+  /** @type {*} */
+  static get metadata() {
+    return {
+      type: "party",
+    };
+  }
+
+  /* -------------------------------------------------- */
+
+  /** @override */
+  static defineSchema() {
+    return {
+      description: new SchemaField({
+        value: new HTMLField(),
+      }),
+      members: new TypedObjectField(
+        new SchemaField({}),
+        { validateKey: key => foundry.data.validators.isValidId(key) },
+      ),
+    };
+  }
+
+  /* -------------------------------------------------- */
+
+  /** @inheritdoc */
+  static LOCALIZATION_PREFIXES = super.LOCALIZATION_PREFIXES.concat([
+    "DRAW_STEEL.Actor.party",
+  ]);
+
+  /* -------------------------------------------------- */
+
+  /** @inheritdoc */
+  async _preCreate(data, options, user) {
+    if ((await super._preCreate(data, options, user)) === false) return false;
+
+    const update = {};
+    if (!foundry.utils.hasProperty(data, "prototypeToken.actorLink"))
+      foundry.utils.setProperty(update, "prototypeToken.actorLink", true);
+
+    if (!foundry.utils.hasProperty(data, "prototypeToken.sight.enabled"))
+      foundry.utils.setProperty(update, "prototypeToken.sight.enabled", false);
+
+    if (!foundry.utils.hasProperty(data, "prototypeToken.disposition"))
+      foundry.utils.setProperty(update, "prototypeToken.disposition", CONST.TOKEN_DISPOSITIONS.FRIENDLY);
+
+    if (!foundry.utils.isEmpty(update)) this.parent.updateSource(update);
+  }
+
+  /* -------------------------------------------------- */
+
+  /** @inheritdoc */
+  prepareBaseData() {
+    super.prepareBaseData();
+
+    Object.defineProperty(this, "members", {
+      enumerable: true,
+      get() {
+        return Object.entries(this._source.members).reduce((acc, [id, data]) => {
+          const actor = game.actors.get(id);
+          if (this.validMember(actor)) acc.set(actor.id, { ...data, actor });
+          return acc;
+        }, new foundry.utils.Collection());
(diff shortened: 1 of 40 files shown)


JPMeehan:
1. RE: `performTokenPlacement`, we had previously discussed not wanting to do `placeToken` *because* it's more likely to be used by core
2. We already have `DrawSteelActors` as a subclass for `game.actors`, just need to add the access point
3. We've got `metadata.invalidActorTypes` on the item subclasses as us trying to more broadly accomodate module-added subtypes. I think that's probably sufficient?

src/module/data/actor/party.mjs
```diff
@@ -0,0 +1,133 @@
+/**
+ * @import DrawSteelActor from "../../documents/actor.mjs";
+ * @import DrawSteelTokenDocument from "../../documents/token.mjs";
+ */
+
+const { HTMLField, SchemaField, TypedObjectField } = foundry.data.fields;
+
+export default class PartyModel extends foundry.abstract.TypeDataModel {
```
JPMeehan: Should probably be `DrawSteelSystemModel`? I don't think it's *super* likely we end up with any pseudo collections but it does seem expected.

src/module/data/actor/party.mjs
```diff
@@ -0,0 +1,133 @@
...
+    if (!foundry.utils.hasProperty(data, "prototypeToken.actorLink"))
+      foundry.utils.setProperty(update, "prototypeToken.actorLink", true);
+
+    if (!foundry.utils.hasProperty(data, "prototypeToken.sight.enabled"))
+      foundry.utils.setProperty(update, "prototypeToken.sight.enabled", false);
+
+    if (!foundry.utils.hasProperty(data, "prototypeToken.disposition"))
+      foundry.utils.setProperty(update, "prototypeToken.disposition", CONST.TOKEN_DISPOSITIONS.FRIENDLY);
+
+    if (!foundry.utils.isEmpty(update)) this.parent.updateSource(update);
```
JPMeehan: I think `foundry.utils.mergeObject(defaultData, data, { insertKeys: false, insertValues: false })` gets us equivalent results while being more maintainable?

src/module/data/actor/party.mjs
```diff
@@ -0,0 +1,133 @@
...
+    Object.defineProperty(this, "members", {
+      enumerable: true,
+      get() {
+        return Object.entries(this._source.members).reduce((acc, [id, data]) => {
+          const actor = game.actors.get(id);
+          if (this.validMember(actor)) acc.set(actor.id, { ...data, actor });
+          return acc;
+        }, new foundry.utils.Collection());
+      },
+    });
```
JPMeehan, line 82: Do we want an a new field for this instead of overwriting the initialized TOF?

src/module/data/actor/party.mjs
```diff
@@ -0,0 +1,133 @@
...
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Is a given actor valid to be a member of this party?
+   * @param {DrawSteelActor} actor
+   * @returns {boolean}
+   */
+  validMember(actor) {
+    return (actor instanceof foundry.documents.Actor) && ["hero"].includes(actor.type)
```
JPMeehan: I wonder if the array should be some static property on the constructor

src/module/data/actor/party.mjs
```diff
@@ -0,0 +1,133 @@
...
+
+  /**
+   * Remove members from the party.
+   * @param {DrawSteelActor[]} [actors]    The actors to remove.
+   * @returns {Promise<DrawSteelActor>}    A promise that resolves to the updated party actor.
+   */
+  async removeMembers(actors = []) {
+    const update = {};
+    actors.forEach(actor => {
+      if (this.validMember(actor) && this.members.has(actor.id)) update[`-=${actor.id}`] = null;
```
JPMeehan: Why do we check if it's a valid member first? e.g. if I do a forced type update to turn a hero into an NPC I can no longer use this method to remove them.

src/module/data/actor/party.mjs
krbz999: I think it would be redundant.

src/module/data/settings/primary-party.mjs
```diff
@@ -0,0 +1,12 @@
+export default class PrimaryPartyModel extends foundry.abstract.DataModel {
```
JPMeehan, line 1: So the goal here is we can extend this model to add more functionality?

JPMeehan (approved):
Yeah this makes a lot of sense to me.

src/module/data/settings/primary-party.mjs
krbz999: Yes, but it also helps being a DataModel since it holds a ForeignDocumentField.

If I remember right, there was an initialization order issue, though this may have changed in recent times.

PR #1674, opened 2026-02-23
[#1446] Implement `toMessage` to leverage message parts

### LINKED ISSUE #1446: Override DSRoll#toMessage
The typical Roll.toMessage workflow macro-users leverage should use our Message Parts by default.

krbz999:
Implements `toMessage` on the base `DSRoll` class. Subclasses should define `PART_TYPE`.

Closes #1446.

### DIFF
--- a/src/module/rolls/base.mjs
+++ b/src/module/rolls/base.mjs
@@ -7,7 +7,7 @@ export default class DSRoll extends foundry.dice.Roll {
    * The type used for the `part` of a standard message.
    * @type {string}
    */
-  static PART_TYPE = "";
+  static PART_TYPE = "roll";
 
   /* -------------------------------------------------- */
 
@@ -20,6 +20,7 @@ export default class DSRoll extends foundry.dice.Roll {
 
     if (!this._evaluated) await this.evaluate({ allowInteractive: rollMode !== CONST.DICE_ROLL_MODES.BLIND });
 
+    const id = foundry.utils.randomID();
     messageData = foundry.utils.mergeObject({
       type: "standard",
       system: {},
@@ -27,7 +28,8 @@ export default class DSRoll extends foundry.dice.Roll {
       sound: CONFIG.sounds.dice,
     }, messageData);
     messageData.system.parts = {
-      [foundry.utils.randomID()]: {
+      [id]: {
+        _id: id,
         flavor: messageData.flavor,
         rolls: [this],
         type: this.constructor.PART_TYPE,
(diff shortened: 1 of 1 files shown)


src/module/rolls/base.mjs
```diff
@@ -2,4 +2,45 @@
...
+      system: {},
+      author: game.user.id,
+      sound: CONFIG.sounds.dice,
+    }, messageData);
+    messageData.system.parts = {
+      [foundry.utils.randomID()]: {
+        flavor: messageData.flavor,
+        rolls: [this],
+        type: this.constructor.PART_TYPE,
+      },
```
JPMeehan: Isn't _id required? I've just been leaning on our `migrateData` function to generate an ID and use it for the key and _id value

src/module/rolls/base.mjs
```diff
@@ -2,4 +2,45 @@
...
+
+    if (!this._evaluated) await this.evaluate({ allowInteractive: rollMode !== CONST.DICE_ROLL_MODES.BLIND });
+
+    messageData = foundry.utils.mergeObject({
+      type: "standard",
+      system: {},
+      author: game.user.id,
+      sound: CONFIG.sounds.dice,
+    }, messageData);
+    messageData.system.parts = {
```
JPMeehan, line 30: I suspect we will need to make it easier for subclasses to adjust this chunk of data.

src/module/rolls/base.mjs
```diff
@@ -2,4 +2,45 @@
 /**
  * Base roll class for Draw Steel.
  */
-export default class DSRoll extends foundry.dice.Roll { }
+export default class DSRoll extends foundry.dice.Roll {
+  /**
+   * The type used for the `part` of a standard message.
+   * @type {string}
+   */
+  static PART_TYPE = "";
```
JPMeehan: Shouldn't the base case be just creating a single `roll` part?

src/module/rolls/base.mjs
```diff
@@ -2,4 +2,47 @@
...
+    messageData.system.parts = {
+      [id]: {
+        _id: id,
+        flavor: messageData.flavor,
+        rolls: [this],
+        type: this.constructor.PART_TYPE,
+      },
+    };
+    delete messageData.flavor;
+    delete messageData.rolls;
```
JPMeehan: Prevents DSN from triggering, IMHO we just leave the `rolls` in

JPMeehan (approved):

PR #1675, opened 2026-02-23
Retainer Data Model

JPMeehan:
`155-retainer` will function as a feature branch for work on Retainers until we're ready to merge to /develop as part of 0.11 development.

This pull request represents work on the Retainer's data model alone. Rules for retainers can be found in the Monsters book on p. 351

### DIFF
--- a/src/module/data/actor/retainer.mjs
+++ b/src/module/data/actor/retainer.mjs
@@ -86,7 +86,7 @@ export default class RetainerModel extends CreatureModel {
   /* -------------------------------------------------- */
 
   /**
-   * Spend a recovery, adding to the hero's stamina and reducing the number of recoveries.
+   * Spend a recovery, adding to the retainer's stamina and reducing the number of recoveries.
    * @returns {Promise<DrawSteelActor>}
    */
   async spendRecovery() {
--- a/src/module/data/actor/retainer.mjs
+++ b/src/module/data/actor/retainer.mjs
@@ -1,0 +1,107 @@
+import CreatureModel from "./creature.mjs";
+import { requiredInteger, setOptions } from "../helpers.mjs";
+import SourceModel from "../models/source.mjs";
+
+export default class RetainerModel extends CreatureModel {
+  /** @inheritdoc */
+  static get metadata() {
+    return {
+      ...super.metadata,
+      type: "retainer",
+    };
+  }
+
+  /* -------------------------------------------------- */
+
+  /** @inheritdoc */
+  static LOCALIZATION_PREFIXES = super.LOCALIZATION_PREFIXES.concat([
+    "DRAW_STEEL.SOURCE",
+    "DRAW_STEEL.Actor.retainer",
+  ]);
+
+  /* -------------------------------------------------- */
+
+  /** @inheritdoc */
+  static defineSchema() {
+    const fields = foundry.data.fields;
+    const schema = super.defineSchema();
+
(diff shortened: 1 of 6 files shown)


src/module/data/actor/retainer.mjs
```diff
@@ -0,0 +1,107 @@
...
+
+    this.parent.updateSource({
+      prototypeToken: {
+        actorLink: true,
+        disposition: CONST.TOKEN_DISPOSITIONS.FRIENDLY,
+        sight: {
+          enabled: true,
+        },
+      },
+    });
```
krbz999, line 83: This would override explicit data if, say, one has good reason to want `prototypeToken.sight.enabled: false` and creates (or imports) an actor as such.

src/module/data/actor/retainer.mjs
```diff
@@ -0,0 +1,107 @@
...
+    if (this.recoveries.value === 0) {
+      ui.notifications.error("DRAW_STEEL.Actor.base.SpendRecovery.Notifications.NoRecoveries", {
+        format: { actor: this.parent.name },
+      });
+      return this.parent;
+    }
+
+    ui.notifications.success("DRAW_STEEL.Actor.base.SpendRecovery.Notifications.Success", {
+      format: { actor: this.parent.name },
+    });
```
krbz999, line 102: Generally not a fan of ui.notifications for general game-related notifications, would prefer a chat message.

But that's a convention we can work towards at a later date.

src/module/data/actor/retainer.mjs
```diff
@@ -0,0 +1,107 @@
...
+          enabled: true,
+        },
+      },
+    });
+  }
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Spend a recovery, adding to the hero's stamina and reducing the number of recoveries.
```
krbz999:
```suggestion
   * Spend a recovery, adding to the retainer's stamina and reducing the number of recoveries.
```

src/module/data/actor/retainer.mjs
```diff
@@ -0,0 +1,107 @@
...
+
+  /** @inheritdoc */
+  prepareDerivedData() {
+    super.prepareDerivedData();
+
+    // allows for stamina bonuses to apply first
+    this.recoveries.recoveryValue = Math.floor(this.stamina.max / this.recoveries.divisor) + this.recoveries.bonus;
+
+    // Winded is set in the base classes derived data, so this needs to run after
+    this.stamina.min = -this.stamina.winded;
```
krbz999, line 65: Do we need to adjust our data prep overall to handle stamina better for subclasses?

krbz999 (approved):
Looks fine. I don't think a deep dive from my end into the retainer rules is necessary just yet as we're comitting to a feature branch and it'll all come up eventually during reviews and testing.

src/module/data/actor/retainer.mjs
JPMeehan: Copied over from `HeroModel` — change there too? Maybe merge in createData or something here?

src/module/data/actor/retainer.mjs
JPMeehan: We can ticket separately since this is copied from HeroModel

src/module/data/actor/retainer.mjs
JPMeehan: It's the same comment as HeroModel; I think this is a fine-and-expected outcome of managing when and where to call `super`

PR #1676, opened 2026-02-23
Artisan and Sage data model, rudimentary sheet.

JPMeehan:
`679-artisan-sage` will function as a feature branch for work on Sages & Artisans until we're ready to merge to /develop as part of 0.11 development.

This pull request represents work on the Followers's data model. While working on it I ran into an important question about roll data structure — I *think* it's correct to have `@item.A` to reference the followers Agility, with the root `@A` being reserved for the owning hero.


templates/sheets/item/partials/follower.hbs
```diff
@@ -0,0 +1,7 @@
+<fieldset class="characteristics" {{disabled isPlay}}>
+  <legend>{{systemFields.characteristics.label}}</legend>
+  {{log this}}
```
krbz999: Gezundheit

src/module/data/item/follower.mjs
```diff
@@ -0,0 +1,97 @@
...
+ * Sages and Artisans are implemented as items because they are primarily associated with a hero and do not need to be represented on the canvas.
+ */
+export default class FollowerModel extends BaseItemModel {
+  /** @inheritdoc */
+  static get metadata() {
+    return {
+      ...super.metadata,
+      type: "follower",
+      packOnly: false,
+      invalidActorTypes: ["object", "npc"],
```
krbz999:
```suggestion
      invalidActorTypes: ["npc", "object"],
```

src/module/data/item/follower.mjs
```diff
@@ -0,0 +1,97 @@
...
+  modifyRollData(rollData) {
+    super.modifyRollData(rollData);
+
+    rollData.item.chr = -5;
+    for (const [key, obj] of Object.entries(this.characteristics)) {
+      const rollKey = ds.CONFIG.characteristics[key].rollKey;
+      rollData.item[rollKey] = obj.value;
+
+      if (obj.value > rollData.item.chr) rollData.item.chr = obj.value;
+    }
```
krbz999:
```suggestion
    const chars = Object.entries(this.characteristics).map(([k, v]) => {
      const rollKey = ds.CONFIG.characteristics[k].rollKey;
      rollData.item[rollKey] = v.value;
      return v.value;
    });
    rollData.item.chr = Math.max(...chars);
```

draw-steel.mjs
```diff
@@ -200,7 +200,7 @@ Hooks.once("i18nInit", () => {
   helpers.localization.performPreLocalization(CONFIG.DRAW_STEEL);
 
   // These fields are not auto-localized due to having a different location in en.json
-  for (const model of Object.values(CONFIG.Actor.dataModels)) {
+  for (const model of Object.values(CONFIG.Actor.dataModels).concat(CONFIG.Item.dataModels.follower)) {
```
krbz999, line 203: Let's rather just add them to a proper location in the i18n file and add that path to `LOCALIZATION_PREFIXES`.

src/module/data/item/follower.mjs
```diff
@@ -0,0 +1,97 @@
...
+    const characteristic = { initial: 0, integer: true, nullable: false };
+
+    schema.characteristics = new fields.SchemaField(
+      Object.entries(ds.CONFIG.characteristics).reduce((obj, [chc, { label, hint }]) => {
+        obj[chc] = new fields.SchemaField({
+          value: new fields.NumberField({ ...characteristic, label, hint }),
+        });
+        return obj;
+      }, {}),
+    );
```
krbz999, line 42: A `CharacteristicsField` could help here but might be overkill.

krbz999 (approved):

draw-steel.mjs
JPMeehan: Yeah I think we're at the point where we should just move the characteristics into CONST. Should I try to do that as part of this PR? Feels like I should just do it as a follow-up where this is a temp patch

src/module/data/item/follower.mjs
JPMeehan: I think I'll do that in the follow-up PR where I migrate into CONST

PR #1680, opened 2026-02-24
[#1364] Move skills to top level of system data

### LINKED ISSUE #1364: Consider migrating `skills` outside of `hero`
Currently, skills are under `hero.skills`, but many other potential actor types (Companion, Artisans, Sages) have access to skills.

JPMeehan:
1. We've got more general need to have it be `skills.value` so there can be the already-prepared `skills.list` and `skills.modifiers`
2. While normal NPCs don't have skills, there's enough other actor types like Companions that have skills. This also jives better with our expected Follower data.

### DIFF
--- a/src/module/data/actor/hero.mjs
+++ b/src/module/data/actor/hero.mjs
@@ -143,6 +143,7 @@ export default class HeroModel extends CreatureModel {
   prepareBaseData() {
     super.prepareBaseData();
 
+    // Existing DrawSteelActiveEffect#_applyAdd override means this also shims active effects targeting hero.skills
     HeroModel.shimSkills(this);
 
     this.combat.initiativeThreshold = 6;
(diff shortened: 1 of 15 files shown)


krbz999:
Good idea, but let's add shims to AEs with deprecation period until 1.0.0.

src/module/data/actor/hero.mjs
```diff
@@ -93,6 +95,25 @@ export default class HeroModel extends CreatureModel {
...
+
+    return super.migrateData(data);
+  }
+
+  /* -------------------------------------------------- */
+
+  /** @inheritdoc */
+  static shimData(data, options) {
+    foundry.abstract.Document._addDataFieldShim(data, "hero.skills", "skills.value", { since: "0.11", until: "1.0" });
+    foundry.abstract.Document._addDataFieldShim(data, "hero.skillModifiers", "skills.modifiers", { since: "0.11", until: "1.0" });
```
krbz999: Looks like there's a plural version too, `_addDataFieldShims`.

krbz999 (changes requested):
One code comment, and we would also need a shim for AEs.

I was also no able to pick a skill in the power roll dialog.

JPMeehan:
Added a permanent migration for our AE changes

krbz999 (approved):

PR #1685, opened 2026-02-25
Implement ctrl+click damage cap

### LINKED ISSUE #1353: Feature Request: Damage capped at minion stamina
### User Experience

From discord about how to improve damage application to minion squads with AoE damage abilities:

> You know how you have the functionality to deal half damage by holding shift while clicking 'Apply'? 
>
> Would it be possible to tell the system to only apply damage worth a minion's stamina (which the system already knows from the squad stamina calculations) when you'd hold down `CTRL` for example?
> 
> Something like
`if CTRL=true,`   
` If organisation=Minion, `. 
`Apply Minion stamina max`

Related to #662

nelizzy:
Caps to max actor stamina per actor.
Tooltip label is pretty long -- maybe someone else has a better way of phrasing it?

Closed #1353

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -2073,12 +2073,12 @@
         "Buttons": {
           "ApplyHeal": {
             "Label": "Regain {amount} {type}",
-            "Tooltip": "Shift+Click: Halve Heal",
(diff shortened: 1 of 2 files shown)


JPMeehan:
I think we're going to go forward with the adjustments in #1840 instead, thank you for this PR.

PR #1697, opened 2026-02-27
Add share button to item sheet header controls (#1505)

### LINKED ISSUE #1505: Add "Share" button to Item Sheet header controls
### User Experience

As a Game Master, I'd like the ability to link items directly to the chat using a "Share" button 

Image

Image

Image

As shown in the images above, it currently isn't possible to "Share" the details of an item directly from it's own form or from the sidebar "items" menu. The only way is to add the item to a Hero inventory first and _then_ share it. 

I currently have a work-around in the form of a hero sheet called "All Loot" and share it from there when items drop, but that's less than ideal and I'd appreciate being able to link things directly.

nelizzy:
Adds share button to item sheet controls.
Checks for toEmbed method for future proofing against other items that may not embed. Not sure if any plans for that, so maybe unnecessary.
Closes #1505

### DIFF
--- a/src/module/applications/sheets/item-sheet.mjs
+++ b/src/module/applications/sheets/item-sheet.mjs
@@ -543,22 +543,22 @@ export default class DrawSteelItemSheet extends DSDocumentSheet {
   /* -------------------------------------------------- */
 
   /**
-   * Whether this item can be updated from a compendium source.
+   * Whether item has been configured for embeds.
    *
    * @this DrawSteelItemSheet
    */
   static #canEmbed() {
-    return "toEmbed" in this.document;
+    return this.document.system.toEmbed !== foundry.abstract.TypeDataModel.prototype.toEmbed;
   }
 
   /* -------------------------------------------------- */
 
   /**
-   * Renders an embedded document's sheet in play or edit mode based on the document sheet view mode.
+   * Sends item as embed in chat.
    *
    * @this DrawSteelItemSheet
    * @param {PointerEvent} event   The originating click event.
-   * @param {HTMLElement} target   The capturing HTML element which defined a [data-action].
+   * @param {HTMLElement} target   The capturing HTML element.
    * @protected
    */
   static async #shareDoc(event, target) {
--- a/src/module/applications/sheets/item-sheet.mjs
+++ b/src/module/applications/sheets/item-sheet.mjs
@@ -548,6 +548,7 @@ export default class DrawSteelItemSheet extends DSDocumentSheet {
    * DrawSteel items extend toEmbed, returning data for use with enricher.
    *
    * @this DrawSteelItemSheet
+   * @returns {boolean}
    */
   static #canEmbed() {
     return this.document.system.toEmbed !== foundry.abstract.TypeDataModel.prototype.toEmbed;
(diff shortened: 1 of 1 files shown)


src/module/applications/sheets/item-sheet.mjs
```diff
@@ -536,6 +542,39 @@ export default class DrawSteelItemSheet extends DSDocumentSheet {
   /*   Actions                                          */
   /* -------------------------------------------------- */
 
+  /**
+   * Whether this item can be updated from a compendium source.
```
JPMeehan: Adjust comment

src/module/applications/sheets/item-sheet.mjs
```diff
@@ -536,6 +542,39 @@ export default class DrawSteelItemSheet extends DSDocumentSheet {
...
+   *
+   * @this DrawSteelItemSheet
+   */
+  static #canEmbed() {
+    return "toEmbed" in this.document;
```
JPMeehan: All documents and TypeDataModels have a `toEmbed` method. The catch is if `document.system.toEmbed` has been extended to return a non-null value. A better check would be if `this.document.system.toEmbed !== foundry.abstract.TypeDataModel.prototype.toEmbed`. This will cover any of the document types we're likely to have this action for (sometimes the Document might have implemented toEmbed directly without reference to the system model, e.g. roll tables, but we're not making sheets for those)

JPMeehan (commented):
Condition check needs improvement

nelizzy:
Wasn't sure how in depth comment should be. Could further extrapolate to explain what a configured embed means?

JPMeehan:
The default behavior of `TypeDataModel#toEmbed` is to just return null (which causes the embed to not process). Our various subclasses extend it when they actually implement the embed.

nelizzy:
I've expanded the comments and think I'm happy with where they're at now?

```
 * Whether item has been configured for use with Draw Steel embed enricher.
 * All Foundry documents have toEmbed by default, but return null (preventing enricher embed).
 * DrawSteel items extend toEmbed, returning data for use with enricher.
```
   
and

```
 * Sends item with embed enricher in chat.
```

JPMeehan:
Yeah, honestly didn't need the comment to be *that* in-depth. I think this PR is good to go but going to hold it back for 0.11 rather than stick it in 0.10.2 (0.11 isn't far away, it'll be a closing release for our v13 development)

src/module/applications/sheets/item-sheet.mjs
```diff
@@ -536,6 +542,41 @@ export default class DrawSteelItemSheet extends DSDocumentSheet {
   /*   Actions                                          */
   /* -------------------------------------------------- */
 
+  /**
+   * Whether item has been configured for use with Draw Steel embed enricher.
+   * All Foundry documents have toEmbed by default, but return null (preventing enricher embed).
+   * DrawSteel items extend toEmbed, returning data for use with enricher.
+   *
+   * @this DrawSteelItemSheet
+   */
```
krbz999, line 552:
```suggestion
   * @this DrawSteelItemSheet
   * @returns {boolean}
   */
```

krbz999 (approved):

Issue #1693, opened 2026-02-26
A way / option that would allow not displaing information about ablities in the Chat messages, but rather just damage roll and effects


DonDon-9097:
### User Experience

Hopefully as illustrated on the image bellow, Draw Steel module displays all of the information about the used ability inside the chat window, compared to, in this example, Pathfinder, which hides monster statistics.

<img width="1453" height="1585" alt="Image" src="https://github.com/user-attachments/assets/c15fe95d-52a2-44e8-ba52-7fe0e7d71178" />

The only way to circumvent that currently is making a private, blind or self roll, which in turn will obfuscate all information about the roll, including information that would be relevant and useful to the player such as for example the name of the NPC that made the roll, what kind of action main/maneuver and in general worsens the ability of the player of keeping track of the action looking through the chat history, since all the gm rolls will be just blank bubbles.
Additionally if the roll is set to private player cannot apply the dmg and effects to themselves through the chat window

**I would like to propose this kind of display of the message to the player:**

<img width="1261" height="981" alt="Image" src="https://github.com/user-attachments/assets/6f509679-0eda-40da-a8d3-e3bf01c562c5" />

In this way the player will receive all of the information that would be natural for his PC to understand as well as required to play the game

**Why would this be important?**
As an argument for a higher priority for this proposal I would like to appeal to the reason for which all of us gravitated to a game such as Draw Steel. Being an heroic battle adventure there is an inherent joy for the players to gradually understand an opponent they are fighting against over multiple turns, the extent of their abilities, surprises, tactics etc, and from the Game Masters side, there is an inherent satisfaction in both surprising the players as well as watching them adapt, discuss, strategize and re-strategize once they find out more about the monster they are facing over the course of combat.
Giving players too much information cheapens this experience, and giving them a private roll window:

<img width="294" height="130" alt="Image" src="https://github.com/user-attachments/assets/e639d712-2c54-4293-a409-c9f0bf346595" />

is unsatisfactory in it's own way

krbz999:
Something to consider as part of the rework of the ChatMessage/ChatLog visuals.

PR #1698, opened 2026-02-27
[#1692] Refactored message parts to improve usage of visibility options.

### LINKED ISSUE #1692: All abilities ( Main action , Maneuver Maliace abilities) are visible irrespectible to the private roll settings in 0.10
### What happened?

As stated above since 0.10 , all abilities when triggerd are displayed to every player even when set as a private, blind gm, self roll 
Image

Additionally even before 0.10 since maliace abilities don't have a roll in them,m there was no way to trigger them, as a private, blind gm, self roll and they were sent straight to chat for all the players to see

Image

### Reproduction Steps

> open a npc character sheet
> go to abilities , and for example Main action, slick any main action ability
> set the roll as a private, blind gm or self roll
> even though the colour in the chat indicates a blind roll the message will be displayed in it's entirety to players

FVTT 13 Draw steel 0.10.1

### Bug Checklist

JPMeehan:
I went with a general "favor hiding information if a message is marked as private" approach. I think #1693 is something we could explore after v14 with its refactor of the message visibility options.

### DIFF
--- a/src/module/data/pseudo-documents/message-parts/base-message-part.mjs
+++ b/src/module/data/pseudo-documents/message-parts/base-message-part.mjs
@@ -90,6 +90,18 @@ export default class BaseMessagePart extends TypedPseudoDocument {
   /* -------------------------------------------------- */
 
   /**
+   * Are inner pieces of this part visible to players?
+   * @type {boolean}
+   */
+  get isContentVisible() {
+    const whisper = this.message.whisper || [];
+    if (whisper.length) return whisper.includes(game.user.id) || (this.message.isAuthor && !this.message.blind);
+    return true;
+  }
+
+  /* -------------------------------------------------- */
+
+  /**
    * Modify the context used to render this part.
    * Called by StandardModel#_renderHTML.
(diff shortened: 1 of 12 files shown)


src/module/data/pseudo-documents/message-parts/base-message-part.mjs
```diff
@@ -89,6 +89,18 @@ export default class BaseMessagePart extends TypedPseudoDocument {
...
   /* -------------------------------------------------- */
 
+  /**
+   * Are inner pieces of this part visible to players?
+   * @type {boolean}
+   */
+  get isContentVisible() {
+    const whisper = this.message.whisper || [];
+    if (whisper.length) return whisper.includes(game.user.id) || (this.message.isAuthor && !this.message.blind);
+    return true;
```
krbz999:
```suggestion
    const { whisper, isAuthor, blind } = this.message;
    if (whisper?.length) return whisper.includes(game.user.id) || (isAuthor && !blind);
    return true;
```

krbz999 (approved):

PR #1706, opened 2026-03-04
Retainer Sheet

JPMeehan:
Adds a retainer sheet. Main TODO is how we want to store the level & handle advancements, but the rest of retainer stuff is covered.

image
image

image
image
image


krbz999 (approved):
I don't love that it basically looks like the Hero sheet, but what can we do.

JPMeehan:
All of our sheets look fairly similar. When we properly redo them we can better distinguish.

PR #1707, opened 2026-03-04
Rollup improvements

JPMeehan:
Removed need for double-registraiton of CSS files. Unlike https://github.com/krbz999/foundry-template/, we have multiple output CSS files, so a single rollup definition is insufficient; still, it's a net improvement to just have a static `_main.mjs`, `_css-elements.mjs`, and `_css-variables.mjs`. I'm open to renaming these files and/or adjusting their locations as might be preferred.


krbz999 (commented):
I don't think I see why we have two entry points; could those not be merged into one?

JPMeehan:
The CSS plugin appears to only be capable of generating one css file per entry in the exported roll-up array

krbz999:
> The CSS plugin appears to only be capable of generating one css file per entry in the exported roll-up array

I think we can have those two css files point to one common file, then?

JPMeehan:
So, the issue is that we have established three unique CSS files to be loaded by Foundry. We can probably cut one of them; I don't actually think that we're using the Typography stuff loaded by `elements`. But as far as I can tell, we need two entries in the `export default` array for rollup, which means two separate `_mjs` files to load.
```json
  "styles": [
    { "src": "css/draw-steel-system.css", "layer": "system" },
    { "src": "css/draw-steel-variables.css", "layer": "variables" },
    { "src": "css/draw-steel-elements.css", "layer": "elements" }
  ],
  ```

krbz999 (approved):

PR #1712, opened 2026-03-04
Follower Data Model and Sheet improvements

JPMeehan:
Implemented the rest of the data model & sheet. The expectation is that this is *not* for play purposes; the next PR will implement sheet display and relevant methods on the model.

image

Here's the Delian Tomb handout for comparison

image

### DIFF
--- a/src/module/constants.mjs
+++ b/src/module/constants.mjs
@@ -134,3 +134,17 @@ export const potencyStrengths = Object.freeze({
     glyph: "s",
   },
 });
+
+/* -------------------------------------------------- */
+
+/**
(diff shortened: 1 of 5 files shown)


src/module/constants.mjs
```diff
@@ -134,3 +134,17 @@ export const potencyStrengths = Object.freeze({
     glyph: "s",
   },
 });
+
+/* -------------------------------------------------- */
+
+/**
+ * Valid follower types for Follower items (excludes Retainer, which is an Actor subtype).
+ */
```
krbz999, line 143: missing type

krbz999 (approved):
Simple enough.

PR #1717, opened 2026-03-05
Follower Item

JPMeehan:
This covers all data and display related parts of implementing followers sufficient for an MVP. The remaining work is to build the API methods (and buttons to access them) for actually making project rolls.


src/module/data/item/follower.mjs
```diff
@@ -0,0 +1,179 @@
...
+  /** @inheritdoc */
+  static LOCALIZATION_PREFIXES = super.LOCALIZATION_PREFIXES.concat("DRAW_STEEL.Item.follower");
+
+  /* -------------------------------------------------- */
+
+  /** @inheritdoc */
+  static defineSchema() {
+    const schema = super.defineSchema();
+
+    schema.followerType = new fields.StringField({ required: true, choices: ds.CONST.followerTypes, initial: "artisan" });
```
krbz999: Not a fan of this name. Suggestions: "category" or "type" or "subtype".

src/module/constants.mjs
```diff
@@ -134,3 +134,18 @@ export const potencyStrengths = Object.freeze({
...
   },
 });
+
+/* -------------------------------------------------- */
+
+/**
+ * Valid follower types for Follower items (excludes Retainer, which is an Actor subtype).
+ * @type {Record<string, { label: string }>}
+ */
+export const followerTypes = Object.freeze({
```
krbz999, line 144: If it's just a label, this can go in the config just fine, and here in the constants we just enumerate the keys.

krbz999 (approved):
Some last minute suggestions.

src/module/data/item/follower.mjs
JPMeehan: So, keep label & just rename internal key to `type`?

krbz999: Seems good

PR #1724, opened 2026-03-06
Added mentor property to Retainer model

JPMeehan:
- Added Mentor as a FDF; mentors must be non-synthetic and not inside of a pack, which allows us to use it as `idOnly: false`
- Added retainers as valid party members

image

Styling problem: What to do with characters that have long names? Ellipsis, clip?
image
image

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -850,6 +850,9 @@
             },
             "role": {
               "label": "Creature Role"
+            },
+            "mentor": {
+              "label": "Mentor"
             }
           }
         },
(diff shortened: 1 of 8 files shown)


JPMeehan:
Relevant rules bits for mentors

> A retainer is a follower of a specific hero. This hero is called the retainer’s mentor, and their player controls that retainer in battle. Each player can control only one retainer at a time. If a hero ever acquires more than one retainer, all but one of those are considered to be working at the party’s stronghold or performing some function at the periphery of any battle, not contributing meaningfully to its outcome

> Part of a Mentor’s Turn: A retainer takes their turn as part of their mentor’s turn. For the purpose of effects that end at the end of the retainer’s turn, or any other rules elements that depend on the start or end of a creature’s turn, the start and end of the mentor’s turn is also the start and end of the retainer’s turn.

> Surges: Whenever a retainer gains surges, those surges are gained by their mentor, with the mentor and retainer sharing surges. The retainer spends surges the same way a hero can, but the extra damage granted to them by spending a surge is equal to their mentor’s highest characteristic score.

krbz999 (approved):

PR #1726, opened 2026-03-07
Retainer data updates

JPMeehan:
- Fixed Source handling (button & data prep)
- Added stub entries for all retainer abilities
- Added all retainers as base stats only
- Fully implemented The Nameless as proof of concept


krbz999 (approved):

krbz999:
Approved but would have preferred two PRs

Issue #1722, opened 2026-03-05
Implement Follower Rolls


JPMeehan:
Follow-up to #679 to add in API methods and click listeners.

<img width="659" height="393" alt="Image" src="https://github.com/user-attachments/assets/cf4ea442-b8e4-4d55-a210-bde1bd1c7527" />

<img width="679" height="373" alt="Image" src="https://github.com/user-attachments/assets/47589193-282b-47d3-91e0-70493346c23c" />

PR #1729, opened 2026-03-09
Follower roll API

JPMeehan:
Basic implementation of a follower roll API. We may want to break it up a bit if we want to enable non-project rolls from followers, but the primary purpose of Artisans and Sages is they make project rolls when the hero is away from home.

> An artisan can contribute one project roll per day to a downtime project you choose, whether you spend those days in respite, adventuring, or other activities. They must remain at your stronghold or at the site where the project is undertaken, and must have access to the necessary materials.

### DIFF
--- a/src/module/data/item/follower.mjs
+++ b/src/module/data/item/follower.mjs
@@ -191,9 +191,11 @@ export default class FollowerModel extends BaseItemModel {
   async roll(options = {}) {
     if (!this.actor) throw new Error("Only followers with a hero can ");
     if (!options.project) {
-      const projectOptions = this.actor.itemTypes.project
-        .filter(p => ds.CONST.followerTypes[this.type].projectTypes.has(p.system.type))
-        .map(p => ({ value: p.id, label: p.name }));
+      const projectOptions = this.actor.itemTypes.project.map(p => ({
+        value: p.id,
+        label: p.name,
+        group: ds.CONFIG.projects.types[p.system.type]?.label,
+      }));
 
       if (!projectOptions.length) {
         ui.notifications.error("DRAW_STEEL.Item.follower.ProjectChoice.NoProjects", { localize: true });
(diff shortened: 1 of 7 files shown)


JPMeehan:
If this API is approved I can expand it to add sheet functionality and close #1722 (I expect the sheet part of this to be relatively trivial)

src/module/data/item/follower.mjs
```diff
@@ -176,4 +180,56 @@ export default class FollowerModel extends BaseItemModel {
...
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Perform a characteristic roll to assist a project.
+   * @param {object} options
+   * @param {DrawSteelItem} [options.project] A specific project to roll for.
+   */
+  async roll(options = {}) {
+    if (!this.actor) throw new Error("Only followers with a hero can ");
```
krbz999, line 192: Can can-can?

src/module/data/item/follower.mjs
```diff
@@ -176,4 +180,56 @@ export default class FollowerModel extends BaseItemModel {
...
   }
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Perform a characteristic roll to assist a project.
+   * @param {object} options
+   * @param {DrawSteelItem} [options.project] A specific project to roll for.
+   */
+  async roll(options = {}) {
```
krbz999, line 191: Since this is a roll method, it should follow our to-be-established convention of rollconfig, dialogconfig, messageconfig. We aren't doing that in any of our other `roll` methods for Abilities and Projects, and I think we want to establish that *first*.

I also see we have `rollPrompt` in project items. We have to unify this before we tack on even more API.

krbz999 (commented):
The idea seems fine but I am of the opinion we hold off on this for 1.0.0 and rework the roll pipeline to get some fixed roll API.

JPMeehan:
I'm going to close this PR and start with a roll refactor *then* we can work followers in.

PR #1732, opened 2026-03-10
[#1725] Linked hero token resets to primary party

### LINKED ISSUE #1725: Manual Party Size
### User Experience

As a Game Master, I'd like to manually set the party size to better automate hero token resets. 

Ideally it would be using a party actor/folder and could count the heroes in there. But until that's implemented, maybe a bandaid solution would be to have a setting in the system settings? Could be a number input. Defaults to empty/0, where the behavior matches current behavior.

JPMeehan:
Closes #1725

### DIFF
--- a/src/module/data/settings/hero-tokens.mjs
+++ b/src/module/data/settings/hero-tokens.mjs
@@ -131,8 +131,14 @@ export class HeroTokenModel extends foundry.abstract.DataModel {
     /** @type {foundry.utils.Collection<string, {actor: DrawSteelActor}>} */
     const partyMembers = game.actors.party?.system.members;
 
-    // Use primary party if available, otherwise fall back to non GM users. Retainers do not count towards Hero Tokens.
-    const heroes = partyMembers?.filter(a => a.actor.type === "hero") ?? game.users.filter(u => !u.isGM);
+    if (!partyMembers) {
+      const msg = game.i18n.localize("DRAW_STEEL.Setting.HeroTokens.NoPrimaryParty");
+      ui.notifications.error(msg);
+      throw new Error(msg);
+    }
+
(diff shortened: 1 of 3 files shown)


src/module/data/settings/hero-tokens.mjs
```diff
@@ -124,12 +128,14 @@ export class HeroTokenModel extends foundry.abstract.DataModel {
       return;
     }
 
-    // TODO: Revisit after #369
+    /** @type {foundry.utils.Collection<string, {actor: DrawSteelActor}>} */
+    const partyMembers = game.actors.party?.system.members;
 
-    const nonGM = game.users.filter(u => !u.isGM);
-    await game.settings.set(systemID, "heroTokens", { value: nonGM.length });
+    // Use primary party if available, otherwise fall back to non GM users. Retainers do not count towards Hero Tokens.
```
krbz999: Disagree. We should enforce using the party.

JPMeehan: Should we revise our "fresh world" migration setup to create a party and flag it as the primary?

JPMeehan: And should our 0.11 migration create a primary party and even add in all current heroes to it?

krbz999: We dont have to be forceful, but features can require a party actor; just letting the user know that a feature requires a Party actor should be enough I think?

We can discuss whether it should wait for 1.0.0 though.

JPMeehan: I don't think this has to wait for 1.0, I'll refactor this to throw an error if there's no primary party and then a separate PR for the setup

krbz999 (approved):

PR #1733, opened 2026-03-10
[#637] Added surge option to Modify Damage dialog

### LINKED ISSUE #637: Implement Surges usage
Currently surges can only be tracked, but not used.

Implement that a player can click somewhere on the sheet relating to surges (not `Surges` itself, as that prompts hero token usage to gain surges) which then provides an `Apply Damage` button in chat based on highest characteristic of hero.

JPMeehan:
Previous API and UI work made this a very simple thing to add in.

image

Closes #637


krbz999 (approved):
Seems fine. Would look a little better with the `slim` class.

PR #1734, opened 2026-03-10
Primary Party initialization

JPMeehan:
- Added check for renaming logic on Gamemaster => Director, in anticipation of v14 start screen changes
- Migration now creates a primary party in new games and with 0.11 migration

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -837,8 +837,7 @@
         },
         "placeMembers": "Place Members",
         "stamina": "Stamina",
-        "recoveries": "Recoveries",
-        "primary": "Primary Party"
+        "recoveries": "Recoveries"
       },
       "retainer": {
         "FIELDS": {
--- a/lang/en.json
+++ b/lang/en.json
@@ -837,7 +837,8 @@
         },
         "placeMembers": "Place Members",
         "stamina": "Stamina",
-        "recoveries": "Recoveries"
+        "recoveries": "Recoveries",
+        "primary": "Primary Party"
       },
       "retainer": {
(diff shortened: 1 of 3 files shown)


krbz999 (changes requested):
I will disagree with creating a party actor for users, and I think we should not-so-gently nudge them to do so manually.

- What if they _wanted to_ create one manually?
- What if they had one stored in a pack that they reuse?
- What if an adventure imports one?
- What if that new-fangled world setup sets up a party actor?

Alternative: If `!game.actors.party`, and an actor of type `party` is created, set this actor as the primary party automatically.

JPMeehan:
- What are the benefits of manual party creation?
- What can party actors do from a pack? Everything is built around world IDs
- I think the adventure case is probably the strongest argument against; _The Condemned_ is in playtesting as an adventure that is explicitly built around pregens, you don't bring normal characters to it. At the same time, you don't actually know which characters your players would pick (You'd probably just pack the pregens in a compendium) so I'm not sure what a prepackaged party does for you
- I think auto-designating a new party as the primary, if none is set, is definitely something we should do.

krbz999:
If you create a party actor yourself, you know it exists and you know why you created it; users tend to forget or overlook things they did not tend to themselves.

A party actor in a pack, I was thinking of users who reuse the same party across worlds (or clear out a world and use a backup from a pack).

If an adventure would want to ship with one, we now have to work around this (or the adventure does). This is already more complicated than just gently nudging users to create a party manually.

krbz999 (approved):

PR #1735, opened 2026-03-10
Sort party members by name

JPMeehan:
- Renamed files to be `party-sheet`
- Rather than insertion based display, always sort alphabetically. We can expand with filters in the future.


krbz999 (commented):
Hot takes:
- Sort by `_source.name`.
- The sheet files are already in `/sheets/`, so should the convention not go the other way?

JPMeehan:
We definitely should commit one way or the other; currently party and npc don't use `-sheet`. To me, one reason to prefer `hero-sheet.mjs` is the vscode headers and search; I found myself doing `ctrl+P hero` and having a harder time parsing which file I was opening. How do you usually navigate repositories?

krbz999 (approved):
I tend to navigate using the filepath if there are name colissions.

I don't feel strongly though.

PR #1738, opened 2026-03-11
Add `MembersCollection` for utility, clean up HeroTokens

krbz999:


src/module/utils/members-collection.mjs
```diff
@@ -0,0 +1,24 @@
+/**
+ * Simplistic extension of Collection to allow splitting contents by type.
+ */
+export default class MembersCollection extends foundry.utils.Collection {
```
JPMeehan, line 5: We can use `@extends` in the implementation to note the contained values, typing this from memory so you may want to adjust
```suggestion
/**
 * Simplistic extension of Collection to allow splitting contents by type.
 * @extends {foundry.utils.Collection<string, { actor: DrawSteelActor }>}
 */
export default class MembersCollection extends foundry.utils.Collection {
```

src/module/data/actor/_types.d.ts
```diff
@@ -198,6 +198,12 @@ declare module "./object.mjs" {
   }
 }
 
+declare module "./party.mjs" {
+  export default interface PartyModel {
+    members: Map<string, DrawSteelActor>
```
JPMeehan: It's not this though, right? At a minimum it's `{ actor: DrawSteelActor }` but also this is an instance of MembersCollection, which if you use `@extends` in the typedef to make it so this *only* needs to be typed as `members: MembersCollection`

JPMeehan (commented):
We can improve the typing here, I can elaborate if you'd like

JPMeehan (approved):

PR #1747, opened 2026-03-13
Fix blink not being properly deleted. Add DataFieldOperator types.

cdverrett94:
Blink wasn't properly being deleted from the movement actions, so it was erroring out. This prevented loading in V14.

Added the typings for the `DataFieldOperator` shorthands, `_del` and `_replace`. Pretty that's correct.

### DIFF
--- a/draw-steel.d.ts
+++ b/draw-steel.d.ts
@@ -24,4 +24,12 @@ declare global {
    * The singleton game canvas.
    */
   const canvas: Canvas;
+  /**
+   * A singleton ForcedDeletion operator instance that can be reused.
+   */
+  const _del: foundry.data.operators.ForcedDeletion;
+  /**
+   * A reference for ForcedReplacement.create that can be easily referenced.
+   */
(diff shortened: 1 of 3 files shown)


draw-steel.d.ts
```diff
@@ -24,4 +24,12 @@ declare global {
...
    */
   const canvas: Canvas;
+  /**
+   * A singleton ForcedDeletion operator instance that can be reused.
+   */
+  const _del: foundry.data.operators.ForcedDeletion;
+  /**
+   * A reference for ForcedReplacement.create that can be easily referenced.
+   */
+  const _replace = foundry.data.operators.ForcedReplacement.create;
```
krbz999: Were these not added to the foundry global.d.ts in 14.356? That's disappointing.

cdverrett94: Oh it looks like they are. I just needed to import the file that has them

krbz999 (approved):

Issue #1166, opened 2025-09-24
Adjust ActiveEffect customizations for v14 structure


JPMeehan:
- Ensure our ActiveEffectConfig extension conforms to new data structures
- Duration is receiving a major refactor which clashes with our current handling
- CONFIG.statusEffects has been reworked from an Array to a Record

Relevant core tickets
- https://github.com/foundryvtt/foundryvtt/issues/13122
- https://github.com/foundryvtt/foundryvtt/issues/13332
- https://github.com/foundryvtt/foundryvtt/issues/13426
- https://github.com/foundryvtt/foundryvtt/issues/13667

PR #1760, opened 2026-03-16
[#1166] Refactored usage of CONFIG.statusEffects

### LINKED ISSUE #1256: Remove non-functional "default" conditions
With the implementation of Active Effect compendiums in v14, there's no need to provide statuses that don't have a specific game meaning.

Image

In particular, Burning, Frozen, Marked, and Targeted all stick out as statuses that should be removed since they falsely imply conditions that the system doesn't provide a definition for.

Of the rest of the conditions, I believe the following serve a purpose
- Dead
- Asleep
- Blind
- Burrowing
- Deaf
- Flying
- Invisible

These all represent clear physical conditions that Directors should have access to, especially since many tie into Foundry's automation or have natural language rules meaning.

JPMeehan:
image

### DIFF
--- a/src/module/helpers/localization.mjs
+++ b/src/module/helpers/localization.mjs
@@ -59,11 +59,8 @@ export function performPreLocalization(config) {
     if (settings.sort) foundry.utils.setProperty(config, keyPath, sortObjectEntries(target, settings.keys[0]));
   }
 
-  // Localize & sort status effects
-  CONFIG.statusEffects.forEach(s => s.name = _loc(s.name));
-  CONFIG.statusEffects.sort((lhs, rhs) =>
-    lhs.id === "dead" ? -1 : rhs.id === "dead" ? 1 : lhs.name.localeCompare(rhs.name, game.i18n.lang),
-  );
+  // Localize status effects
+  Object.values(CONFIG.statusEffects).forEach(s => s.name = _loc(s.name));
 }
 
 /* -------------------------------------------------- */
--- a/src/module/helpers/localization.mjs
+++ b/src/module/helpers/localization.mjs
@@ -59,7 +59,7 @@ export function performPreLocalization(config) {
     if (settings.sort) foundry.utils.setProperty(config, keyPath, sortObjectEntries(target, settings.keys[0]));
   }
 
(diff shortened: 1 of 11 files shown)


src/module/helpers/localization.mjs
```diff
@@ -59,11 +59,8 @@ export function performPreLocalization(config) {
     if (settings.sort) foundry.utils.setProperty(config, keyPath, sortObjectEntries(target, settings.keys[0]));
   }
 
-  // Localize & sort status effects
-  CONFIG.statusEffects.forEach(s => s.name = _loc(s.name));
-  CONFIG.statusEffects.sort((lhs, rhs) =>
-    lhs.id === "dead" ? -1 : rhs.id === "dead" ? 1 : lhs.name.localeCompare(rhs.name, game.i18n.lang),
-  );
+  // Localize status effects
+  Object.values(CONFIG.statusEffects).forEach(s => s.name = _loc(s.name));
```
krbz999, line 63: I think I need a reminder why we localize these; conditions are localized as soon as they're created via `Actor#toggleStatusEffects`.

Is it because of all the areas in which we use `CONFIG.statusEffects` directly? In that case, perhaps we should refer to the `ds.config` object instead, which makes more sense to prelocalize.

It's fine that this isn't part of this PR. Perhaps we can do that as part of moving more data into `ds.CONST`.

src/styles/system/applications/canvas/token-hud.css
```diff
@@ -2,6 +2,9 @@
   .status-effects.palette {
     display: flex;
     flex-direction: column;
+    /* Undoes v14 status palette changes */
+    overflow: inherit;
+    height: revert;
```
krbz999, line 7: Add a link to https://github.com/foundryvtt/foundryvtt/issues/13912

krbz999 (approved):

src/styles/system/applications/canvas/token-hud.css
JPMeehan: Relevant ticket is https://github.com/foundryvtt/foundryvtt/issues/3175 but will do

PR #1761, opened 2026-03-16
Update AE change `mode` to use new `type`

cdverrett94:
Also updated the wiki verbiage to use the new language.


JPMeehan (approved):
LGTM, we'll want to revisit all the docs in stable to make sure nothing's changed on us at the last minute

PR #1762, opened 2026-03-16
[#1501] Use `HTMLFormulaInputElement` for formula fields

### LINKED ISSUE #1501: Extend toInput in FormulaField to use HTMLFormulaInputElement
See https://github.com/foundryvtt/foundryvtt/issues/13696

cdverrett94:
Closes #1501

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ This version of the system does not support v13, it is exclusively for Foundry v
 
 ### Added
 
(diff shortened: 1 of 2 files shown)


JPMeehan (approved):
Well that's pleasantly simple. I expect us to need to do more here going forward.

cdverrett94:
Yeah, I think the autocomplete stuff is the next bit to get working with this, and that should be as simple as registering entries into `CONFIG.formulaEditor.contexts`. I was looking into how to get that setup, so may take a stab at it soon.

PR #1775, opened 2026-03-20
Adjusted actor preCreate logic

JPMeehan:
- updateSource already checks if the update is an empty object
- applied our mergeObject fix to retainers


krbz999 (approved):
Pushed some additional changes in the same vein.

PR #1779, opened 2026-03-22
[#1528] Added token AE support

### LINKED ISSUE #1528: Implement token size changes via AE
See https://github.com/foundryvtt/foundryvtt/issues/13461

JPMeehan:
Current behavior is that applying the AE will adjust size, but removing/disabling the AE will not undo that change. I filed this as a core bug. I've also opted to avoid trying to calculate wall nonsense; the typical size 1 to size 2 adjustment will end up with the token maintaining its center, which users can adjust afterwards.

Closes #1528


JPMeehan:
From Shark

> If you update a token's size from an active effect and then disable or remove that effect, the unchanged value is now what you updated it to. This is why core doesn't take care of updating for you, since there's information loss with a source update that wouldn't jive well with some systems.
> Obviously the underlying implementation is different in pf2e, but the workflow there is along the same lines except the system doesn't care what a token's "real" dimensions are: there's a low-priority override to set the dimensions according to an actor's size category, and then an effect can change that. When the effect is removed, there remains the underlying basic override to inform the token how big it should be.

src/docs/Active-Effects.md
```diff
@@ -231,6 +231,28 @@ Draw Steel also supports modifying the default roll formula for tests. `system.c
...
+| Disposition* |`token.disposition`|
+| Use dynamic token ring |`token.ring.enabled`|
+| Image (Dynamic Token Ring) |`token.ring.subject.texture`|
+| Turn marker image |`token.turnMarker`|
+| Movement Action** |`token.movementAction`|
+
+*Token dispositions are -2 for Secret, -1 for Hostile, 0 for Neutral, and 1 for Friendly.
+
+**Movement actions include "climb", "fly", "swim", "teleport", and "walk"
+
```
krbz999, line 247: Hot take, but I think we should just list the common ones.


```suggestion
Token properties can be targeted by prefixing with `token.`. The following properties are some of the valid targets for these changes.

|Property|Attribute Key|
|:---:|---|
| Name |`token.name`|
| Size (x-axis) |`token.width`|
| Size (y-axis) |`token.height`|
| Size (z-axis) |`token.depth`|
| Image (Static) |`token.texture.src`|
| Use dynamic token ring |`token.ring.enabled`|
| Image (Dynamic Token Ring) |`token.ring.subject.texture`|

```

krbz999 (approved):
So to make sure I understand the premise:
- an effect is applied which will perform an update to the token dimensions.
- removing the effect will not revert this change.

I understand why that's a limitation, I suppose. I might say this is a rare case where pf2e has the right idea of having the actor control the token size. Perhaps we could do similar, but use the prototype token properties.

Food for thought, I suppose. I don't feel strongly about it at all and it certainly doesn't need to hold up this PR. But I think people will expect to be able to undo a change in size.

JPMeehan:
Since we have an async method here, we could work in a flags update or something. I think there's room to iterate and improve for sure.

Issue #662, opened 2025-06-28
implement indicator if a stamina pool of a squad has been implemented enough to kill a minion


KJTailor:
Currently squad stamina pools are only displayed as `56/56 (7)` for example.

It would be really nice if they would include an extra value that displays his many minions are still in a squad.

So for example the above could be displayed as `56/56 (7; 8/8)` or a different one as `18/24 (3; 6/8)`

The exact format of that indicator is up for debate of course.

JPMeehan:
Honestly the post-backer packet rules made this a lot harder because now you have to work through the various limitations on damage, but some way to cleanly remove a minion from the squad & reduce the HP pool would be good. Maybe defeated combatants shouldn't count towards it?

KJTailor:
> Maybe defeated combatants shouldn't count towards it?

How do you mean? The problem is also, because all minions have stamina pool HP, it is hard to discern which minion has how much stamina (for the user I mean)

JPMeehan:
Minions don't individually lose stamina. My point is that ideally defeated minions no longer count towards the maximum stamina.

snyckling-cell:
Since at least the current implementation is that you have to select a token to deal damage, it feels most convenient to just deal damage to the selected minion first, then prompt to select more minions to deal further damage, dealing the remaining damage to selected minions until there is no more damage to be dealt.. Then remove stamina from the stamina pool of the squad accordingly.

nelizzy:
The implementation I have at my table is more verbose.
It ties minion count to the stamina thresholds of the squad. Currently doesn't decrease when marking a minion as defeated, but I have plans to integrate that.

Picture below:
<img width="281" height="146" alt="Image" src="https://github.com/user-attachments/assets/8d40ee4b-82a7-43f2-85d3-7e667dfb4253" />

nelizzy:
Would a message being sent when a squad takes damage saying something like "squad takes x damage, this kills x minions" also help?

cdverrett94:
Comment for assignment

PR #1783, opened 2026-03-24
Add a defeated minion selection dialog

cdverrett94:
I think this would be a good solution for #662. Forgetting to mark minions as defeated is a problem I've consistently had, so a reminder/prompt would be a big boon in my games.

When a minion squad is damaged, if the threshold to kill a minion is reached, a dialog will spawn for the active owner (prioritizing player if applicable). It will have a list of non-defeated minions to select, limiting to the number of minions that need to be defeated. Possible improvement to this would be indicated which of the minions are currently controlled to easily indicate which ones were selected when damaged.

Demo:
https://github.com/user-attachments/assets/68d61435-1aed-4279-91be-fe951d1eb2ae

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ This version of the system does not support v13, it is exclusively for Foundry v
 - New Player-Facing Compendium Content:
 - New Director-Facing Compendium Content:
   - Added common effects like Burning, Marked, Petrified, and Wet. (#552, #1254)
+- Added a dialog for marking minions as defeated once a squads stamina thresholds have been met. (#662)
 - Created an "Effects and Imbues" compendium pack. (#1163)
 - Added Draw Steel inserts to the "Custom" section of prosemirror, allowing users to easily format their text with system-specific html. (#1267)
 - Updated formula fields to use the new core `HTMLFormulaInputElement`. (#1501)
--- a/src/module/data/combatant-group/squad.mjs
+++ b/src/module/data/combatant-group/squad.mjs
@@ -90,7 +90,10 @@ export default class SquadModel extends BaseCombatantGroupModel {
   _onUpdate(changed, options, userId) {
     super._onUpdate(changed, options, userId);
 
-    if (changed.system && ("staminaValue" in changed.system)) this.refreshSquad();
+    if (changed.system && ("staminaValue" in changed.system)) {
+      this.refreshSquad();
+      this.checkDefeatedMinions();
+    }
     if (options.ds?.staminaDiff) this.displayMinionStaminaChange(options.ds.staminaDiff, options.ds.damageType);
   }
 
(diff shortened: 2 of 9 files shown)


src/module/data/combatant-group/squad.mjs
```diff
@@ -116,4 +119,31 @@ export default class SquadModel extends BaseCombatantGroupModel {
...
+  /* -------------------------------------------------- */
+
+  /**
+   * Determine if any minions should be defeated based on stamina value and threshold and prompt the active owner to mark them as defeated.
+   */
+  async checkDefeatedMinions() {
+    const minions = this.minions;
+    if (!minions.size) return;
+
+    const activePlayerOwner = game.users.find(user => !user.isGM && user.active && this.parent.testUserPermission(user, "OWNER") && this.minions.every(minion => minion.testUserPermission(user, "OWNER")));
```
JPMeehan:
```suggestion
    const activePlayerOwner = game.users.find(user => !user.isGM && user.active && this.parent.testUserPermission(user, "OWNER") && minions.every(minion => minion.testUserPermission(user, "OWNER")));
```

Already composed the set

src/module/data/combatant-group/squad.mjs
```diff
@@ -116,4 +119,31 @@ export default class SquadModel extends BaseCombatantGroupModel {
...
+    const shouldBeDefeated = Math.clamp(minions.size - Math.ceil(this.staminaValue / staminaPerMinion), 0, minions.size);
+    const needToDefeat = shouldBeDefeated - defeated.length;
+    // Only prompt if the number of minions that need to be defeated exceeds 0.
+    if (needToDefeat <= 0) return;
+
+    const fd = await ds.applications.apps.DefeatedMinionSelection.create({ context: { undefeated, needToDefeat, combat: this.combat, squad: this.parent } });
+    if (!fd || !fd.selectedMinions.length) return;
+
+    const selectedMinions = fd.selectedMinions.map(id => this.combat.combatants.get(id));
+    for (const minion of selectedMinions) ui.combat._onToggleDefeatedStatus(minion);
```
JPMeehan: We can bundle this update better than deferring to a protected combat tracker function.

JPMeehan (commented):
Looks really good, just some nitpicks

src/module/data/combatant-group/squad.mjs
cdverrett94: I thought to use that one because it handled both the combatant change and the actor effect creation. I've updated it to batch the combatant changes together, and just call `toggleStatusEffect` myself.

cdverrett94:
I've made the changes per review. I also changed the hbs to use a `label` for the minion name along with the `for` property so that clicking the name will toggle the checkbox.

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ This version of the system does not support v13, it is exclusively for Foundry v
 
 ### Added
 
+- Added a dialog for marking minions as defeated once a squads stamina thresholds have been met. (#662)
```
krbz999:
```suggestion
- Added a dialog for marking minions as defeated once a squad's stamina thresholds have been met. (#662)
```

src/module/data/combatant-group/squad.mjs
```diff
@@ -116,4 +119,36 @@ export default class SquadModel extends BaseCombatantGroupModel {
...
+  /**
+   * Determine if any minions should be defeated based on stamina value and threshold and prompt the active owner to mark them as defeated.
+   */
+  async checkDefeatedMinions() {
+    const minions = this.minions;
+    if (!minions.size) return;
+
+    const activePlayerOwner = game.users.find(user => !user.isGM && user.active && this.parent.testUserPermission(user, "OWNER") && minions.every(minion => minion.testUserPermission(user, "OWNER")));
+    const promptedUser = activePlayerOwner ?? game.users.activeGM;
+    if (game.user !== promptedUser) return;
```
krbz999:
```suggestion
    if (!promptedUser?.isSelf) return;
```

JPMeehan (approved):
I'm happy with this now, will merge after cleanup

PR #1785, opened 2026-03-24
[#1161] Define custom elements in detached windows; swap uses of `document` to `HTMLElement#ownerDocument`

### LINKED ISSUE #1161: Validate that all applications and hooks still work with detached windows
For v14

Reference: https://github.com/foundryvtt/foundryvtt/issues/230

krbz999:
Closes #1161.

Cleaned up the main file a bit as well. Unlike mentioned in 1161, we do register the element(s) but arent using them anywhere yet.

I applied a liberal use of `HTMLElement#ownerDocument` where an element could be found, and probably much more than needed.

### DIFF
--- a/draw-steel.mjs
+++ b/draw-steel.mjs
@@ -10,29 +10,35 @@ import * as rolls from "./src/module/rolls/_module.mjs";
 import * as utils from "./src/module/utils/_module.mjs";
 
 globalThis.ds = {
+  applications,
   canvas,
   compatibility,
+  data,
   documents,
-  applications,
   helpers,
   rolls,
-  data,
   utils,
-  CONST: DS_CONST,
   CONFIG: DS_CONFIG,
+  CONST: DS_CONST,
   registry: new helpers.DrawSteelRegistry(),
 };
 
-// Register custom elements.
-for (const element of Object.values(applications.elements)) {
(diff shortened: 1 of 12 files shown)


draw-steel.mjs
```diff
@@ -10,30 +10,36 @@ import * as rolls from "./src/module/rolls/_module.mjs";
...
   game.system.socketHandler = new helpers.DrawSteelSocketHandler();
   helpers.DrawSteelSettingsHandler.registerSettings();
   applications.apps.DocumentSourceInput.addModuleSources();
 
+  // Define custom elements.
+  const defineElements = window => {
+    window.customElements.define(applications.elements.DSIconElement.tagName, applications.elements.DSIconElement);
+  };
+  defineElements(window);
+  Hooks.on("openDetachedWindow", (id, window) => defineElements(window));
```
JPMeehan, line 41: I suppose it doesn't really matter but is there a reason we need to scope this inside the `init` callback? esm means that a top level `const` doesn't turn into a global variable.

JPMeehan (approved):

draw-steel.mjs
krbz999: Not at all. It's a stylistic choice on my part.

PR #1786, opened 2026-03-24
Replace additional uses of "-=" and "==" operators

krbz999:


src/module/data/migrations.mjs
```diff
@@ -176,8 +176,8 @@ export async function migrateType(collection, options = {}) {
 export async function migrateChanges(parentDocument) {
   const toMigrate = parentDocument.effects.filter(effect => effect.getFlag(systemID, "migrateChanges")).map(doc => ({
     _id: doc.id,
-    "==changes": [...doc.changes],
-    "flags.draw-steel.-=migrateChanges": null,
+    changes: _replace([...doc.changes]),
+    "flags.draw-steel.migrateChanges": _del,
```
JPMeehan: I addressed this in my duration PR but `changes` got moved to be under `system`

JPMeehan (approved):
I did the migration changes in my Duration PR since I had to update migrations anyways, maybe just skip that file?

PR #1791, opened 2026-03-27
[#161] Ability Configuration Dialog

### LINKED ISSUE #161: Create configuration dialog class for Ability Usage
1. Better for modules because distinct hooks
2. More solid than just a basic dialog

Should extend AppV2, take the item being used in the options passed in.

### LINKED ISSUE #550: Implement special HR conditions for Shadow
Shadow abilities cost 1 less HR if the ability has a strike. That should be a option when setting up the class.

JPMeehan:
Implemented a new Ability Configuration Dialog, pulling a large amount of the current Power Roll logic out of that class and into just abilities

image

image

image

image

image

Closes #161

Closes #550


krbz999 (changes requested):
The application feels too clunky to use.

- Are all the changes in this PR required for this new application?
- It feels frustrating to adjust the Resource Cost field or any of the other inputs as focus is lost with each change. This must be due to an issue with how the application gets re-re-rendered, but I was not able to locate it. I made a commit that added a proper root id and ids to the form but this did not change the behavior.
- At a glance, the "targets" section looks unintuitive. Take the below example, where it isn't clear who's a target and who's a formula.
- We should look into hooking onto token targeting to re-render the application, one shouldn't need to target tokens prior to rendering the application.

Image

JPMeehan:
1. Broadly yes, the PR  is factoring *out* a lot of logic that already existed
2. Will investigate
3. I think a discord suggestion of a "pan to target" would help
4. I agree a more dynamic targets logic could work if we make sure the ordering is consistent

JPMeehan:
I think this adds both a substantive feature and clarifies formula vs. target

image

cdverrett94:
Should the roll part be scrollable. I ran a combat today with a lot of enemies in a burst (big burst and enemies in various vertical spaces) and the overflow made the roll button hidden.

Should the button to pan to the token have a tooltip indicating what it does?

JPMeehan:
Yeah, both suggestions sound good

krbz999 (changes requested):
Receiving errors when attempting to target. Also not seeing targets in the dialog.

image

JPMeehan:
Hmm, not replicating? Starting with targets or not it's fine. Are you somehow targeting between "app constructed" and "context is prepped"?

JPMeehan:
The issue shows up for abilities that don't have a roll, fixed it, but we wouldn't expect targets for a non-rolled ability currently

krbz999:
Need to style the dialog such that it splits into two columns.

The buttons for panning would be better as anchors, and they need a tooltip.

Hovering out of the dialog does not always call the tokens' hover-out event, leaving the hover rectangle on the token.

Multiple tokens from the same actor appear once each when each actor should appear at most once.

JPMeehan:
Demo of <a> styling

image

From what I can tell, the hover is only "sticky" if you've got stacked windows and your mouse switches from a valid token to a new window. I'm not sure how to fix that without moving the listener binding outside of the element. Would it be better to bind to the `ownerDocument` and then clean up the listener on close? Seems like it would add complications to the detaching process.

Right now I'm using token IDs for uniqueness. How important is it that I check if the linked actor for each token is unique?

krbz999:
> Right now I'm using token IDs for uniqueness. How important is it that I check if the linked actor for each token is unique?

It is the correct thing to do.

As for token hovers, I don't have a good solution. Frankly I admit I don't see the value.

The pan icon itself would be the better anchor rather than the full name and icon.

JPMeehan:
Made the pan & actor adjustments. I think the token hover behavior is useful, and it's very easy to fix the hover status by returning to the application.

templates/api/roll-dialog-footer.hbs
```diff
@@ -1,6 +1,6 @@
 <footer class="form-footer">
   <button type="submit">
-    <i class="fa-solid fa-dice-d10"></i> {{localize "DRAW_STEEL.ROLL.Button"}}
+    <i class="fa-solid fa-dice-d10"></i> {{buttonLabel}}
```
krbz999:
```suggestion
    <i class="fa-solid fa-dice-d10" inert></i> <span>{{buttonLabel}}</span>
```

krbz999 (approved):
Can't have this hold up the release any longer. I grant that this is good enough for an MVP and more; any feedback can be incorporated in 1.1.0.

PR #1792, opened 2026-03-27
[#1722] Follower Roll API

JPMeehan:
Similar adjustments to #161 with deprecating the Roll.prompt methods in favor of just flowing through the actual dialog class. This doesn't add the sheet interactivity yet, I figure if the rest of the API logic checks out it's not hard to add a data-action to the actual follower items

### DIFF
--- a/src/module/data/item/follower.mjs
+++ b/src/module/data/item/follower.mjs
@@ -4,7 +4,7 @@ import { setOptions } from "../helpers.mjs";
 import { systemPath } from "../../constants.mjs";
 
 /**
- * @import DrawSteelItem from "../../documents/item.mjs";
+ * @import { DrawSteelChatMessage, DrawSteelItem } from "../../documents/_module.mjs";
  * @import { ApplicationConfiguration } from "@client/applications/_types.mjs";
  * @import { DatabaseCreateOperation } from "@common/abstract/_types.mjs";
  */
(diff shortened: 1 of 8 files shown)


src/module/data/item/follower.mjs
```diff
@@ -176,4 +182,59 @@ export default class FollowerModel extends BaseItemModel {
...
+  /* -------------------------------------------------- */
+
+  /**
+   * Perform a characteristic roll to assist a project.
+   * @param {object} config
+   * @param {DrawSteelItem} [config.project] A specific project to roll for.
+   * @param {ApplicationConfiguration} [dialogOptions={}] Options to be forwarded to the roll dialog.
+   * @param {DatabaseCreateOperation} [messageOptions]    Options to be forwarded to the final created chat message.
+   */
+  async roll(config = {}, dialogOptions = {}, messageOptions = {}) {
```
krbz999, line 196: Missing a return value.

src/module/data/item/project.mjs
```diff
@@ -67,22 +70,6 @@ export default class ProjectModel extends BaseItemModel {
 
   /* -------------------------------------------------- */
 
-  /** @inheritdoc */
-  preparePostActorPrepData() {
-    super.preparePostActorPrepData();
```
krbz999, line 72: Is this removal related to the rest of the PR?

krbz999 (approved):

src/module/data/item/project.mjs
JPMeehan: Yes — the value we were calculating assumed that the actor could tell us the highest available characteristic, but if we're using the follower then we have to calculate it during the the roll

Issue #1420, opened 2025-11-28
Allow ability selection to support non-core options.


JPMeehan:
As a player, I would like class and subclass ability selection to support expansions beyond what is provided in the core rules, e.g. a third party Conduit Signature ability option.

From a technical POV, this requires knowing what class or subclass an ability should be associated with, which suggests a DSID field. All subclass DSIDs should be unique, so no need to specify both. This is not a required field, as Titles, Kits, and Perks can all provide abilities that aren't tied to any class or subclass.

While most ability choices are strictly determined by class vs subclass & heroic resource cost, the Elementalist is an exception with some 5 cost heroic abilities being available at level 1 and also at 3, while others are only available in the level 3 slot. This means we also need a level requirement field.

PR #1795, opened 2026-03-28
Implemented ability registry

JPMeehan:
- Added registration of abilities
- Abilities are organized by resource cost rather than level because when the two are *not* in sync, resource cost is the primary determinant (setup for #1420)
- Leveraged ability registry to improve intelligence of Heroic Resource labels for compendium abilities

image

Note: Testing must be done on the Conduit because that's the only class we've structured with the appropriate prerequisites data. Fortunately, doing updates will be relatively simple because our foldering corresponds to the necessary data values.

Not labeling this as closing the ticket because a follow-up issue should tackle feature prerequisites (relevant for properly genericizing the Conduit subclass feature choices)

### DIFF
--- a/draw-steel.mjs
+++ b/draw-steel.mjs
@@ -64,7 +64,14 @@ Hooks.once("init", function () {
   }
 
   // Indexing DSID, class primary name, subclass associated classes, and perk types
-  CONFIG.Item.compendiumIndexFields.push("system._dsid", "system.primary", "system.classLink", "system.perkType");
+  CONFIG.Item.compendiumIndexFields.push(
+    "system._dsid",
+    "system.primary",
+    "system.classLink",
+    "system.perkType",
+    "system.prerequisites",
+    "system.resource",
+  );
   // Need to be able to find "configuration" type pages
   CONFIG.JournalEntry.compendiumIndexFields.push("pages.type");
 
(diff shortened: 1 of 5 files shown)


JPMeehan:
image

src/module/helpers/registry.mjs
```diff
@@ -84,6 +99,32 @@ export default class DrawSteelRegistry {
...
+          signature: new Set(),
+          heroic3: new Set(),
+          heroic5: new Set(),
+          heroic7: new Set(),
+          heroic9: new Set(),
+          heroic11: new Set(),
+        };
+        this.abilities.set(dsid, entry);
+      }
+      if (index.system.resource) entry[`heroic${index.system.resource}`]?.add(index.uuid);
```
krbz999, line 122: Wont this lead to some not being stored?

Rather if `index.system.resource` is numeric, we should construct the Set if it does not exist.

draw-steel.mjs
```diff
@@ -64,7 +64,15 @@ Hooks.once("init", function () {
   }
 
   // Indexing DSID, class primary name, subclass associated classes, and perk types
```
krbz999: comment needs adjustment

krbz999 (approved):

src/module/helpers/registry.mjs
JPMeehan: Not every ability should be stored in the registry since not everything is an advancement choice, e.g. the Tactician's Mark that every Tactician automatically gets

PR #1796, opened 2026-03-28
[#1436] Implemented plural rules

### LINKED ISSUE #1436: Generalize plural rules handling
The system has a lot of needs for plural rules. We should consider a common place to access it, like `ds.plural` or something, that gets initialized in `i18nInit`.

JPMeehan:
Closes #1436

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ This version of the system does not support v13, it is exclusively for Foundry v
 ### Changed
 
 - Refactored CONFIG.statusEffects usage from array to record, in accordance with core migration. (#1166)
+- Restructured i18n usage to conform to Intl.PluralRules. (#1436)
 - Migrated roll modes to the new core message modes. (#1681)
 
 ### Deprecated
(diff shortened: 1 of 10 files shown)


krbz999 (commented):
this doesnt really save any effort, one still has to instantiate a new `PluralRules` class for the ordinal case.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/PluralRules

JPMeehan:
I'm just using the instance that Foundry added in v14 and broadly leveraging it to improve our i18n

krbz999 (commented):
Do we want to hold off here until we discuss conventions for the i18n files?

JPMeehan:
I think these changes are just objectively correct, but I would be happy to have a discussion between now and 1.0 of overall structure. I think we're in a fairly good spot and if we move some pieces around that's fine, the possibility of that shouldn't block this pluralization improvement.

krbz999 (approved):
i agree, easier to merge and then refactor after in this case

PR #1798, opened 2026-03-29
Formula editor autocomplete context

cdverrett94:
Adds the formula field autocomplete contexts for a few of the most important roll data paths. I used existing localization keys for the characteristics, but for the other made new ones. I could have used the potency localization form the ability stuff, but without the context "Weak"/"Strong"/"Average" felt too vague, so wanted to include "Potency" in it.

It should be easy enough now to add additional entries if we decide there should be more.

Also added the `_applyChangeSubtract` to `FormulaField`. Adjusted when we check for `!value` to account for double negatives.

### DIFF
--- a/src/module/config.mjs
+++ b/src/module/config.mjs
@@ -2607,6 +2607,9 @@ export const formulaEditorContexts = {
     level: {
       label: "DRAW_STEEL.FormulaEditor.Level",
     },
+    echelon: {
+      label: "DRAW_STEEL.FormulaEditor.Echelon",
+    },
     "potency.weak": {
       label: "DRAW_STEEL.FormulaEditor.Potency.Weak",
     },
--- a/src/module/config.mjs
+++ b/src/module/config.mjs
@@ -2577,3 +2577,45 @@ export const references = {
   wealth: "Compendium.draw-steel.journals.JournalEntry.f8eNK5Pte4CSdex0.JournalEntryPage.I38Puf98ti2zZbiz",
   winded: "Compendium.draw-steel.journals.JournalEntry.f8eNK5Pte4CSdex0.JournalEntryPage.3yd9XsnxkDbR7Ywd",
 };
+
+/* -------------------------------------------------- */
+
+/**
+ * Default formula editor context for autocompletion.
+ * @type {Record<string, { label: string }>}    A record of roll data paths to labels.
+ */
+export const formulaEditorContexts = {
+  default: {
+    M: {
+      label: "DRAW_STEEL.Actor.characteristics.might.full",
+    },
+    A: {
+      label: "DRAW_STEEL.Actor.characteristics.agility.full",
+    },
+    R: {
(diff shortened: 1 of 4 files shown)


src/module/data/fields/formula-field.mjs
```diff
@@ -75,7 +85,9 @@ export default class FormulaField extends foundry.data.fields.StringField {
     return `min(${value}, ${delta})`;
   }
 
-  /** @override*/
+  /* -------------------------------------------------- */
+
+  /** @override */
```
JPMeehan:
```suggestion
  /** @inheritdoc */
```

src/module/config.mjs
JPMeehan, line 1: We should add Echelon here too

src/module/data/fields/formula-field.mjs
```diff
@@ -47,6 +47,16 @@ export default class FormulaField extends foundry.data.fields.StringField {
 
   /* -------------------------------------------------- */
 
+  /** @inheritdoc */
+  _applyChangeSubtract(value, delta, model, change) {
+    const operator = delta.startsWith("-") ? "+" : "-";
+    delta = delta.replace(/^[+-]/, "").trim();
+    if (!value) return (operator === "-") ? `-${delta}` : delta;
+    return `${value} ${operator} ${delta}`;
+  }
```
JPMeehan, line 54: What are the advantages of your implementation vs. what dnd5e did?
```js
  _applyChangeSubtract(value, delta, model, change) {
    if ( !value ) return `-(${delta})`;
    return `${value} - (${delta})`;
  }
```

cdverrett94: I had just modified the `_applyChangeAdd` method, but functionally? Probably none. But I think it makes the roll formula show up as weird. 

`1d10 - (5)`, `1d10 - (-5)` vs `1d10 - 5`, `1d10 + 5`

src/module/data/fields/formula-field.mjs
cdverrett94: When should we be using override vs inheritdoc? Seems like we prefer inheritdoc, but there are a number of instances of override in the system.

JPMeehan: Inheritdoc is basically always correct, `override` is really only correct if for some specific reason we *don't* want to inherit the core annotations (which I don't think is ever true)

src/module/data/fields/formula-field.mjs
krbz999: I do not see this working correctly with a delta with multiple parts. The parentheses are required.

krbz999:
I'm sure we can implement a way to simplify roll expressions further if that is a concern.

src/module/data/fields/formula-field.mjs
cdverrett94: Good point.

krbz999 (approved):

PR #1800, opened 2026-03-30
Compendium Data Migrations for 1.0

JPMeehan:
Implements the relevant migrations from the Duration update as well as some of the path shuffling from the schema updates

### DIFF
--- a/src/module/data/migrations.mjs
+++ b/src/module/data/migrations.mjs
@@ -185,7 +185,7 @@ async function version_1_0_migration() {
     const wasLocked = pack.config.locked;
     if (wasLocked) await pack.configure({ locked: false });
     const operation = [];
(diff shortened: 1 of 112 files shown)


src/module/data/migrations.mjs
```diff
@@ -238,7 +238,6 @@ export async function migrateChanges(parentDocument) {
     "flags.draw-steel.migrateChanges": _del,
   }));
 
-  // TODO: Batch this in v14
```
krbz999, line 241: I guess not?

krbz999 (approved):

src/module/data/migrations.mjs
JPMeehan: I'm unsure how aggressively to implement batching in a bunch of methods that were not currently using and work fine.

PR #1801, opened 2026-03-30
[#1771] Re-implemented turn event logic

### LINKED ISSUE #1771: Redo start/end of turn trigger logic
See https://github.com/foundryvtt/foundryvtt/issues/13983, but basically need to override `_manageTurnEvents` and redo all of the internals if changes are not made.

JPMeehan:
Core has a bunch of logic around skipped turns and assuming initiative only goes from the top to the bottom that was causing problems not only for our own bespoke end/start of turn logic but also was causing unwanted region event triggers (easy to recreate with Road to Broadhurst). Unfortunately, the logic I'm adjusting had a lot of private methods, so there was a bit of a baby-with-bathwater to just override the whole thing. I filed a ticket in core to get them to un-private this stuff but it seems unlikely to happen before Stable.

Closes #1771

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,7 @@ This version of the system does not support v13, it is exclusively for Foundry v
 
 ### Fixed
 
+- Corrected end/start of turn logic when going directly from monster to hero turns as well as overall processing for scene regions. (#1771)
 - Corrected CSS URL asset handling, restoring decorative boxes to asides.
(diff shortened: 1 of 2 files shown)


krbz999 (approved):
This is sad but unsurprising.

Generally I do not like having to basically re-implement core features to make things work correctly, but in the interest of bugfixing it seems necessary.

Only thing I would ask here is you add some heavy-handed comments for later reference as to why we (re)implemented these methods and how or when we can remove them.

PR #1802, opened 2026-03-30
Implemented batch operations for effect application

JPMeehan:
- Implemented `foundry.documents.modifyBatch` for our two effect application routes
- Addressed warning created by `ActiveEffect#origin` checking for relative UUIDs
- Caught  and fixed old expiry data (was being handled by the migration)

### DIFF
--- a/src/module/applications/ux/enrichers/apply-effect.mjs
+++ b/src/module/applications/ux/enrichers/apply-effect.mjs
@@ -3,6 +3,7 @@ import DrawSteelActiveEffect from "../../../documents/active-effect.mjs";
 
 /**
  * @import { ActiveEffectData } from "@common/documents/_types.mjs";
+ * @import { DatabaseWriteOperation } from "@common/abstract/_types.mjs";
  * @import { TextEditorEnricher, TextEditorEnricherConfig } from "@client/config.mjs";
  * @import HTMLEnrichedContentElement from "@client/applications/elements/enriched-content.mjs";
  * @import { ParsedConfig } from "../helpers.mjs";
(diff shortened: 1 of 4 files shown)


src/module/applications/ux/enrichers/apply-effect.mjs
```diff
@@ -152,25 +157,22 @@ async function onClickAnchor() {
...
-        },
-      ];
-
-      canvas.interface.createScrollingText(...scrollingTextArgs);
+        });
     }
   }
+
+  await foundry.documents.modifyBatch(toDelete);
+  await foundry.documents.modifyBatch(toCreate);
```
krbz999, line 177: We could combine these into one async call if we also had a `toUpdate`. It would require `diff: false, recursive: false`, so maybe you think that will be controversial.

src/module/data/pseudo-documents/power-roll-effects/applied-effect.mjs
```diff
@@ -198,24 +200,30 @@ export default class AppliedPowerRollEffect extends BasePowerRollEffect {
...
       // deleting instead of updating because there may be variances between the old copy and new
-      if (existing?.disabled || existing?.duration.expired) await existing.delete();
-      // not awaited to allow parallel processing
-      actor.createEmbeddedDocuments("ActiveEffect", [tempEffect.toObject()], { keepId: noStack });
+      if (existing) toDelete.push({ action: "delete", parent: actor, documentName: "ActiveEffect", ids: [tempEffect.id] });
+      toCreate.push({ action: "create", parent: actor, documentName: "ActiveEffect", data: [createData], keepId: noStack });
     }
+
+    await foundry.documents.modifyBatch(toDelete);
+    return foundry.documents.modifyBatch(toCreate);
```
krbz999, line 227: The return value is not documented.

It is perhaps not a very useful return value, but if my above suggestion was implemented it would be more meaningful.

krbz999 (approved):
Micro performance enhancement suggestions. Take it or leave it but please re-request review if you do.

src/module/applications/ux/enrichers/apply-effect.mjs
JPMeehan: The reason we're not updating is we don't know what adjustments have been made and how we should update (e.g. same effect different duration), only that there's a collision. You wanted delete and recreate so that's what this implements.

PR #1806, opened 2026-03-31
[#1159] Implemented extendFields

### LINKED ISSUE #1159: Implement `SchemaField#extendFields` and `removeFields`
In several data models we use helper functions to generate the internal contents of SchemaField instances, such as the Actor Biography. In v14 we should move to using the new SchemaField methods to improve nested property handling.

Reference: https://github.com/foundryvtt/foundryvtt/issues/13384

JPMeehan:
Fairly simple follow up to just remove the _actorBiography logic with extendFields

Closes #1159


krbz999 (approved):
You love to see it.

PR #1813, opened 2026-04-01
[#1753] Ensure core rolls array is populated

### LINKED ISSUE #1753: 3d Dice using Dice So Nice not consistently rolling
### What happened?

Hi,

While using DS system version 0.11.1 with the Dice So Nice module and 3d dice, when a player starts their turn and get a random 1d3 of heroic resources, there is no 3d die rolled on the screen.  The roll result is in chat, and the HR is gained, it is just the visual 3d die that is missing.  My players insist (and I agree) that the HR die roll used to include the visual 3d die, so we think something changed in a recent update to DS.  

With some more testing, what I've found is that when using Draw Steel as the system, and all modules disabled except for the Dice So Nice and Dice Tray (and the ones that have extra dice colors/textures):

- If you enter a dice formula in chat and hit enter, it rolls the 3d die.  
- If you use Dice Tray to select a die type and click the Roll button, it will not use the 3d die and only posts the result to chat.   
- If you use Dice Tray to select a die type, it adds the die formula in chat, and then if you click in chat and hit enter it will roll the 3d die.  

So there seems to be two ways to "submit" a die roll, and in the latest Draw Steel system one way triggers the 3d die (including hitting enter in chat or using a power roll), and one way does not (including the HR roll at start of turn or the Roll button in the dice tray).

Note:  Using the D&D 5e game system with the same Dice So Nice / Dice Tray modules active, the Roll button on the tray works normally and rolls the 3d die.  So it does seem to be Draw Steel system specific.

### Reproduction Steps

Install Draw Steel, Dice So Nice and Dice Tray.  When making a power roll, the 3d dice appear.  When starting a turn with a character class that rolls for heroic resources, the 3d die does not appear.  

In addition, in the Dice Tray, if you click a die type and hit the roll button it will not trigger the 3d die.  If you click a die type and then click in the chat box and hit enter, it will trigger the 3d die. 

Foundry v13.351, Draw Steel v0.11.1, Dice So Nice v5.2.5

### Bug Checklist

- [x] I am using The Forge for hosting

JPMeehan:
- DSN expects the `rolls` array to be populated, which the baseline DSRoll#toMessage was not doing
- Adjusted the logic in the `rollSave` method to have a fixed ID for the saving throw part

Closes #1753


krbz999 (approved):
Seems fine, though why do we want a static id?

JPMeehan:
Static ID makes weird usage more reliable for merging in messageData

PR #1823, opened 2026-04-05
Fix using non-power roll abilities

cdverrett94:
Using non-power roll abilities was erroring out because `roll` wasn't a valid mode. `applyMode` will infer the mode from the core setting if it's not provided.


JPMeehan (approved):
I had also noticed this and fixed it in another branch and also irrelevant once we do the ability use dialog, but yeah temp fix for the branch is helpful

PR #1825, opened 2026-04-05
Refactored size logic to use actor size

### LINKED ISSUE #1529: Implement token adjustments in relevant effects
See ~~https://github.com/foundryvtt/foundryvtt/issues/13461~~ https://github.com/foundryvtt/foundryvtt/issues/13969 for syntax

JPMeehan:
Now strongly inferring size for creatures. This change does not affect objects or anything else not inheriting from `CreatureModel` because they may have more irregular shapes.

### DIFF
--- a/src/module/data/actor/creature.mjs
+++ b/src/module/data/actor/creature.mjs
@@ -64,27 +64,9 @@ export default class CreatureModel extends BaseActorModel {
 
     // Creature token sizes can be strongly determined from actor size
     this.parent.tokenActiveEffectChanges.initial.push(
-      {
-        key: "width",
-        phase: "initial",
-        priority: 5,
-        type: "override",
-        value: this.combat.size.value,
-      },
-      {
-        key: "height",
-        phase: "initial",
-        priority: 5,
-        type: "override",
-        value: this.combat.size.value,
-      },
-      {
-        key: "depth",
-        phase: "initial",
-        priority: 5,
-        type: "override",
-        value: this.combat.size.value,
-      },
+      ...["width", "height", "depth"].map(key => ({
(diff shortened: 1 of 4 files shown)


src/module/data/actor/creature.mjs
```diff
@@ -61,6 +61,31 @@ export default class CreatureModel extends BaseActorModel {
...
+        value: this.combat.size.value,
+      },
+      {
+        key: "depth",
+        phase: "initial",
+        priority: 5,
+        type: "override",
+        value: this.combat.size.value,
+      },
+    );
```
krbz999, line 70:
```suggestion
    this.parent.tokenActiveEffectChanges.initial.push(
      ...["width", "height", "depth"].map(key => ({
        key, phase: "initial", priority: 5, type: "override", value: this.combat.size.value,
      })),
    );
```

src/module/documents/active-effect.mjs
```diff
@@ -87,7 +87,7 @@ export default class DrawSteelActiveEffect extends foundry.documents.ActiveEffec
 
   /** @inheritdoc */
   static applyChange(targetDoc, change, options = {}) {
-    if (typeof change.effect.system.apply === "function") return change.effect.system.apply(targetDoc, change, options);
+    if (typeof change.effect?.system.apply === "function") return change.effect.system.apply(targetDoc, change, options);
```
krbz999:
```suggestion
    if (typeof change.effect?.system.apply === "function")
      return change.effect.system.apply(targetDoc, change, options);
```

krbz999 (approved):
Just some style suggestions.

PR #1829, opened 2026-04-08
[#1694] Adjusted actor sheet handling of inactive statuses

### LINKED ISSUE #1694: Condition Buttons and Assign Status Effect for Inactive Effects
### User Experience

As a Director, when a condition ends either at the end of a turn or on save the condition is moved to the inactive effects. The red around the Status Conditions buttons and the bold on the bold for the Assign Status Effects however remain on as if the condition was already active. Clicking on it the button or status effect fully removes the condition. Would it be possible to treat the conditions as turned off when they are inactive? This would avoid potential confusiong when looking at those areas as well as mean that clicking the button will set activate it as currently it would require two clicks.

Image

JPMeehan:
1. The status displays at the top of the effects tab now *only* show as active if a status is active, not if an effect with that status is simply present (aka don't show as active if expired)
2. Toggling an effect will activate that status even if there are inactive copies of the status

Closes #1694


krbz999:
What?

JPMeehan:
1. We had the status highlighted even if the effects tagged with the status were inactive
2. In that scenario, clicking the status button would only clean up the inactive statuses instead of applying it

JPMeehan:
@krbz999 turns out we *can't* batch this operation because sometimes we need to delete a canonical status effect *and then* recreate that same effect, which Foundry doesn't like. Just reactivating is unreliable because we don't know what other differences there are like expiry event. I've adjusted the PR to move the logic into toggleStatusEffect itself, but maybe that's undesirable because this is a core method and now we're changing the behavior.

krbz999:
> @krbz999 turns out we _can't_ batch this operation because sometimes we need to delete a canonical status effect _and then_ recreate that same effect, which Foundry doesn't like. Just reactivating is unreliable because we don't know what other differences there are like expiry event. I've adjusted the PR to move the logic into toggleStatusEffect itself, but maybe that's undesirable because this is a core method and now we're changing the behavior.

You can call `_fromStatusEffect` and provide an update if an effect with the same `_id` exists, and use `diff: false`.

If the issue and this PR is also meant to improve the ux of 'noncanonical' status conditions, then this is not enough by far and we need to consider an alternative display on the actor sheets entirely.

So I remain unconvinced that this is the way to go.

JPMeehan:
What's your ideal handling on both the sheet & token HUD for effects with a static ID (all of the ones we really care about). Here's the current, pre-PR behavior.

- The Token HUD & `toggleStatusEffect` only care about the canonical status effect being *present*, not active
- Our actor sheets check against the `statuses` set of *all* applicable effects (e.g. if you've got a custom status that's "Weakened and Bleeding (save ends)")

I think the most defensible change in the PR is changing our actor sheet to also check if the applicable effect is active (i.e. not expired). 

The questions that I think need to be answered
1. What's the desired behavior when trying to toggle a status effect when the canonical one is expired?
2. How should we handle the few single-status effects we have without a fixed _id, like `flying`?

The PR's current answer is that if everything that might be adding a status is inactive, we should re-activate the status rather than only clean up the expired effects. Arguably, we should override `TokenHUD#_getStatusEffectChoices` to match our actor sheets.

krbz999:
Additional questions:
- Do we, and should we, consider noncanonical statuses on equal terms with the canonical ones? It seems like yes.
- If so, should we get rid of static ids?
- What purpose do the status effect buttons on the hero sheet serve if they toggle *other* effects and not the canonical effect?

JPMeehan:
The line we currently draw is that static `_id` means it's something in the Draw Steel rules, e.g. Bleeding. `flying` does not have an `_id` and does not appear on the actor sheet. Part of the logic here is "what of these conditions have tooltips". Flying and Burrowing exist for foundry's movement logic integration, same with blinded & invisible for the vision logic.

krbz999:
Aye, but a condition can have a tooltip without a static id, those two concepts are not related.

JPMeehan:
Agreed, although it's moderately useful in general with how Foundry has generally constructed the usage of _id with status effects.

JPMeehan:
Hmm. Weird staleness, closing PR and redoing.

PR #1831, opened 2026-04-09
[#1705] Characteristics Field

### LINKED ISSUE #1705: Establish fixed list of characteristics
This ends support for hypothetical adjustments to the characteristics, but simplifies a lot of the handling of these actor properties (especially since we plan to continue using the DS glyphs). A CharacteristicsField may be useful to capture the commonalities between Followers (an Item subtype) and Creatures (an Actor subtype).

JPMeehan:
A common shared field with a hard-private list of characteristics seemed like a better solution than using ds.CONST

Closes #1705

### DIFF
--- a/src/module/data/fields/characteristics-field.mjs
+++ b/src/module/data/fields/characteristics-field.mjs
@@ -1,0 +1,40 @@
+/**
+ * @import { DataFieldContext, DataFieldOptions } from "@common/data/_types.mjs";
+ */
+
+const { SchemaField, StringField, NumberField } = foundry.data.fields;
+
+/**
+ * A field that unifies and hardens the Characteristic data across creatures and followers.
+ */
+export default class CharacteristicsField extends foundry.data.fields.SchemaField {
+  /**
+   * @param {DataFieldOptions} [options]        Options which configure the behavior of the field.
+   * @param {DataFieldContext} [context]        Additional context which describes the field.
+   */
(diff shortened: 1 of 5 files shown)


src/module/data/fields/characteristics-field.mjs
```diff
@@ -0,0 +1,40 @@
+/**
+ * @import { DataFieldContext, DataFieldOptions } from "@common/data/_types.mjs";
+ */
+
+const { SchemaField, StringField, NumberField } = foundry.data.fields;
+
+/**
+ * A field that unifies and hardens the Characteristic data across creatures and followers.
+ */
+export default class CharacteristicsField extends foundry.data.fields.SchemaField {
```
krbz999:
```suggestion
export default class CharacteristicsField extends SchemaField {
```

src/module/data/fields/characteristics-field.mjs
```diff
@@ -0,0 +1,40 @@
+/**
+ * @import { DataFieldContext, DataFieldOptions } from "@common/data/_types.mjs";
+ */
+
+const { SchemaField, StringField, NumberField } = foundry.data.fields;
```
krbz999:
```suggestion
const { NumberField, SchemaField, StringField } = foundry.data.fields;
```

krbz999 (approved):

PR #1833, opened 2026-04-09
Test Edge and Forced Movement for Growing Ferocity. Feedback requested

alinuxfan:
@JPMeehan or @krbz999 , requesting feedback on if these are appropriate usages of `power.roll.edges` and `forced.bonuses.xxx`


JPMeehan (commented):
Hmm, interesting question. The filters aren't super precise (all abilities), but since it's applied that seems helpful. I believe we also support edges for might and agility tests directly?

PR #1840, opened 2026-04-10
Implement a `SquadModel#takeDamage` method

### LINKED ISSUE #1353: Feature Request: Damage capped at minion stamina

cdverrett94:
Benefits and Improvements: 
- Allows us to only account for immunity and weakness only once for the whole squad instead of per minion as was before.
- Adds an `aoe` option that when true, clamps the damage per minion to their stamina threshold
   - The `aoe` option can either be set as an option on the damage roll or passed as an option to the `takeDamage` method. 
   - It automatically gets set from the damage PRE's `toDamageRoll` method.

### DIFF
--- a/src/module/data/combatant-group/squad.mjs
+++ b/src/module/data/combatant-group/squad.mjs
@@ -1,4 +1,5 @@
 import BaseCombatantGroupModel from "./base.mjs";
+import { DrawSteelActor } from "../../documents/_module.mjs";
 import DrawSteelCombatant from "../../documents/combatant.mjs";
 
 const fields = foundry.data.fields;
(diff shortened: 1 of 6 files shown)


src/module/data/pseudo-documents/power-roll-effects/damage-effect.mjs
```diff
@@ -208,6 +208,7 @@ export default class DamagePowerRollEffect extends BasePowerRollEffect {
       flavor,
       type: damageType,
       ignoredImmunities,
+      aoe: this.parent.hasTemplate,
```
JPMeehan: I think we should base on keywords, for example `"special"` does not get tagged as having a valid template but could still have the area keyword

src/module/data/combatant-group/squad.mjs
```diff
@@ -151,4 +152,33 @@ export default class SquadModel extends BaseCombatantGroupModel {
...
+   * @param {Array<DrawSteelActor>} minions    The minions that are taking the damage.
+   * @param {number} damagePerMinion           The amount of damage to take.
+   * @param {object} [options]                 Options to modify the damage application.
+   * @param {string} [options.type]            Valid damage type.
+   * @param {boolean} [options.aoe]            Is this an AOE that should have the damage capped?
+   * @param {Array<string>} [options.ignoredImmunities]  Which damage immunities to ignore.
+   * @returns {Promise<DrawSteelCombatantGroup>}
+   */
+  async takeDamage(minions, damagePerMinion, options = {}) {
+    if (!minions.length) return this;
```
JPMeehan, line 169: Feels like we should have some check that the actors are part of the squad? It's an async method running only when called so that shouldn't be a meaningful performance concern

src/module/rolls/damage.mjs
```diff
@@ -117,5 +134,11 @@ export default class DamageRoll extends DSRoll {
       }
       else await actor.system.takeDamage(amount, { type: this.type, ignoredImmunities: this.ignoredImmunities });
     }
+
+    // Damage minion sqauds
+    for (const actors of Object.values(groups)) {
+      const group = actors[0].system.combatGroup;
```
JPMeehan: If you used the group UUID as the key, you could then just fromUuidSync rather than this actors[0] bit which feels messy

JPMeehan (commented):
Overall looks good and makes sense. Will have to sit for a bit until we open up the 1.1 branch.

src/module/data/combatant-group/squad.mjs
cdverrett94: Are you thinking:
- Filter the `minions` array down to only minions in the squad 
- or check if every minion in `minions` is in the squad and if not, throw an error?

cdverrett94:
Alright, I've made the changes per the review. 

I did also add an option for damage enrichers to accept an aoe parameter with it automatically tagging as aoe if the `relativeTo` document is an area ability. This is so that abilities like the Edict of Disruptive Isolation Censor ability can properly cap their damage.

src/module/data/combatant-group/squad.mjs
JPMeehan: Throw an error I think

PR #1864, opened 2026-04-17
[#1849] Added place template button that targets to ability dialog

### LINKED ISSUE #1849: Place Template to Target in Ability Configuration Dialog
### User Experience

As a Player, I would like a 'Place Template' button added to the Ability Configuration Dialog. Ideally, a template placed via this would target all creatures inside the template (even more ideally, it appropriately targets enemy/ally depending on the ability). This makes it much easier to identify who/where I can target _before_ I roll.

I know there is an option nested in the dropdown menu of the ability itself, but it feels a little awkward to find/navigate to. Similarly, having the template in the chat message after I've already rolled does not help in the initial targeting phase.

JPMeehan:
image

Closes #1849

### DIFF
--- a/src/module/applications/apps/ability-configuration-dialog.mjs
+++ b/src/module/applications/apps/ability-configuration-dialog.mjs
@@ -17,6 +17,7 @@ export default class AbilityConfigurationDialog extends PowerRollDialog {
     classes: ["ability-configuration-dialog"],
     actions: {
       panToken: this.#panToken,
+      placeTemplate: this.#placeTemplate,
     },
   };
 
@@ -264,5 +265,27 @@ export default class AbilityConfigurationDialog extends PowerRollDialog {
     const { tokenUuid } = target.closest("[data-token-uuid]").dataset;
     const token = fromUuidSync(tokenUuid);
     await canvas.animatePan({ x: token.x, y: token.y });
+  }
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Place a template on the canvas and select all tokens inside.
+   * @this AbilityConfigurationDialog
+   * @param {PointerEvent} event   The originating click event.
+   * @param {HTMLElement} target   The capturing HTML element which defined a [data-action].
+   */
+  static async #placeTemplate(event, target) {
+    const region = await this.item.system.placeTemplate();
+    if (!region) return;
+
+    /** @type {Set<DrawSteelToken>} */
+    const tokens = canvas.tokens.quadtree.getObjects(region.bounds);
+    for (const token of tokens) {
+      const tokenDoc = token.document;
+      if (!this.item.system.validTarget(tokenDoc)) continue;
+      if (!tokenDoc.testInsideRegion(region, tokenDoc._source)) continue;
+      token._updateTarget(true, game.user);
+    }
   }
 }
--- a/src/module/applications/apps/ability-configuration-dialog.mjs
+++ b/src/module/applications/apps/ability-configuration-dialog.mjs
@@ -276,6 +276,6 @@ export default class AbilityConfigurationDialog extends PowerRollDialog {
    * @param {HTMLElement} target   The capturing HTML element which defined a [data-action].
    */
   static async #placeTemplate(event, target) {
-    await this.item.system.placeTemplate({ setTargets: "acquire" });
+    this.item.system.placeTemplate({ setTargets: "acquire" });
   }
 }
--- a/src/module/applications/apps/ability-configuration-dialog.mjs
+++ b/src/module/applications/apps/ability-configuration-dialog.mjs
@@ -276,18 +276,6 @@ export default class AbilityConfigurationDialog extends PowerRollDialog {
    * @param {HTMLElement} target   The capturing HTML element which defined a [data-action].
    */
   static async #placeTemplate(event, target) {
-    const region = await this.item.system.placeTemplate();
-    if (!region) return;
-
-    /** @type {Set<DrawSteelToken>} */
-    const tokens = canvas.tokens.quadtree.getObjects(region.bounds);
-    const targetIds = [];
-    for (const token of tokens) {
-      const tokenDoc = token.document;
-      if (!this.item.system.validTarget(tokenDoc)) continue;
-      if (!tokenDoc.testInsideRegion(region, tokenDoc._source)) continue;
-      targetIds.push(token.id);
-    }
-    if (targetIds.length) canvas.tokens.setTargets(targetIds, { mode: "acquire" });
+    await this.item.system.placeTemplate({ setTargets: "acquire" });
   }
 }
--- a/src/module/applications/apps/ability-configuration-dialog.mjs
+++ b/src/module/applications/apps/ability-configuration-dialog.mjs
@@ -281,11 +281,13 @@ export default class AbilityConfigurationDialog extends PowerRollDialog {
 
     /** @type {Set<DrawSteelToken>} */
     const tokens = canvas.tokens.quadtree.getObjects(region.bounds);
+    const targetIds = [];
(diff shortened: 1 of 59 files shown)


cdverrett94:
Just was looking through this. Should the target selection be a part of the `AbilityModel#placeTemplate` so that target selection can work when using it from the context menu on the actor sheet, the button on the chat card, or just via the API in like a macro?

JPMeehan:
Possibly yeah, maybe throw in an option on place template?

cdverrett94:
> Possibly yeah, maybe throw in an option on place template?

Yeah, adding an option to the place template method would be good. It could possibly take `false | "acquire" | "replace"` with aquire/replace being used in the set targets call. 

Just an idea.

src/module/applications/apps/ability-configuration-dialog.mjs
```diff
@@ -265,4 +266,16 @@ export default class AbilityConfigurationDialog extends PowerRollDialog {
...
+  /* -------------------------------------------------- */
+
+  /**
+   * Place a template on the canvas and select all tokens inside.
+   * @this AbilityConfigurationDialog
+   * @param {PointerEvent} event   The originating click event.
+   * @param {HTMLElement} target   The capturing HTML element which defined a [data-action].
+   */
+  static async #placeTemplate(event, target) {
+    await this.item.system.placeTemplate({ setTargets: "acquire" });
```
krbz999:
```suggestion
    this.item.system.placeTemplate({ setTargets: "acquire" });
```

src/module/data/actor/creature.mjs
```diff
@@ -44,6 +44,16 @@ export default class CreatureModel extends BaseActorModel {
 
   /* -------------------------------------------------- */
 
+  /**
+   * Simple getter to indicate this counts as an creature for targeting purposes.
```
krbz999:
```suggestion
   * Simple getter to indicate this counts as a creature for targeting purposes.
```

src/module/data/item/ability.mjs
```diff
@@ -746,6 +746,46 @@ export default class AbilityModel extends BaseItemModel {
...
+
+    if (token.actor === this.actor) return targetOptions.has("self");
+    if (token.actor.system.isObject && targetOptions.has("object")) return true;
+
+    if (token.actor.system.isCreature) {
+      if (targetOptions.has("ally") && targetOptions.has("enemy")) return true;
+      else if (targetOptions.has("ally")) return sourceDisposition === token.disposition;
+      else if (targetOptions.has("enemy")) return (sourceDisposition !== token.disposition) && polarization.includes(sourceDisposition);
+    }
+    return false;
```
krbz999, line 843: Some of these variables are not used outside the isCreature check.

```suggestion
  validTarget(token) {
    const targetOptions = ds.CONFIG.abilities.targets[this.target.type]?.targetOptions;
    if (!targetOptions) return false;

    if (token.actor.uuid === this.actor.uuid) return targetOptions.has("self");
    if (token.actor.system.isObject && targetOptions.has("object")) return true;

    if (token.actor.system.isCreature) {
      const sourceDisposition = this.actor.getDependentTokens({ scenes: canvas.scene })?.at(0)?.disposition
        ?? this.actor.prototypeToken.disposition;
      const polarization = [CONST.TOKEN_DISPOSITIONS.FRIENDLY, CONST.TOKEN_DISPOSITIONS.HOSTILE];
      if (targetOptions.has("ally") && targetOptions.has("enemy")) return true;
      else if (targetOptions.has("ally")) return sourceDisposition === token.disposition;
      else if (targetOptions.has("enemy")) return (sourceDisposition !== token.disposition) && polarization.includes(sourceDisposition);
    }
    return false;
```

templates/apps/ability-configuration-dialog/roll.hbs
```diff
@@ -2,7 +2,14 @@
 <section class="roll-config">
   {{#if ability.system.power.roll.enabled}}
   <div class="global group">
-    <div class="header">{{formula}}</div>
+    <div class="header">
+      {{formula}}
+      {{#if ability.system.hasTemplate}}
+      <a data-action="placeTemplate" data-tooltip="DRAW_STEEL.Item.ability.placeTemplate">
+        <i class="fa-solid fa-ruler-combined"></i>
```
krbz999:
```suggestion
        <i class="fa-solid fa-ruler-combined" inert></i>
```

krbz999 (changes requested):
Discoverability is going to be a major issue, I can already tell.

Follow-up questions:
- I can click the template button multiple times. Should a subsequent usage remove previous ones? Should finalizing the roll remove the regions?
- Should we use the region icon instead of the old template icon?

Overall, I see the value, but I don't love the UI. Maybe that's just the dialog's fault, hopefully Jalen gets to general forms.

JPMeehan:
- I think caching and cleaning up would make a fair amount of sense. Can easily switch the targeting to replace & clean up the template at the end of the roll or on re-use. I don't think we have "pick x spaces" a la meteor swarm type stuff to worry about, and if we do we can add in a `count` field or something.
- I've been using the old template icon throughout this in part because I like the *terms* even if we're actually creating regions. Plus it lines up with the "Measured Template mode".

krbz999 (changes requested):
While the code looks fine and I agree we should clean up regions (and replace targets when doing so), we do have the ability to configure an Ability to place multiple regions in my testing, so storing only 1 region is not sufficient.

JPMeehan:
Could you elaborate on "configure to place multiple regions"? The Wall setup is one region with multiple shapes

krbz999:
> Could you elaborate on "configure to place multiple regions"? The Wall setup is one region with multiple shapes

Yeah that's my mistake. It's one region.

src/module/config.mjs
```diff
@@ -1586,12 +1586,18 @@ const abilityDistances = {
...
+
 /**
  * @typedef AbilityTarget
  * @property {string} label
  * @property {string} [all]       I18n key for an ability that targets everything within an area.
- * @property {string} embedLabel  Format string for display in the ability embed. Can point to a plural object @see Intl.PluralRules.
+ * @property {string} embedLabel  Format string for display in the ability embed. Can point to a plural object.
  * @property {Reference} [reference] An optional UUID with a description of the ability targets.
+ * @property {Set<TargetOptions>} targetOptions Target options.
  */
```
krbz999:
```suggestion
/**
* @typedef AbilityTarget
* @property {string} label
* @property {string} [all]       I18n key for an ability that targets everything within an area.
 * @property {string} embedLabel  Format string for display in the ability embed. Can point to a plural object.
* @property {Reference} [reference] An optional UUID with a description of the ability targets.
 * @property {Set<"ally" | "enemy" | "self" | "object">} targetOptions Target options.
*/
```

JPMeehan:
Yeah, and thinking back even something like Meteor Swarm would just be more shapes in a single region (it's literally the API docs example)

src/module/data/item/ability.mjs
```diff
@@ -806,6 +806,44 @@ export default class AbilityModel extends BaseItemModel {
...
+
+    /** @type {Set<DrawSteelToken>} */
+    const tokens = canvas.tokens.quadtree.getObjects(region.bounds);
+    const targetIds = [];
+    for (const token of tokens) {
+      const tokenDoc = token.document;
+      if (!this.validTarget(tokenDoc)) continue;
+      if (!tokenDoc.testInsideRegion(region)) continue;
+      targetIds.push(token.id);
+    }
```
```suggestion
    const targetIds = tokens.filter(token => {
      return this.validTarget(token.document) && token.document.testInsideRegion(region);
    }).map(token => token.id);
```

krbz999 (approved):

Issue #1843, opened 2026-04-11
Implement Imbue Project Data


JPMeehan:
The various "Imbue" projects have semi-standardized characteristic and goal values based on level and category, but each specific imbue has a unique item prerequisite & language source. Like treasures, effects should support project data that when dropped onto the projects tab gets turned into an Imbue project with the AE stored, and then either a new item with the imbued effect is created at the end of a project *or* an existing item is updated.

cdverrett94:
I'll tackle this.

cdverrett94:
With this getting worked on, how do we want to handle #1195?

JPMeehan:
We also need to figure out how to handle ordinary treasures granting abilities. I think for now we make the abilities separately and figure out the linking later.

PR #1866, opened 2026-04-17
Add project details to AE schema

cdverrett94:
Related to #1843 

Figured I'd do this in 3 stages. Schema -> Project Item Integration -> Compendium Updates


JPMeehan (approved):
I think the project info should probably only show if it *doesn't* have a parent?

cdverrett94:
That's a good idea!

PR #1878, opened 2026-04-23
AE project integration

### LINKED ISSUE #1843: Implement Imbue Project Data

cdverrett94:
Closes #1843 

Adds the integration with AE to Projects. If an AE is dropped on the project tab, it will create a new project with the yielded item set to the AE. Once the project is completed, it prompts for which item to attach the AE to. If blank is selected or it's closed without selection, it will create a new item and attach the AE to it.

### DIFF
--- a/src/module/applications/sheets/hero-sheet.mjs
+++ b/src/module/applications/sheets/hero-sheet.mjs
@@ -569,7 +569,7 @@ export default class DrawSteelHeroSheet extends DrawSteelActorSheet {
   async _onDropActiveEffect(event, effect) {
     // If the effect is dropped onto the project tab, create the effect as a project instead
     const projectDropTarget = event.target.closest("[data-application-part='projects']");
-    if (projectDropTarget && (this.actor.uuid !== effect.parent?.uuid)) {
+    if (projectDropTarget && (this.actor !== effect.parent)) {
       await effect.system.createProject(this.actor);
       return;
     }
--- a/src/module/applications/sheets/hero-sheet.mjs
+++ b/src/module/applications/sheets/hero-sheet.mjs
@@ -547,7 +547,7 @@ export default class DrawSteelHeroSheet extends DrawSteelActorSheet {
         // If it's a treasure dropped on the project tab, create the item as a project
         if (projectDropTarget && (item.type === "treasure")) {
           const name = _loc("DRAW_STEEL.Item.project.Craft.ItemName", { name: item.name });
-          return { name, type: "project", "system.yield.item": item.uuid };
+          return { name, type: "project", "system.yield.document": item.uuid };
         }
         else if (item.supportsAdvancements && (item.getEmbeddedCollection("Advancement").size > 0)) {
           ui.notifications.error("DRAW_STEEL.SHEET.NoCreateAdvancement", { format: { name: item.name } });
--- a/src/module/applications/sheets/hero-sheet.mjs
+++ b/src/module/applications/sheets/hero-sheet.mjs
@@ -562,4 +562,18 @@ export default class DrawSteelHeroSheet extends DrawSteelActorSheet {
     await this.actor.createEmbeddedDocuments("Item", droppedItemData.filter(_ => _), { keepId: true });
     return folder;
   }
+
+  /* -------------------------------------------------- */
+
+  /** @inheritdoc */
+  async _onDropActiveEffect(event, effect) {
+    // If the effect is dropped onto the project tab, create the effect as a project instead
+    const projectDropTarget = event.target.closest("[data-application-part='projects']");
+    if (projectDropTarget && (this.actor.uuid !== effect.parent?.uuid)) {
+      await effect.system.createProject(this.actor);
+      return;
+    }
+
+    return super._onDropActiveEffect(event, effect);
+  }
 }
--- a/src/module/data/item/project.mjs
+++ b/src/module/data/item/project.mjs
@@ -404,7 +404,7 @@ export default class ProjectModel extends BaseItemModel {
   async _createCraftedEffect(doc, amount) {
     // Prompt for adding to existing item, or adding to a new item.
     const treasures = this.actor.items.documentsByType.treasure
-      .filter(treasure => treasure.system.category === "leveled")
+      .filter(treasure => (treasure.system.category === "leveled") && (treasure.system.kind === doc.system.project.yield.kind))
       .map(treasure => ({ label: treasure.name, value: treasure.id }));
     const { createFormGroup, createSelectInput } = foundry.applications.fields;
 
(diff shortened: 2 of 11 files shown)


src/module/applications/sheets/hero-sheet.mjs
```diff
@@ -562,4 +562,18 @@ export default class DrawSteelHeroSheet extends DrawSteelActorSheet {
...
     return folder;
   }
+
+  /* -------------------------------------------------- */
+
+  /** @inheritdoc */
+  async _onDropActiveEffect(event, effect) {
+    // If the effect is dropped onto the project tab, create the effect as a project instead
+    const projectDropTarget = event.target.closest("[data-application-part='projects']");
+    if (projectDropTarget && (this.actor.uuid !== effect.parent?.uuid)) {
```
JPMeehan: can't we simplify and do `this.actor !== effect.parent`?

src/module/data/item/project.mjs
```diff
@@ -75,23 +75,23 @@ export default class ProjectModel extends BaseItemModel {
     const allowed = await super._preCreate(data, options, user);
     if (allowed === false) return false;
 
-    // If creating with item UUID, transfer the item project data to the project item
-    const itemUUID = data.system?.yield?.item;
-    const yieldItem = await fromUuid(itemUUID);
-    if (yieldItem?.type === "treasure") {
-      const { prerequisites, rollCharacteristic, goal, source } = yieldItem.system.project;
+    // If creating with a doucment UUID, transfer the document's project data to the project item.
```
JPMeehan:
```suggestion
    // If creating with a document UUID, transfer the document's project data to the project item.
```

src/module/data/item/project.mjs
JPMeehan: I think we should migrate the key of this field, it's trivial to do since it's a 1:1 swap. Doesn't even need a permanent migration. We should also probably specify embed: false?

src/module/data/item/project.mjs
```diff
@@ -344,29 +344,83 @@ export default class ProjectModel extends BaseItemModel {
...
+      },
+    });
+  }
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Perform the item creation and updates when the yielded document is an Item.
+   * @param {*} doc      The document yielded from this project.
+   * @param {*} amount   The amount yielded.
```
JPMeehan: Improve precision of these + add `@protected`

src/module/data/item/project.mjs
```diff
@@ -344,29 +344,83 @@ export default class ProjectModel extends BaseItemModel {
...
-    ui.notifications.success(`DRAW_STEEL.Item.project.Craft.CompletedNotification.${labelSuffix}`, {
-      format: {
-        actor: this.actor.name,
-        amount,
-        item: item.name,
-      },
+  /**
+   * Perform the item creation and updates when the yielded document is an Item.
+   * @param {*} doc      The document yielded from this project.
+   * @param {*} amount   The amount yielded.
```
JPMeehan: Improve precision of these + add `@protected`

src/module/data/item/project.mjs
```diff
@@ -344,29 +344,83 @@ export default class ProjectModel extends BaseItemModel {
...
+   */
+  async _createCraftedEffect(doc, amount) {
+    // Prompt for adding to existing item, or adding to a new item.
+    const treasures = this.actor.items.documentsByType.treasure.filter(treasure => treasure.system.category === "leveled").map(treasure => ({ label: treasure.name, value: treasure.id }));
```
JPMeehan: Break up into multiple lines, probably at the `.filter` and `.map`

src/module/data/item/project.mjs
```diff
@@ -344,29 +344,83 @@ export default class ProjectModel extends BaseItemModel {
...
+    const { createFormGroup, createSelectInput } = foundry.applications.fields;
+
+    const content = document.createElement("div");
```
JPMeehan, line 411: I think we've adjusted how we handle this with something like `ownerDocument` or the like

src/module/data/item/project.mjs
```diff
@@ -344,29 +344,83 @@ export default class ProjectModel extends BaseItemModel {
...
+      input: createSelectInput({
+        name: "treasure",
+        options: treasures,
+        blank: "",
+      }),
+    }));
+
+    const fd = await ds.applications.api.DSDialog.input({
+      content,
+      window: { title: "DRAW_STEEL.Item.project.Craft.EffectDialog.Title" },
```
JPMeehan: I can fill in an icon if you think the only good options are pro, but I'm pretty sure there's good free icons for this. Remember it's now 7.2

src/module/data/item/project.mjs
```diff
@@ -344,29 +344,83 @@ export default class ProjectModel extends BaseItemModel {
...
     });
+
+    let item;
+    if (fd?.treasure) item = this.actor.items.get(fd.treasure);
+    else {
+      const defaultName = getDocumentClass("Item").defaultName({ type: "treasure", parent: this.actor });
+      item = (await this.actor.createEmbeddedDocuments("Item", [{ name: defaultName, type: "treasure", "system.category": "leveled" }]))[0];
```
JPMeehan: I think we can do `Item.create` here and specify the parent & I suppose `pack`, the [0] is aesthetically displeasing

src/module/data/item/project.mjs
cdverrett94: Whoops!

src/module/data/item/project.mjs
cdverrett94: I thought about `ownerDocument`, but unless I'm missing something, there's no element to grab it from.

src/module/data/item/project.mjs
cdverrett94: Forgot `Item.create` can take a `parent` option. What would the `pack` option be in this case?

src/module/data/item/project.mjs
cdverrett94: Got the migration done. As far as doing embedded: false, that would break the ability to start a crafting project from a treasure. We could change that to get the `_stats.duplicateSource` or `_stats.compendiumSource`, but that feels like a bigger change than for this PR.

src/module/data/item/project.mjs
cdverrett94: So thoughts on icons. We could add a `kind` field to the AE project details which would take armor, weapon, or implement. Then for the icon, we could display either a FA icon for armor, sword, or wand depending on the kind. We could also use this to create or filter items on the actor to leveled treasures of a specific kind.

src/module/data/item/project.mjs
JPMeehan: yeah nah that's fine

src/module/data/item/project.mjs
JPMeehan: Can we grab from `this.parent.sheet`?

src/module/data/item/project.mjs
JPMeehan: That makes a lot of sense to me

src/module/data/item/project.mjs
JPMeehan: Would just be `actor.pack`

JPMeehan: Double checked, if you provide a parent you do not need to add in `pack`

JPMeehan (commented):
I think the `kind` + icons addition is the only thing this PR needs

cdverrett94:
Sorry, I had been busy. I've got the icons and kind done. The only icon I couldn't manage was a sword icon. Looks like those are all pro, so if you can add "fa-solid fa-sword" to the `icon` entry I added to the config.

JPMeehan (approved):

PR #1886, opened 2026-04-26
NPC Embeds (No ability list yet)

### LINKED ISSUE #901: Define NPC Embed
As a content creator, I would like to be able to use `@Embed` and produce a stat block.

jalensailin:
This MR adds the ability to embed NPC Actors using the `@Embed` syntax:

image

- As discussed, this just covers the "top part" of the stat block, and does not implement the NPC's list of abilities. That will happen in a future MR.
- The primary gradient color is automatically determined by the NPC's role. 
- The only thing that isn't implemented in the "top part" is the `With Captain` section. I noticed that this is an AE on the Actor, and thus it's a little difficult to translate into human readable format without some hacky work pulling the labels from the data model. I propose that we add the human-readable text to the AE's description, which I am happy to do as part of this MR. For example, if the AE gives a `+2` bonus to `system.stamina.max`, we would simply add the description "+2 bonus to Stamina". Let me know what you think.

Closes #901

### DIFF
--- a/src/module/data/actor/base-actor.mjs
+++ b/src/module/data/actor/base-actor.mjs
@@ -453,4 +453,35 @@ export default class BaseActorModel extends DrawSteelSystemModel {
   async updateResource(delta) {
     throw new Error("This method is abstract and must be implemented by a subclass");
   }
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Constructs an object with the formatted immunities and weaknesses with a list of damage labels.
+   * @returns {{immunities: string, weaknesses: string, labels: Record<string, string>}}
+   * @protected
+   */
+  _getImmunitiesWeaknesses() {
+    const labels = {
+      all: _loc("DRAW_STEEL.Actor.base.FIELDS.damage.immunities.all.label"),
+      ...Object.entries(ds.CONFIG.damageTypes).reduce((acc, [type, { label }]) => {
+        acc[type] = label;
+        return acc;
+      }, {}),
+    };
+
+    const immunities = Object.entries(this.damage.immunities)
+      .filter(([damageType, value]) => value > 0)
+      .map(([damageType, value]) => `<span class="immunity">${labels[damageType]} ${value}</span>`);
+    const weaknesses = Object.entries(this.damage.weaknesses)
+      .filter(([damageType, value]) => value > 0)
+      .map(([damageType, value]) => `<span class="weakness">${labels[damageType]} ${value}</span>`);
+
+    const formatter = game.i18n.getListFormatter({ type: "unit" });
+    return {
+      immunities: formatter.format(immunities),
+      weaknesses: formatter.format(weaknesses),
+      labels,
+    };
+  }
 }
--- a/src/module/data/actor/base-actor.mjs
+++ b/src/module/data/actor/base-actor.mjs
@@ -484,4 +484,40 @@ export default class BaseActorModel extends DrawSteelSystemModel {
       labels,
     };
   }
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Constructs an object with the actor's movement types as well as all options available from CONFIG.Token.movement.actions.
+   * @param {boolean} [excludeWalk=false] Whether to include the Walk movement type.
+   * @returns {{canHover: boolean, list: string, options: FormSelectOption[]}}
+   * @protected
+   */
+  _getMovement(excludeWalk = false) {
+    const formatter = game.i18n.getListFormatter({ type: "unit" });
+    const actorMovement = this.movement;
+    const canHover = actorMovement.types.has("fly") || actorMovement.types.has("teleport");
+    const movementList = Array.from(actorMovement.types).map(m => {
+      let label = _loc(CONFIG.Token.movement.actions[m]?.label ?? m);
+      if ((m === "teleport") && (actorMovement.teleport !== actorMovement.value)) label += " " + actorMovement.teleport;
+      return label;
+    });
+
+    if (canHover && actorMovement.hover) movementList.push(_loc("DRAW_STEEL.Actor.base.FIELDS.movement.hover.label"));
+
+    if (excludeWalk) {
+      const walkIndex = movementList.indexOf(_loc(CONFIG.Token.movement.actions.walk.label));
+      movementList.splice(walkIndex, 1);
+    }
+
+    return {
+      canHover,
+      list: formatter.format(movementList),
+      options: Object.entries(CONFIG.Token.movement.actions)
+        .filter(([key, _action]) => ds.CONFIG.speedOptions.includes(key))
+        .map(([value, { label }]) => ({ value, label })),
+      show: !!this.movement.value,
+
+    };
+  }
 }
--- a/src/module/data/actor/base-actor.mjs
+++ b/src/module/data/actor/base-actor.mjs
@@ -459,7 +459,6 @@ export default class BaseActorModel extends DrawSteelSystemModel {
   /**
    * Constructs an object with the formatted immunities and weaknesses with a list of damage labels.
    * @returns {{immunities: string, weaknesses: string, labels: Record<string, string>}}
-   * @protected
    */
   _getImmunitiesWeaknesses() {
     const labels = {
@@ -491,7 +490,6 @@ export default class BaseActorModel extends DrawSteelSystemModel {
    * Constructs an object with the actor's movement types as well as all options available from CONFIG.Token.movement.actions.
    * @param {boolean} [excludeWalk=false] Whether to include the Walk movement type.
    * @returns {{canHover: boolean, list: string, options: FormSelectOption[]}}
-   * @protected
    */
   _getMovement(excludeWalk = false) {
     const formatter = game.i18n.getListFormatter({ type: "unit" });
--- a/src/module/data/actor/npc.mjs
+++ b/src/module/data/actor/npc.mjs
@@ -211,7 +211,7 @@ export default class NPCModel extends CreatureModel {
     const context = {
       actor: this.parent,
       characteristics: this._getCharacteristics(false),
-      damageIW: this.parent.sheet._getImmunitiesWeaknesses(),
+      damageIW: this._getImmunitiesWeaknesses(),
       monsterKeywords: this.parent.sheet._getMonsterKeywords().join(", "),
       movement: this.parent.sheet._getMovement().list.replace(walkRe, "").trim(),
       system: this,
--- a/src/module/data/actor/npc.mjs
+++ b/src/module/data/actor/npc.mjs
@@ -1,8 +1,7 @@
-import { systemID, systemPath } from "../../constants.mjs";
 import { requiredInteger, setOptions } from "../helpers.mjs";
+import { systemID, systemPath } from "../../constants.mjs";
 import CreatureModel from "./creature.mjs";
 import SourceModel from "../models/source.mjs";
-import { systemID } from "../../constants.mjs";
 
 /**
  * @import { DrawSteelActor, DrawSteelItem } from "../../documents/_module.mjs";
--- a/src/module/data/actor/npc.mjs
+++ b/src/module/data/actor/npc.mjs
@@ -204,10 +204,11 @@ export default class NPCModel extends CreatureModel {
   async toEmbed(config, options) {
     // All NPCs are rendered inline
     config.inline = true;
-    console.log(config, options);
+    // console.log(config, options);
 
     const context = {
       actor: this.parent,
+      monsterKeywords: this.parent.sheet._getMonsterKeywords(),
       system: this,
       systemFields: this.schema.fields,
     };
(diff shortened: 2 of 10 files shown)


JPMeehan:
We can keep the AE updates to a separate PR, but I think that makes a lot of sense as an adjustment

src/module/data/actor/npc.mjs
```diff
@@ -200,6 +200,36 @@ export default class NPCModel extends CreatureModel {
 
   /* -------------------------------------------------- */
 
+  /** @inheritdoc */
+  async toEmbed(config, options) {
+    // All NPCs are rendered inline
+    config.inline = true;
```
krbz999, line 206: is this a hard requirement?

src/module/data/actor/npc.mjs
```diff
@@ -200,6 +200,36 @@ export default class NPCModel extends CreatureModel {
...
+
+    const walkLabel = _loc(CONFIG.Token.movement.actions.walk.label);
+    const walkRe = new RegExp(`${walkLabel}[,]?`);
+
+    const context = {
+      actor: this.parent,
+      characteristics: this.parent.sheet._getCharacteristics(true),
```
krbz999: We need to move this out of the sheet class, the sheet and the embed cannot be connected.

Moving `_getCharacteristics` to the `npc` model makes perfect sense.

src/module/data/actor/npc.mjs
```diff
@@ -200,6 +200,36 @@ export default class NPCModel extends CreatureModel {
...
+      damageIW: this.parent.sheet._getImmunitiesWeaknesses(),
+      monsterKeywords: this.parent.sheet._getMonsterKeywords().join(", "),
+      movement: this.parent.sheet._getMovement().list.replace(walkRe, "").trim(),
```
krbz999: same here

src/styles/system/applications/components/embeds.css
```diff
@@ -45,6 +45,152 @@ document-embed, .item-embed {
...
+        }
+
+        h2 {
+          font-size: 1.125em;
+          margin: 0;
+
+          a {
+            color: unset !important;
+
+          }
```
krbz999, line 82:
```suggestion
            color: unset !important;
          }
```

src/styles/system/applications/components/embeds.css
```diff
@@ -45,6 +45,152 @@ document-embed, .item-embed {
...
+              font-weight: normal;
+              letter-spacing: 1px;
+            }
+
+
+
+        }
+
+      }
+    }
```
krbz999, line 153:
```suggestion
            }
        }
      }
```

src/styles/system/applications/components/embeds.css
```diff
@@ -45,6 +45,152 @@ document-embed, .item-embed {
...
+          align-items: center;
+          justify-content: center;
+          gap: 0.5ch;
+
+
+          .label::first-letter {
+              font-family: 'Draw Steel Glyphs';
+              font-weight: normal;
+              letter-spacing: 1px;
+            }
```
krbz999:
```suggestion
          .label::first-letter {
            font-family: 'Draw Steel Glyphs';
            font-weight: normal;
            letter-spacing: 1px;
          }
```

src/styles/system/applications/components/embeds.css
```diff
@@ -45,6 +45,152 @@ document-embed, .item-embed {
...
+
+    &.mount {
+      --color-statblock: #BADDEA;
+    }
+
+    &.support {
+      --color-statblock: #F4D8C2;
+    }
+
+  }
```
krbz999, line 190:
```suggestion
    }
  }
```

templates/embeds/actor/npc.hbs
```diff
@@ -0,0 +1,107 @@
...
+    </div>
+    <div class="movement">
+      <span class="label">{{systemFields.movement.label}}:</span>
+      <span class="value">
+        {{#if movement}}
+        {{movement}}
+        {{else}}
+        —
+        {{/if}}
+      </span>
```
krbz999:
```suggestion
      <span class="value">{{ifThen movement movement "—"}}</span>
```

krbz999 (changes requested):
Quick Sunday drive-by review prior to testing.

Just a few thoughts and some linting issues with the css.

src/module/data/actor/npc.mjs
JPMeehan: This was my addition, it's what we use for the Ability embeds. I don't think Foundry's standard embed structure makes a lot of sense for NPCs.

src/module/data/actor/npc.mjs
JPMeehan: Yeah, and shouldn't be any problem with the other PRs since they're operating in different spaces

src/module/data/actor/npc.mjs
jalensailin: I've addressed these by moving the relevant functions to their  associated DataModels. Everything works as far as I can tell, but I would appreciate someone double checking since I am not as familiar with this system.

src/styles/system/applications/components/embeds.css
jalensailin: Not sure why my formatter isn't running for css files, I'll have to look into it. Thanks for the catch

src/module/data/actor/npc.mjs
jalensailin: I've addressed these by moving the relevant functions to their associated DataModels. Everything works as far as I can tell, but I would appreciate someone double checking since I am not as familiar with this system.

jalensailin:
> We can keep the AE updates to a separate PR, but I think that makes a lot of sense as an adjustment

Sounds good, I'll make an issue

src/module/data/actor/base-actor.mjs
```diff
@@ -453,4 +453,71 @@ export default class BaseActorModel extends DrawSteelSystemModel {
   async updateResource(delta) {
     throw new Error("This method is abstract and must be implemented by a subclass");
   }
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Constructs an object with the formatted immunities and weaknesses with a list of damage labels.
+   * @returns {{immunities: string, weaknesses: string, labels: Record<string, string>}}
+   * @protected
```
JPMeehan: I guess these aren't *technically* protected because we are calling them externally, but the underscore is correct as a "you don't usually need to call this"

JPMeehan (approved):
I think this does a great job establishing a baseline of what the embed here will look like, with room to work in the features and abilities afterwards.

src/module/data/actor/base-actor.mjs
jalensailin: Good catch, thank you. Fixed in recent commit

krbz999:
Putting these side by side,

image

I'm having a few concerns:
- Tags missing in the header (I admit I didnt check whether it was just entirely missing in the actor data).
- The general size of the html makes it very easy for an embed to go out of bounds as seen here with a journal entry whose sheet has been adjusted only a little bit. We might have to consider smaller fonts (not by much tho) and the ability to have multiple columns.

src/module/data/actor/npc.mjs
```diff
@@ -200,6 +200,36 @@ export default class NPCModel extends CreatureModel {
...
+
+    const embed = document.createElement("div");
+
+    embed.classList.add("draw-steel", "actor", "npc", this.monster.role || "no-role");
+
+    embed.innerHTML = await foundry.applications.handlebars.renderTemplate(systemPath("templates/embeds/actor/npc.hbs"), context);
+
+    return embed;
+
+  }
```
krbz999, line 228:
```suggestion
    return embed;
  }
```

src/module/data/actor/npc.mjs
```diff
@@ -276,4 +306,15 @@ export default class NPCModel extends CreatureModel {
...
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Fetches the printable string for the monster's keywords.
+   * @returns {string[]}
+   */
+  _getMonsterKeywords() {
+    const monsterKeywords = ds.CONFIG.monsters.keywords;
+    return Array.from(this.monster.keywords).map(k => monsterKeywords[k]?.label).filter(k => k);
```
krbz999:
```suggestion
    return Array.from(this.monster.keywords).map(k => monsterKeywords[k]?.label).filter(_ => _);
```

src/styles/system/applications/components/embeds.css
```diff
@@ -45,6 +45,146 @@ document-embed, .item-embed {
...
+          }
+
+          a:hover {
+            text-decoration: underline;
+          }
+        }
+      }
+
+
+      .level-ev {
```
krbz999, line 90:
```suggestion
      }

      .level-ev {
```

src/styles/system/applications/components/embeds.css
```diff
@@ -45,6 +45,146 @@ document-embed, .item-embed {
...
+
+    background-image: linear-gradient(var(--color-statblock), transparent 5rem);
+    border-radius: 8px 8px 0px 0px ;
+
+    header {
+      display: flex;
+      justify-content: space-between;
+      gap: 8px;
+
+      font-size: 1.25em;
```
krbz999:
```suggestion
      gap: 8px;
      font-size: 1.25em;
```

src/styles/system/applications/components/embeds.css
```diff
@@ -45,6 +45,146 @@ document-embed, .item-embed {
...
+      font-family: 'Newsreader', serif;
+
+      border-bottom: 1px solid gray;
```
krbz999, line 63:
```suggestion
      font-family: 'Newsreader', serif;
      border-bottom: 1px solid gray;
```

src/styles/system/applications/components/embeds.css
```diff
@@ -45,6 +45,146 @@ document-embed, .item-embed {
...
+        font-weight: 600;
+      }
+
+      .basic-stats {
+        gap: 8px;
+
+        font-size: 1.125rem;
+        text-align: center;
+
+        margin-block: 8px 0;
```
krbz999, line 112:
```suggestion
        gap: 8px;
        font-size: 1.125rem;
        text-align: center;
        margin-block: 8px 0;
```

src/styles/system/applications/components/embeds.css
```diff
@@ -45,6 +45,146 @@ document-embed, .item-embed {
...
+        }
+      }
+
+      .secondary-stats {
+        display: grid;
+        grid-template-columns: repeat(2, 1fr);
+        align-items: center;
+        font-size: 1.125em;
+
+        padding-inline: 8px;
```
krbz999:
```suggestion
        font-size: 1.125em;
        padding-inline: 8px;
```

src/styles/system/applications/components/embeds.css
```diff
@@ -45,6 +45,146 @@ document-embed, .item-embed {
...
+        border-block: 1px solid gray;
+
+        .characteristic {
+          display: flex;
+          align-items: center;
+          justify-content: center;
+          gap: 0.5ch;
+
+
+          .label::first-letter {
```
krbz999, line 146:
```suggestion
          gap: 0.5ch;

          .label::first-letter {
```

templates/embeds/actor/npc.hbs
```diff
@@ -0,0 +1,101 @@
+<header>
+  <div class="flexrow image-name-keywords">
+    <img src="{{actor.img}}" alt="{{actor.name}}">
+    <div class="flexcol name-keywords">
+      <h2 class="name">
+        <a data-link data-uuid="{{actor.uuid}}">{{actor.name}}</a>
+      </h2>
```
krbz999, line 7: I think this should be a click event listener attached in the `onRender` method (or whatever the equivalent is for embeds, I do not remember).

templates/embeds/actor/npc.hbs
```diff
@@ -0,0 +1,101 @@
...
+        {{chr.field.label}}
+      </div>
+      <div class="value">
+        {{numberFormat chr.value sign=true}}
+      </div>
+    </div>
+    {{/each}}
+  </div>
+
+</section>
```
krbz999, line 100:
```suggestion
  </div>
</section>
```

krbz999 (changes requested):

krbz999:
In addition to above comment and review, I wonder if some of this sheet context and embed context prep can be moved into the data model's own data prep.

That doesn't have to be part of this PR, however.

JPMeehan:
Tag issue is the actor data (we just never actually added all the minor tags)

jalensailin:
>  The general size of the html makes it very easy for an embed to go out of bounds as seen here with a journal entry whose sheet has been adjusted only a little bit. We might have to consider smaller fonts (not by much tho) and the ability to have multiple columns.

Font size smaller: 

image

---

`@Embed` width is now dynamic, here's how it looks at the journal's smallest width: 

image

templates/embeds/actor/npc.hbs
jalensailin: Done in recent commit via `onEmbed` method on the `NPCModel`

krbz999 (approved):

PR #1891, opened 2026-05-01
[#475] Added new pseudodocumentfor Special effect texts

### LINKED ISSUE #329: Create option to ass 2nd and 3rd 'Spend resource' fields to abilities
I'm sure there is other examples, but for example the Ashen Hoarder has an ability that allows it to spend 2 OR 3 Malice with different effects:

Image

Please add the option to add multiple 'spend resource' effects to abilities, like tier effects

### LINKED ISSUE #475: Account for "Special" ability entries
Some classes have a special ability effect line that shares a keyword that needs to be accounted for somehow.

Examples:
* **Persistent X** - the Elementalist has many abilities that can last through multiple turns with a variable cost in Essence. (This may also need to be accounted for in start of turn resource gains, but maybe that's another ticket.)
* **Strained** - the Talent has effects that take place if they have negative Clarity.

These are all of the ones I found in classes so far. However, it might be good to allow for custom versions of these for homebrew or future classes.

JPMeehan:
Implements a new pseudodocument named "SpecialEffect" that has a `description` HTML field. These can be placed before or after the power roll and cover a plain "Effect/Special" text entry, "Spend", "Persistent", and "Strained". A migration has been provided to turn `effect.before`, `effect.after`, and `spend` fields into SpecialEffects, however a one-time migration will be desirable because this was not a simple field-to-field transfer.

image

Closes #475

### DIFF
--- a/src/module/applications/sheets/pseudo-documents/special-effect-sheet.mjs
+++ b/src/module/applications/sheets/pseudo-documents/special-effect-sheet.mjs
@@ -14,7 +14,7 @@ export default class SpecialEffectSheet extends PseudoDocumentSheet {
     details: {
       template: systemPath("templates/sheets/pseudo-documents/special-effect/content.hbs"),
       // Modules can push to this or otherwise load their details partial templates
-      templates: ["spend.hbs"].map(t => systemPath(`templates/sheets/pseudo-documents/special-effect/${t}`)),
+      templates: ["persistent.hbs", "spend.hbs"].map(t => systemPath(`templates/sheets/pseudo-documents/special-effect/${t}`)),
     },
   };
 
--- a/src/module/applications/sheets/pseudo-documents/special-effect-sheet.mjs
+++ b/src/module/applications/sheets/pseudo-documents/special-effect-sheet.mjs
@@ -40,12 +40,8 @@ export default class SpecialEffectSheet extends PseudoDocumentSheet {
    * @returns {Promise<object>}     Mutated rendering context.
    */
   async #prepareDetailsContext(context, options) {
-
     context.detailsPartial = this.pseudoDocument.detailsPartial;
-
     context.ctx = await this.pseudoDocument.getSheetContext(options);
-
     return context;
   }
-
 }
--- a/src/module/applications/sheets/pseudo-documents/special-effect-sheet.mjs
+++ b/src/module/applications/sheets/pseudo-documents/special-effect-sheet.mjs
@@ -1,0 +1,51 @@
+import PseudoDocumentSheet from "../../api/pseudo-document-sheet.mjs";
+import { systemPath } from "../../../constants.mjs";
+
+export default class SpecialEffectSheet extends PseudoDocumentSheet {
+  /** @inheritdoc */
+  static DEFAULT_OPTIONS = {
+    classes: ["special-effect"],
+  };
+
+  /* -------------------------------------------------- */
+
+  /** @inheritdoc */
+  static PARTS = {
+    details: {
+      template: systemPath("templates/sheets/pseudo-documents/special-effect/content.hbs"),
+      // Modules can push to this or otherwise load their details partial templates
+      templates: ["spend.hbs"].map(t => systemPath(`templates/sheets/pseudo-documents/special-effect/${t}`)),
+    },
+  };
+
+  /* -------------------------------------------------- */
+
+  /** @inheritdoc */
+  async _preparePartContext(partId, context, options) {
+    context = await super._preparePartContext(partId, context, options);
+
+    switch (partId) {
+      case "details":
+        return this.#prepareDetailsContext(context, options);
+    }
+
+    return context;
+  }
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Prepare context for the details tab.
+   * @param {object} context    Rendering context.
+   * @returns {Promise<object>}     Mutated rendering context.
+   */
+  async #prepareDetailsContext(context, options) {
+
+    context.detailsPartial = this.pseudoDocument.detailsPartial;
+
+    context.ctx = await this.pseudoDocument.getSheetContext(options);
+
+    return context;
+  }
+
+}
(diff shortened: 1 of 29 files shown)


src/module/applications/sheets/pseudo-documents/special-effect-sheet.mjs
```diff
@@ -0,0 +1,51 @@
...
+  async #prepareDetailsContext(context, options) {
+
+    context.detailsPartial = this.pseudoDocument.detailsPartial;
+
+    context.ctx = await this.pseudoDocument.getSheetContext(options);
+
+    return context;
+  }
+
+}
```
krbz999, line 47:
```suggestion
  async #prepareDetailsContext(context, options) {
    context.detailsPartial = this.pseudoDocument.detailsPartial;
    context.ctx = await this.pseudoDocument.getSheetContext(options);
    return context;
  }
}
```

Bit too much room to breathe.

src/module/data/pseudo-documents/special-effect/base-special-effect.mjs
```diff
@@ -0,0 +1,72 @@
...
+  /** @inheritdoc */
+  static LOCALIZATION_PREFIXES = super.LOCALIZATION_PREFIXES.concat("DRAW_STEEL.SPECIAL_EFFECT");
+
+  /* -------------------------------------------------- */
+
+  /** @inheritdoc */
+  static defineSchema() {
+    return Object.assign(super.defineSchema(), {
+      description: new HTMLField(),
+      before: new BooleanField(),
```
krbz999, line 37: We don't want this to be `initial: true`?

src/module/data/pseudo-documents/special-effect/spend-effect.mjs
```diff
@@ -0,0 +1,52 @@
...
+  get detailsPartial() {
+    return systemPath("templates/sheets/pseudo-documents/special-effect/spend.hbs");
+  }
+
+  /* -------------------------------------------------- */
+
+  /** @inheritdoc */
+  get label() {
+    return _loc(`DRAW_STEEL.SPECIAL_EFFECT.spend.${this.resource.multiple ? "multiple" : "single"}`, { value: this.resource.value, resourceName: this.parent.resourceName });
+  }
```
krbz999, line 44:
```suggestion
  get label() {
    return _loc(`DRAW_STEEL.SPECIAL_EFFECT.spend.${this.resource.multiple ? "multiple" : "single"}`,
      { value: this.resource.value, resourceName: this.parent.resourceName },
    );
  }
```

src/module/config.mjs
```diff
@@ -1883,6 +1883,43 @@ preLocalize("MessagePart", { key: "label" });
 
 /* -------------------------------------------------- */
 
+/**
+ * @typedef SpecialEffectType
+ * @property {string} label                                                 Human-readable label.
+ * @property {string} defaultImage                                          The default image for PowerRollEffects of this type.
+ * @property {pseudoDocuments.specialEffects.BaseSpecialEffect} documentClass   The pseudo-document class.
+ */
```
krbz999, line 1891:
```suggestion
 * @typedef SpecialEffectType
 * @property {string} label                                                     Human-readable label.
 * @property {string} defaultImage                                              The default image for PowerRollEffects of this type.
 * @property {pseudoDocuments.specialEffects.BaseSpecialEffect} documentClass   The pseudo-document class.
 */
```

templates/sheets/pseudo-documents/special-effect/spend.hbs
```diff
@@ -0,0 +1,5 @@
+
+<div class="form-group">
+  {{formGroup fields.resource.fields.value value=source.resource.value name="resource.value" label=ctx.resourceName classes="slim"}}
+  {{formGroup fields.resource.fields.multiple value=source.resource.multiple name="resource.multiple"}}
+</div>
```
krbz999:
```suggestion
<div class="form-group">
  {{formGroup fields.resource.fields.value value=source.resource.value name="resource.value" label=ctx.resourceName classes="slim"}}
  {{formGroup fields.resource.fields.multiple value=source.resource.multiple name="resource.multiple"}}
</div>
```

templates/sheets/pseudo-documents/special-effect/identity.hbs
```diff
@@ -0,0 +1,6 @@
+<section class="{{tabs.identity.cssClass}}" data-tab="{{tabs.identity.id}}" data-group="{{tabs.identity.group}}">
+  {{formGroup fields.name value=source.name placeholder=(localize (concat "TYPES.SpecialEffect." source.type)) name="name"}}
```
krbz999, line 2: I would think that we can get at least the placeholder automatically derived using `FIELDS`, or do you not think that is worth it?

krbz999 (changes requested):
Very impressed by the migration being seamless, it involved no effort on my part in getting Healing Grace up to speed for example.

Code mostly looks good, and a follow-up PR should migrate our system packs.

As a final note, I think the lesson here is that we were too quick in implementing these properties that we have now migrated. If we had held off, perhaps just left them in a free-form HTML field, there'd be no migration to speak of.

src/module/data/pseudo-documents/special-effect/base-special-effect.mjs
JPMeehan: Effects being listed after the power roll is more common than before by a fairly large margin.

templates/sheets/pseudo-documents/special-effect/identity.hbs
JPMeehan: Right now we actually don't have separate LOCALIZATION_PREFIXES set up, they're all just kinda jammed together in one `FIELDS`. Could definitely split up/make subgroups to extend if you'd prefer.

JPMeehan:
I'll note that what we currently have basically *is* a (pair of) free-form HTML fields, the only notable bit is the current `spend` field but it's still a fairly trivial migration.

krbz999 (approved):

Issue #1888, opened 2026-04-28
Implement "With Captain" section of NPC Embed


jalensailin:
The "With Captain" section in the NPC Embed is currently always null. This is implemented as an AE on the Actor, and thus it's a little difficult to translate into human readable format without some hacky work pulling the labels from the data model.

I propose that we add the human-readable text to the AE's description. For example, if the AE gives a +2 bonus to system.stamina.max, we would simply add the description "+2 bonus to Stamina".

JPMeehan:
How were you currently planning to ID the effects? We have a separate ticket for adding a property to AEs for this as part of automating suppressing the effect when there's no captain.

I agree the text updates make perfect sense

jalensailin:
I think that makes sense, a Boolean property on the AE that says `Is "With Captain" Effect?` The field would be disabled if another AE on the actor already has the "With Captain" setting `true`. 

Also tagging the related issue so it's easy to find: #348

Issue #348, opened 2025-03-24
Implement "With Captain" condition for Active Effects


KJTailor:
There should be a feature type for minions that captures 'with captain' bonuses and applies them similar to how kits do for player characters.

Maybe with a tick box that on the minion sheet that applies these bonuses when ticked, if a minion squad does have a captain/

JPMeehan:
Captain bonuses can uniformly be handled by AEs, just need this conditional logic to be available for both `base` and `abilityModifier` types.

PR #1892, opened 2026-05-01
Implement "With Captain" effects.

jalensailin:
**Note**: Leaving as draft until #1886 is merged, as this is based off of the branch in that one.

This MR does the following:

- Adds a `system.isWithCaptain` `boolean` field to `BaseEffectModel`. 
  - If this is `true` this will be the Effect from which the Document pulls "With Captain" data.
  - Only one Effect on an NPC can be designated as the "With Captain" effect (the input disappears from the UI when another AE on the same Actor has this marked `true`.
  
     image

  - NPC `@Embeds` fill their "With Captain" section from the `innerHTML` of the first element of the description of the "With Captain" Effect. 
  
     image

- Updates NPCs that were missing descriptions for their "With Captain" Effects.
- Adds a "With Captain" section on the NPC sheet, where the description of the "With Captain" effect is displayed and where the Effect can be directly toggled on/off.
  
  image
  
  - As part of this, slightly adjusted the css for the "Resources" and "Free Strike" sections on the character sheet to make things consistent.

Questions: 

- Do non-NPC actors have "With Captain" effects? Currently everything is built off of the NPC model/sheet but I can move things around as necessary if other Actors would have this effect.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -1811,7 +1811,7 @@
           },
           "isWithCaptain": {
             "label": "\"With Captain\" Effect",
-            "hint": "Apply effect when the Actor is in the presence of their Captain. Only one \"With Captain\" effect can be active at a time."
+            "hint": "Effect to be applied when the Actor is in the presence of their Captain. An Actor may have only one \"With Captain\" effect at a time."
           },
           "project": {
             "label": "Project",
--- a/lang/en.json
+++ b/lang/en.json
@@ -1806,6 +1806,10 @@
               "label": "Saving Throw Formula"
             }
           },
+          "isWithCaptain": {
+            "label": "\"With Captain\" Effect",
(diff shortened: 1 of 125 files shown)


krbz999:
Unsure if there's a discussion somewhere I missed but there are two concerns here:
- This is a boolean field where I vastly prefer `Set<string>`.
- The rendering is dependent on other documents.

JPMeehan:
The context is the "With Captain" benefit displays in the header of Minion stat blocks so we need some reliable way to find that active effect

jalensailin:
> Unsure if there's a discussion somewhere I missed...

The issues are raised in #1888 and #348. Should have linked them in the description. There is also some discussion in #1886

JPMeehan:
Mm yeah I see that this is forked off the NPC embed - we should get that merged first to simplify some of the delta here

jalensailin:
This PR closed in favor of #1902. See note in that PR's description

Issue #702, opened 2025-07-07
Implement Effect Conditionals


JPMeehan:
The Fury's "Growing Ferocity" feature seems to require bespoke handling

> You gain certain benefits in combat based on the amount of ferocity you have (see 1st-Level Aspect Features for details). These benefits last until the end of your turn, even if a benefit would become unavailable to you because of the amount of ferocity you spend during your turn.
> 
> Some Growing Ferocity benefits can be applied only if you are a specific level or higher, with the level of those benefits noted in the various Growing Ferocity tables in this section.

The Null has a similar feature

> As you advance in your chosen null tradition, you gain certain benefits in combat, including benefits based on the amount of discipline you have. Benefits based on how much discipline you have last until the end of your turn, even if a benefit would become unavailable to you because of the amount of discipline you spend during your turn.
> Some Discipline Mastery benefits can be applied only if you are a specific level or higher, with the level of those benefits noted in the tables below.

A JavaScriptField would be one solution, but could also just specifically have `requirements` field, maybe with character level but definitely with Heroic Resource.

KJTailor:
Might be worthwhile to also thinking about Null's accumulated Discipline features in this context.

JPMeehan:
Another use case would be Leveled Treasures

JPMeehan:
Minion "Captain Benefits" would also be a logical use case

JPMeehan:
Possible core solution means this must wait until v14

https://github.com/foundryvtt/foundryvtt/issues/13465

JPMeehan:
Beastheart joins the club.

PR #1902, opened 2026-05-14
Improve "With Captain" implementation

### LINKED ISSUE #1888: Implement "With Captain" section of NPC Embed

jalensailin:
**Note**: Because of the squash of commits in the merge of #1886, #1892 became a headache to rebase. Therefore, this is a new branch with the commits from the original branch cherry-picked. I've copied my original description here.

This MR does the following:

- Adds a `system.monster.withCaptainEffect` `DocumentIdField` to `NPCModel`. 
  - This stores the `id` of the designated "With Captain" Effect.
  - NPC `@Embeds` fill their "With Captain" section from the `enrichedHTML` of the description of the "With Captain" Effect. 
  
     image

- Updates NPCs that were missing descriptions for their "With Captain" Effects.
- Adds a "With Captain" section on the NPC sheet, where the description of the "With Captain" effect is displayed and where the Effect can be directly toggled on/off.
  
image
  
  - As part of this, slightly adjusted the css for the "Resources" and "Free Strike" sections on the character sheet to make things consistent.

Questions: 

- Do non-NPC actors have "With Captain" effects? Currently everything is built off of the NPC model/sheet but I can move things around as necessary if other Actors would have this effect. 

Closes #1888


JPMeehan (changes requested):
OK, two thoughts on review
1. I want this grouped under some kind of `conditions` schema field because we've got others that are going into #702 
2. I'd much rather more automated enable/disable, we have info on squads & captains. Maybe that should be a follow up PR, but it also means I *don't* think a straight up checkbox on the stats tab is correct. It's possible that it's better to rework that `freeStrike` fieldset into a space that supports these.

src/module/data/actor/npc.mjs
```diff
@@ -55,6 +55,7 @@ export default class NPCModel extends CreatureModel {
       level: requiredInteger({ initial: 1 }),
       role: new fields.StringField({ required: true }),
       organization: new fields.StringField({ required: true }),
+      withCaptainEffect: new fields.DocumentIdField({ required: false }),
```
JPMeehan: Needs `readonly: false` too. Also, good candidate for a local document field implementation @krbz999

src/module/data/actor/npc.mjs
```diff
@@ -323,4 +322,19 @@ export default class NPCModel extends CreatureModel {
...
+
+  /* ------------------------------------------------- */
+
+  /**
+   * Fetches the description for the "With Captain" effect.
+   * @returns {string|null} The inner HTML of the first element of the description.
+   *                        Null if no such effect exists.
+   */
+  _getWithCaptainDescription() {
+    const withCaptainAE = this.parent.effects.find(e => e.id === this.monster.withCaptainEffect);
```
JPMeehan: Why `find` instead of `get`?

src/module/data/actor/npc.mjs
```diff
@@ -323,4 +322,19 @@ export default class NPCModel extends CreatureModel {
...
+    if (!withCaptainAE) return null;
+    let html = foundry.utils.parseHTML(withCaptainAE?.description);
```
JPMeehan: Definitely need notes that we don't want to support enrichment if we're doing this

krbz999: No, this should be using `CONFIG.ux.TextEditor.enrichHTML`. The surrounding method is async, so that will be fine.

templates/sheets/actor/npc-sheet/stats.hbs
```diff
@@ -76,6 +76,15 @@
         {{/if}}
       </div>
     </fieldset>
+    {{#if withCaptain.exists}}
+    <fieldset class="with-captain">
+      <legend>{{localize "DRAW_STEEL.ActiveEffect.WithCaptainEffects.embedLabel"}}</legend>
+      <div>
+        <input type="checkbox" data-action="toggleWithCaptainEffect" data-effect-id="{{withCaptain.effectId}}" id="{{concat rootId "-withCaptainToggle"}}" {{checked withCaptain.effectEnabled}}>
```
krbz999: Not assigning click event listeners to checkboxes. This should be using `_onChangeForm`. Alternatively, make it a button.

src/module/data/actor/npc.mjs
jalensailin: Updated to use `TextEditor.enrichHTML`

src/module/data/actor/npc.mjs
jalensailin: Fixed in recent commit

src/module/data/actor/npc.mjs
jalensailin: Added `readonly: false` in recent commit

templates/sheets/actor/npc-sheet/stats.hbs
jalensailin: I made this a button in a recent commit. However, this still doesn't address the above comment from @JPMeehan:

> I'd much rather more automated enable/disable, we have info on squads & captains. Maybe that should be a follow up PR, but it also means I don't think a straight up checkbox on the stats tab is correct. It's possible that it's better to rework that freeStrike fieldset into a space that supports these.

This is how it looks currently:

image

Do you want me to incorporate the button into the freeStrike `fieldset` rather than its own space for now? And leave the higher level of automation for a future MR?

JPMeehan: Yeah, we've got some styling for buttons in the fieldset; make sure to conditionally hide the button based on if the sheet is editable.

src/module/data/actor/npc.mjs
```diff
@@ -349,4 +348,17 @@ export default class NPCModel extends CreatureModel {
     const monsterKeywords = ds.CONFIG.monsters.keywords;
     return Array.from(this.monster.keywords).map(k => monsterKeywords[k]?.label).filter(_ => _);
   }
+
+  /* ------------------------------------------------- */
+
+  /**
+   * Fetches the description for the "With Captain" effect.
+   * @returns {string|null} The inner HTML of the first element of the description.
```
JPMeehan:
```suggestion
   * @returns {Promise<string|null>} The inner HTML of the first element of the description.
```

src/module/data/actor/npc.mjs
```diff
@@ -349,4 +348,17 @@ export default class NPCModel extends CreatureModel {
...
+   *                        Null if no such effect exists.
+   */
+  _getWithCaptainDescription() {
+    const withCaptainAE = this.parent.effects.get(this.monster.withCaptainEffect);
+    if (!withCaptainAE) return null;
+    return CONFIG.ux.TextEditor.enrichHTML(withCaptainAE.description);
```
JPMeehan: 1. May as well just mark it async
2. We have an `enrichHTML` wrapper method that allows you just to pass the `relativeTo` argument

src/module/data/actor/npc.mjs
```diff
@@ -57,6 +57,7 @@ export default class NPCModel extends CreatureModel {
       level: requiredInteger({ initial: 1 }),
       role: new fields.StringField({ required: true }),
       organization: new fields.StringField({ required: true }),
+      withCaptainEffect: new fields.DocumentIdField({ required: false, readonly: false }),
```
JPMeehan: LocalDocumentField got merged so you can grab that. Adjust access accordingly.

templates/sheets/actor/npc-sheet/stats.hbs
jalensailin: This should be addressed now:

**Play Mode:**

image

**Edit Mode:**

image

**Disabled:**

image


**Player View:**

image

src/module/data/actor/npc.mjs
jalensailin: Fixed in recent commits

src/module/data/actor/npc.mjs
jalensailin: Resolved in recent commit

src/module/data/actor/npc.mjs
jalensailin: Resolved in recent commit

src/module/data/actor/npc.mjs
jalensailin: This is resolved

src/module/applications/sheets/npc-sheet.mjs
```diff
@@ -189,4 +210,18 @@ export default class DrawSteelNPCSheet extends DrawSteelActorSheet {
...
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Toggle the "With Captain" effect.
+   * @this DrawSteelNPCSheet
+   * @param {PointerEvent} event   The originating click event.
+   * @param {HTMLElement} target   The capturing HTML element which defined a [data-action].
+   */
+  static #toggleWithCaptainEffect(event, target) {
```
JPMeehan: mark it async and await the update, on the off chance someone is calling this directly

templates/apps/document-input/monster-metadata-input.hbs
```diff
@@ -4,4 +4,5 @@
   {{formGroup monsterFields.organization value=systemSource.monster.organization options=monsterOrganizations}}
   {{formGroup monsterFields.role value=systemSource.monster.role options=monsterRoles}}
   {{formGroup systemFields.ev value=systemSource.ev}}
+  {{formGroup monsterFields.withCaptainEffect value=systemSource.monster.withCaptainEffect options=effects}}
```
JPMeehan, line 8: With captain should only display if the role is minion. Shouldn't need any fancy listeners since this is a live-updating form that re-renders itself, just an `if (eq)` on the `systemSource.monster.organization` value.

templates/embeds/actor/npc.hbs
```diff
@@ -81,7 +81,8 @@
       <span class="value">{{ifThen movement movement "—"}}</span>
     </div>
     <div class="with-captain">
-      <span class="label">{{localize "With Captain"}}:</span>
+      <span class="label">{{localize "DRAW_STEEL.ActiveEffect.WithCaptainEffects.embedLabel"}}:</span>
+      <span class="value">{{{ifThen withCaptain withCaptain "—"}}}</span>
     </div>
```
JPMeehan, line 73: Similarly should also only show if this is a minion.

templates/sheets/actor/npc-sheet/stats.hbs
```diff
@@ -57,25 +57,44 @@
...
       </div>
+      {{#if (and isPlay withCaptain.exists)}}
+      <div class="with-captain">
+        <button id="{{concat rootId "-withCaptainToggle"}}" data-action="toggleWithCaptainEffect" data-effect-id="{{withCaptain.effectId}}">
+          {{localize "DRAW_STEEL.ActiveEffect.WithCaptainEffects.embedLabel"}}
+          <i class="fa-solid fa-lg {{ifThen withCaptain.effectEnabled "fa-toggle-on" "fa-toggle-off"}}"></i>
+        </button>
+        <label for="{{concat rootId "-withCaptainToggle"}}">{{{withCaptain.description}}}</label>
+      </div>
+      {{/if}}
```
JPMeehan, line 97: Maybe have the withCaptain.exists check also confirm that this is a minion?

src/module/data/actor/npc.mjs
```diff
@@ -349,4 +350,18 @@ export default class NPCModel extends CreatureModel {
...
+  /**
+   * Fetches the description for the "With Captain" effect.
+   * @param {object} [options] Options to forward to the TextEditor.enrichHTML method.
+   * @returns {Promise<string|null>} The inner HTML of the first element of the description.
+   *                                 Null if no such effect exists.
+   */
+  async _getWithCaptainDescription(options = {}) {
+    const { withCaptainEffect } = this.monster;
+    if (!withCaptainEffect) return null;
+    return CONFIG.ux.TextEditor.enrichHTML(withCaptainEffect.description, { ...options, relativeTo: this.parent });
```
JPMeehan: Helper enrichHTML method is from `src/module/utils/enrich-html.mjs`, nothing fancy just handles secrets and the like. The `relativeTo` should also be the AE itself rather than the item (we implemented `ActiveEffect#getRollData`)

JPMeehan (commented):
95% of the way there just a few quick adjustments

src/module/data/actor/npc.mjs
jalensailin: Should be addressed in recent commit

templates/sheets/actor/npc-sheet/stats.hbs
jalensailin: Addressed in recent commit

templates/embeds/actor/npc.hbs
jalensailin: Addressed in recent commit

templates/apps/document-input/monster-metadata-input.hbs
jalensailin: Addressed in recent commit

src/module/applications/sheets/npc-sheet.mjs
jalensailin: Addressed in recent commit

JPMeehan (approved):

Issue #1872, opened 2026-04-20
Minion healthbar very inconsistently updating


nelizzy:
### What happened?

<img width="1668" height="781" alt="Image" src="https://github.com/user-attachments/assets/4c44bdc1-e5e6-451b-8d3b-1d5c69c904f3" />

When updating the health of a minion group, only the tokens of minions that have a condition change seem to update (the most typical being defeated, but any condition i.e. bleeding, prone, etc. will trigger it). Even then, it seems inconsistent, sometimes not triggering even when marked defeated.

I expect when a minion group takes damage, regardless of if it's enough to kill a minion or not, _all_ tokens in minion group will update to show the new minion group heatlh.

### Reproduction Steps

1. Create a combatant group with minions.
2. Apply damage to minion group (either via ability or editing token health).
3. If proccing minion death, only select a few minions.
4. For extra testing, toggle defeat of minions in combat tracker and/or apply conditions via token condition button.

Tested in Portable Windows Foundry GUI and Firefox.

### What core and system version are you reporting this for?

FVTT 14.360, Draw Steel 1.0.1

### Bug Checklist

- [x] The issue occurs while all Modules are disabled
- [ ] I am using The Forge for hosting

nelizzy:
In local testing, it looks like the token's registered bar data isn't actually keeping up to date with `getBarAttribute()` on damage application. Triggering a `token._prepareBars()` and _then_ doing a bar refresh works (though bar animation doesn't trigger.)

Additionally, on loading into foundry, a minion's bar1 is still set to its own stamina rather than the combat group stamina. It looks like the `combatGroup.size` logic fails because `combatGroup` isn't even available at the time that it's called.

JPMeehan:
Are you testing in 14.360 or 14.361? I believe there were some changes to bar logic between the versions

nelizzy:
Currently testing in 14.361

nelizzy:
Noodling through code to try and get something working. At least for animating changes and getting bar1 to properly update, editing the forloop in`refreshSquad()` code gets there:
```js
    for (const combatant of this.parent.members) {
      const barData = (barName) => {
        const data = combatant.token.getBarAttribute(barName);
        const value = data?.value ?? null;
        const max = data?.max ?? null;
        return { value, max };
      };

      const to = { bar1: barData("bar1"), bar2: barData("bar2") };

      combatant.token.object.animate(to);

      combatant.refreshCombatant();
    }
```

[...]

PR #1903, opened 2026-05-14
Update minion token bars to combat group health

nelizzy:
- Properly animates minions health bars
- Updates bar attributes when drawing tokens for first time (not sure if token.system.combatGroup can be found any sooner than that?)

Solves #1872

### DIFF
--- a/src/module/canvas/placeables/token.mjs
+++ b/src/module/canvas/placeables/token.mjs
@@ -184,11 +184,12 @@ export default class DrawSteelToken extends foundry.canvas.placeables.Token {
 
   /** @inheritdoc */
   _draw(options = {}) {
+    super._draw(options);
+
     if (this.actor.system.combatGroups.size === 1) {
       this.document._prepareBars();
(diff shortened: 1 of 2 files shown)


JPMeehan:
Looking at the upcoming releases, it seems that 14.362 may have some adjustments for bar logic we'll need to react to.

JPMeehan:
Could you revisit your analysis and see how the changes in 14.362 impact the fix?

nelizzy:
Fix continues to work in 14.362, and problem continues to persist _without_ fix.

I did notice my fix doesn't cover the healthbars on the ghost token for planning movement. I'm struggling to determine what to target in that case.
2026-05-21_Foundry_Virtual_Tabletop_OJlrorlpa3

krbz999:
I think we should be able to properly resolve this with additional changes to `DrawSteelToken#_drawBar`.

JPMeehan:
This is better implemented as a fix to _drawBar — we're just obviously failing to account for minion stamina

PR #1905, opened 2026-05-15
[#1904] Fix choices not being limited in trait advancements.

### LINKED ISSUE #1904: `TraitAdvancement` configuration is not limiting the choices
### What happened?

When selecting the skills or languages you want to select from a `TraitAdvancement` the options are not becoming disabled after the max choices have been selected.

This is happening because Foundry is refreshing the disabled status after the change event completes, so our setting of disabled is getting removed.

### Reproduction Steps

Drop a culture onto the character sheet and then try selecting a skill and you'll see you can select as many skills as you want.

_No response_

### Bug Checklist

cdverrett94:
Closes #1904 

I will note that everytime `disableOption` is called, the disabled status of all checkboxes is refreshed which means we're looping through the checkboxes n^2 times. We could only loop through only once, but it would require manually updating the the protected property `_disabledOptions`. Not sure how important this is on something that's not happening regularly.

### DIFF
--- a/src/module/data/pseudo-documents/advancements/trait-advancement.mjs
+++ b/src/module/data/pseudo-documents/advancements/trait-advancement.mjs
@@ -116,12 +116,13 @@ export default class TraitAdvancement extends BaseAdvancement {
       const submit = dialog.element.querySelector(".form-footer [type=submit]");
       multiCheckbox.addEventListener("change", () => {
         for (const checkbox of multiCheckbox.querySelectorAll("input")) {
-          const disabled = !multiCheckbox.value.includes(checkbox.value) && (multiCheckbox.value.length >= chooseN);
-          multiCheckbox.disableOption(checkbox.value, disabled);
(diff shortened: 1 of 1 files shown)


JPMeehan:
Is this some weird v14 change? O(n^2) is certainly not ideal, but also the size of the list we have means that I'm sure performance is negligible.

cdverrett94:
I had to go pull up a v13 build to see the differences. Yeah, this is a change in v14. Now in v14 the `HTMLMultiCheckboxElement#_refresh` also sets the disabled property on each checkbox based on the values in the Set `HTMLMultiCheckboxElement#_disabledOptions`. In v13 `_refresh` only modified the checked property.

JPMeehan:
Hmm, maybe we should be interacting with that property? It's protected not private

cdverrett94:
Yeah, that was the suggestion in the original comment, I just wasn't sure if we were meant to be messing with protected stuff or not. I can update to do that.

JPMeehan:
Yeah would prefer that I think

cdverrett94:
Okay, swapped to using the protected method instead.

JPMeehan (approved):

PR #1909, opened 2026-05-20
[#509] Implemented Potency Check Method

### LINKED ISSUE #509: Add data path for "Characteristic counts as x higher for resisting potencies"
The dwarven trait "Stand Tough" allows them to treat their "Might as 1 higher for resisting potencies". This should at least be targetable with an Active Effect, if not directly part of the DB-backed schema.

JPMeehan:
Adds `chr.resist` as a data field for handling when a characteristic should be treated as lower/higher for potency purposes as well as a `checkPotency` method.

Does not implement spending surges to increase effective potency, leaving that feature for a follow-up PR

Closes #509


src/module/data/actor/creature.mjs
```diff
@@ -208,4 +208,19 @@ export default class CreatureModel extends BaseActorModel {
       return obj;
     }, {});
   }
+
+  /* ------------------------------------------------- */
+
+  /**
+   * Checks if this character passes a potency check.
```
cdverrett94: Probably need to update this to reflect that it's actually checking if the character fails and is affected by a potency.

src/module/data/actor/creature.mjs
```diff
@@ -208,4 +208,19 @@ export default class CreatureModel extends BaseActorModel {
...
+
+  /* ------------------------------------------------- */
+
+  /**
+   * Checks if this character fails a potency check.
+   * @param {number} value
+   * @param {string} characteristic A valid characteristic key.
+   * @returns {boolean} Returns true if the value is greater than this character's effective characteristic.
+   */
+  checkPotency(value, characteristic) {
```
krbz999, line 220: Suggest a name like "evaluatePotencyResistance" that is more descriptive.

Also feel like it should be inverted.

src/module/data/pseudo-documents/power-roll-effects/base-power-roll-effect.mjs
```diff
@@ -167,6 +167,12 @@ export default class BasePowerRollEffect extends TypedPseudoDocument {
    */
   static constructPotencyHTML(characteristic, strength) {
 
+    const span = document.createElement("span");
+    if (!characteristic) return span;
+    span.classList.add("potency");
+    span.dataset.potencyValue = strength;
+    span.dataset.potencyCharacteristic = Object.entries(ds.CONFIG.characteristics).find(([,chr]) => chr.rollKey === characteristic).at(0);
```
krbz999:
```suggestion
    span.dataset.potencyCharacteristic = Object.entries(ds.CONFIG.characteristics).find(([, chr]) => chr.rollKey === characteristic).at(0);
```

src/module/documents/actor.mjs
```diff
@@ -85,6 +85,19 @@ export default class DrawSteelActor extends BaseDocumentMixin(foundry.documents.
 
   /* -------------------------------------------------- */
 
+  /**
+   * A helper function to check if a potency applies to this actor.
+   * @param {number} value
+   * @param {string} characteristic
+   * @returns {boolean} True if the value exceeds the relevant characteristic, null if this actor does not have characteristics to evaluate.
+   */
+  checkPotency(value, characteristic) {
```
krbz999, line 94: Little value (for now) in adding this to the document class imo.

krbz999 (commented):
you got an ability to test this with?

JPMeehan:
We've got a lot of abilities with potencies already set up, but just for the two examples

"Concussive Strike" (level 1 Tactician 3 Focus ability) as part of a power roll effect

"To Me Squad!" (level 6 Tactician 9 Focus ability) where it's an enricher in the descriptive text

krbz999 (changes requested):
OK, it took me a good minute or more to figure out what I was actually meant to be seeing here.

This just seems incredibly redundant and near-useless with how difficult it is to even find, given the size of this feature itself which doesn't really do much.

I think this is something we should just leave for the eventual chat message overhaul. That way we also don't have to carry forward a bunch of features that we're likely to forget about.

There are some fine changes here for the data but the UI changes are unnecessary bloat.

JPMeehan:
So, basically just do 509 and add `resist` as a key + the function to evaluate?

JPMeehan:
@krbz999 could use another review here

krbz999:
image

JPMeehan:
@krbz999 was previously fixed issue that didn't have the fix merged into this branch; this PR is now API only.

krbz999 (approved):
Code looks fine.

If you could, update the outdated PR description, please.

PR #1910, opened 2026-05-20
[#676] Added conditional show logic for spend & strained message parts

### LINKED ISSUE #676: Only show additional `Spend HR` Entry if HR was spent or highlight it
### User Experience

Basically the title.

Right now if an ability has the option to spend additional HR for a certain effect, then it just shows up in the ability chat card, with no indication if that additional HR has been spent or not.

Options would be: Only show that part of the ability if HR has been spent, or highlight in some way that the HR has been spent for this effect.

Somewhat related to #398

JPMeehan:
Strained & spend effects will be hidden in the ability use part if not relevant.

Closes #676

### DIFF
--- a/src/module/data/helpers.mjs
+++ b/src/module/data/helpers.mjs
@@ -1,3 +1,7 @@
+/**
+ * @import { StringFieldOptions } from "@common/data/_types.mjs";
+ */
+
 const { NumberField, SchemaField, StringField } = foundry.data.fields;
 
 /**
@@ -16,7 +20,7 @@ export const requiredInteger = ({ initial = 0, min = 0, max, label, persisted = 
 
 /**
  * Constructs a string field for use inside of a SetField.
- * @param {import("@common/data/_types.mjs").StringFieldOptions} [options] Options to forward to the field.
+ * @param {StringFieldOptions} [options] Options to forward to the field.
  * @returns A string field that is always truthy.
  */
 export const setOptions = (options) => new StringField({ required: true, blank: false, ...options });
--- a/src/module/data/helpers.mjs
+++ b/src/module/data/helpers.mjs
@@ -16,7 +16,7 @@ export const requiredInteger = ({ initial = 0, min = 0, max, label, persisted = 
 
 /**
  * Constructs a string field for use inside of a SetField.
- * @param {object} [options] Options to forward to the field.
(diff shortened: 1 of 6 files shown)


src/module/data/pseudo-documents/special-effect/strained-effect.mjs
```diff
@@ -15,4 +15,13 @@ export default class StrainedSpecialEffect extends BaseSpecialEffect {
   get label() {
     return _loc("TYPES.SpecialEffect.strained");
   }
+
+  /* -------------------------------------------------- */
+
+  /** @inheritdoc */
+  showUse(_formData) {
```
krbz999: Is underscoring unused parameters a new paradigm we are introducing?

src/module/data/helpers.mjs
```diff
@@ -16,7 +16,7 @@ export const requiredInteger = ({ initial = 0, min = 0, max, label, persisted =
 
 /**
  * Constructs a string field for use inside of a SetField.
- * @param {object} [options] Options to forward to the field.
+ * @param {import("@common/data/_types.mjs").StringFieldOptions} [options] Options to forward to the field.
```
krbz999: Prefer importing this at the top of the file.

src/module/data/item/ability.mjs
```diff
@@ -550,6 +552,10 @@ export default class AbilityModel extends BaseItemModel {
             _id: abilityPartId,
             type: "abilityUse",
             abilityUuid: this.parent.uuid,
+            effects: this.effects.reduce((arr, e) => {
+              if (e.showUse(fd)) arr.push(e.id);
+              return arr;
+            }, []),
```
krbz999, line 558: You can fight me on this suggestion:
```suggestion
            effects: this.effects.filter(e => e.showUse(fd)).map(e => e.id),
```

krbz999 (approved):

src/module/data/item/ability.mjs
JPMeehan: Yeah for this kind of filter + map I do prefer the reduce, even if the code looks prettier.

src/module/data/pseudo-documents/special-effect/strained-effect.mjs
JPMeehan: Eh nah

PR #1913, opened 2026-05-21
[#1420] Implemented ability expansion logic

### LINKED ISSUE #1420: Allow ability selection to support non-core options.

JPMeehan:
Expanded ability logic to allow non-core drops. Best example would be updating the Elementalist advancements at levels like 2 where there's "Alternatively, you can choose one of the 5-essence abilities you didn't select at 1st level", but this also generally builds support for 1st & 3rd party supplements that add new abilities for existing classes.

image

Closes #1420


src/module/applications/apps/advancement/item-grant-configuration-dialog.mjs
```diff
@@ -202,6 +202,7 @@ export default class ItemGrantConfigurationDialog extends DSApplication {
       const perkLabels = Array.from(this.advancement.additional.perkType).map(p => perkOptions.find(o => o.value === p)?.label).filter(_ => _);
       const listFormatter = game.i18n.getListFormatter({ type: "disjunction" });
       const formatData = {
+        cost: _loc(`DRAW_STEEL.ADVANCEMENT.ITEM_GRANT.CostDrop.${this.advancement.additional.cost ? "heroic" : "signature"}`, { value: this.advancement.additional.cost }),
```
krbz999:
```suggestion
        cost: _loc(
          `DRAW_STEEL.ADVANCEMENT.ITEM_GRANT.CostDrop.${this.advancement.additional.cost ? "heroic" : "signature"}`,
          { value: this.advancement.additional.cost },
        ),
```

src/module/data/pseudo-documents/advancements/item-grant-advancement.mjs
```diff
@@ -25,6 +25,7 @@ export default class ItemGrantAdvancement extends BaseAdvancement {
       additional: new SchemaField({
         type: new StringField({ required: true }),
         perkType: new SetField(setOptions()),
+        cost: new NumberField({ required: true, integer: true, min: 1 }),
```
krbz999: Are there not a limited set of options?

krbz999 (approved):
Good for now. Will eventually be replaced with a button to open a Compendium Browser, so we need this data available the sooner the better.

src/module/data/pseudo-documents/advancements/item-grant-advancement.mjs
JPMeehan: There are, we could absolutely switch it to a select

PR #1923, opened 2026-05-29
[#1922] Implement LocalDocumentField

### LINKED ISSUE #1922: Implement a 'local document field'
A subclass of StringField or DocumentIdField that initializes a getter for an actor's owned item.

krbz999:
Closes #1922.

### DIFF
--- a/src/module/data/effect/local-document-field.mjs
+++ b/src/module/data/effect/local-document-field.mjs
@@ -1,0 +1,88 @@
+/**
+ * @import { StringFieldOptions, DataFieldContext } from "@client/data/_types.mjs"
+ * @import Document from "@common/abstract/document.mjs"
+ */
+
+/**
+ * @typedef _LocalDocumentFieldOptions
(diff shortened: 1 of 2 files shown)


src/module/data/effect/local-document-field.mjs
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * A reference to the model class which is stored in this field.
+     * @type {typeof Document}
+     */
+    this.model = model;
+
+    if (!(options.subtype in CONFIG[model.documentName].dataModels)) {
+      throw new Error("A LocalDocumentField must specify a specific document subtype.");
+    }
```
JPMeehan: For our use case, we want to accept "base" *or* "abilityModifier"

JPMeehan (changes requested):

src/module/data/effect/local-document-field.mjs
krbz999: Made `subtype` optional.

JPMeehan (approved):

PR #1932, opened 2026-05-30
Update incorrect before effects to after

cdverrett94:
Per our discord conversation, I noticed some effects were listed as before instead of after. Updated those and then corrected the UUID to match convention used when migrating.

Also found some strained effects not listed as strained effects while doing this and added those too.

This does not handle any potentially incorrect before/after effects inside of actors. Only base items in compendiums.


JPMeehan (approved):
LGTM

PR #1933, opened 2026-05-31
Add Abilities and Features to NPC Embeds & other small tweaks

### LINKED ISSUE #1916: Finish NPC Embeds
The top part of the NPC is finished, barring #1902. We need to finish the bottom part of the NPC embed, i.e., the list of abilities/features.

jalensailin:
#### This PR does the following:

- Adds Features to the NPC Embeds
- Adds Abilities to the NPC Embeds
- Displays an icon for each feature/ability
  - For Abilities, there is a best-effort function to assign the icon that it corresponds with in the PDF's stat blocks. However, there are some abilities that seem to defy the central logic, and thus there is the option to add a flag to any Ability item. The flag exists under the namespace `"draw-steel.embedDisplay"` and is under the property `iconOverride` (a string literal). Set that to one of the following options to override the icon that is used:
    - "area"
    - "burst"
    - "melee"
    - "meleeRanged"
    - "ranged"
    - "self"
    - "special"
    - "trait"
    - "triggered-action"
    - "villain-action"
  - For Features, there isn't sufficient information in the data model to assign the icon using logic. Thus, you must use the flag described above to override the icon. Otherwise Features default to the `"trait"` (star) icon.
- In statblocks, some Features come at the beginning of the list, before Abilities, and others come at the end. Thus there is another flag option. This one also exists under the same namespace, `"draw-steel.embedDisplay"` under the `displayAtEnd` property. This is a `boolean` that exists on the Feature. Set it to `true` to put that Feature after Abilities. The order is otherwise retained.
- Other small tweaks:
  - Add box-shadow to npc embeds, with shadow that matches the color. I think it helps identify the NPC's organization when there are many abilities, and the user would have to scroll down.
  - Adds the horizontal-line/triangular separator to the embed.

image

#### To test:
1. Grab the attached `json` file [Crux_of_Fire](https://github.com/user-attachments/files/28431017/npc_Crux_of_Fire_65JYST23SoHMm6FN.json), and copy (overwrite) it in the directory: `src/packs/monsters/Elementals_hatSTUdV3EGvlnvE/`
2. Run `npm run build:packs` to build that monster into the compendium.
3. `@Embed` the above monster into a Journal Entry.
4. Notice that the Ability named "Split" has the `self` icon (when otherwise it would have the `special` icon. 
5. Notice that "Fickle and Free" Feature is listed at the end, rather than the beginning.

**Note:** This PR does *not* include any data entry. That is a lot to go through, with the many monsters and their many more Abilities/Features. That should be a separate task for a separate PR.

Closes #1916

### DIFF
--- a/src/module/data/actor/npc.mjs
+++ b/src/module/data/actor/npc.mjs
@@ -227,7 +227,20 @@ export default class NPCModel extends CreatureModel {
 
     embed.classList.add("draw-steel", "actor", "npc", this.monster.role || "no-role");
 
-    // Prepare list of features and abilities (`ul`).
+    context.itemList = await this._getItemsList();
+
+    embed.innerHTML = await foundry.applications.handlebars.renderTemplate(systemPath("templates/embeds/actor/npc.hbs"), context);
+
+    return embed;
+  }
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Prepare the list of items for the NPC embed.
+   * @returns {Promise<string>}
+   */
+  async _getItemsList() {
     const abilities = this._getOrderedAbilities();
     const { startFeatures, endFeatures } = this._getOrderedFeatures();
     const orderedItems = [...startFeatures, ...abilities, ...endFeatures];
@@ -245,12 +258,10 @@ export default class NPCModel extends CreatureModel {
       itemList.append(li);
     }
 
-    context.itemList = itemList.outerHTML;
-
-    embed.innerHTML = await foundry.applications.handlebars.renderTemplate(systemPath("templates/embeds/actor/npc.hbs"), context);
-
-    return embed;
+    return itemList.outerHTML;
   }
+
+  /* -------------------------------------------------- */
 
   /**
    * Orders Ability Items in the same order
@@ -263,6 +274,8 @@ export default class NPCModel extends CreatureModel {
       (a, b) => keys.indexOf(a.system.type) - keys.indexOf(b.system.type),
     );
   }
+
+  /* -------------------------------------------------- */
 
   /**
    * Orders Feature Items via the `embedDisplay.displayAtEnd` flag.
--- a/src/module/data/actor/npc.mjs
+++ b/src/module/data/actor/npc.mjs
@@ -216,6 +216,7 @@ export default class NPCModel extends CreatureModel {
       actor: this.parent,
       characteristics: this._getCharacteristics(false),
       damageIW: this._getImmunitiesWeaknesses(),
+      itemEmbeds: await this._getItemEmbeds(),
       monsterKeywords: this._getMonsterKeywords().join(", "),
       movement: this._getMovement(true).list,
       system: this,
@@ -227,8 +228,6 @@ export default class NPCModel extends CreatureModel {
 
     embed.classList.add("draw-steel", "actor", "npc", this.monster.role || "no-role");
 
-    context.itemList = await this._getItemsList();
-
     embed.innerHTML = await foundry.applications.handlebars.renderTemplate(systemPath("templates/embeds/actor/npc.hbs"), context);
 
     return embed;
@@ -237,28 +236,19 @@ export default class NPCModel extends CreatureModel {
   /* -------------------------------------------------- */
 
   /**
-   * Prepare the list of items for the NPC embed.
-   * @returns {Promise<string>}
+   * Prepare embeds of Items for the NPC embed.
+   * @returns {Promise<DrawSteelItemEmbed[]>}
    */
-  async _getItemsList() {
+  async _getItemEmbeds() {
     const abilities = this._getOrderedAbilities();
     const { startFeatures, endFeatures } = this._getOrderedFeatures();
     const orderedItems = [...startFeatures, ...abilities, ...endFeatures];
 
-    const itemList = document.createElement("ul");
-    itemList.classList.add("abilities-list");
-    for (const item of orderedItems) {
-      const itemEmbed = await item.system.toEmbed({ includeName: true });
-      const li = document.createElement("li");
-      li.innerHTML = itemEmbed.outerHTML;
-
-      const itemIcon = item.system.getStatBlockIcon?.();
-      if (itemIcon) li.classList.add(itemIcon);
-
-      itemList.append(li);
-    }
-
-    return itemList.outerHTML;
+    return Promise.all(orderedItems.map(async (item) => {
+      const embed = await item.system.toEmbed({ includeName: true });
+      embed.icon = item.system.getStatBlockIcon?.();
+      return embed;
+    }));
   }
 
   /* -------------------------------------------------- */
--- a/src/module/data/actor/npc.mjs
+++ b/src/module/data/actor/npc.mjs
@@ -260,7 +260,7 @@ export default class NPCModel extends CreatureModel {
    */
   _getOrderedAbilities() {
     const abilityTypes = Object.keys(ds.CONFIG.abilities.types);
-    return this.parent.items.documentsByType.ability.sort(
+    return this.parent.items.documentsByType.ability.toSorted(
       (a, b) => abilityTypes.indexOf(a.system.type) - abilityTypes.indexOf(b.system.type) || a.sort - b.sort,
     );
   }
@@ -272,9 +272,9 @@ export default class NPCModel extends CreatureModel {
    * @returns {{ startFeatures: DrawSteelItem[], endFeatures: DrawSteelItem[] }}
    */
   _getOrderedFeatures() {
-    const features = this.parent.items.documentsByType.feature.sort((a, b) => a.sort - b.sort);
+    const features = this.parent.items.documentsByType.feature.toSorted((a, b) => a.sort - b.sort);
 
-    const [ endFeatures, startFeatures ] = features.partition(f => {
+    const [ startFeatures, endFeatures ] = features.partition(f => {
       return !!f.getFlag(ds.CONST.systemID, "embedDisplay.displayAtEnd");
     });
     return { startFeatures, endFeatures };
--- a/src/module/data/actor/npc.mjs
+++ b/src/module/data/actor/npc.mjs
@@ -228,7 +228,8 @@ export default class NPCModel extends CreatureModel {
     embed.classList.add("draw-steel", "actor", "npc", this.monster.role || "no-role");
 
     // Prepare list of features and abilities (`ul`).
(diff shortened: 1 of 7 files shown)


src/module/data/actor/npc.mjs
```diff
@@ -224,6 +224,34 @@ export default class NPCModel extends CreatureModel {
...
+
+    const startFeatures = [];
+    const endFeatures = [];
+    feature.forEach(f => {
+      /** @type {EmbedDisplayFlags} */
+      const flag = f.getFlag(ds.CONST.systemID, "embedDisplay");
+      if (flag?.displayAtEnd) endFeatures.push(f);
+      else startFeatures.push(f);
+    });
+    const orderedItems = [...startFeatures, ...abilities, ...endFeatures];
```
JPMeehan, line 245: hmm, should we be using `sort` and other logic to the abilities?

src/module/data/item/base-item.mjs
```diff
@@ -123,13 +126,66 @@ export default class BaseItemModel extends DrawSteelSystemModel {
...
 
   /* -------------------------------------------------- */
 
+  /**
+   * Logic to get the stat-block icon for use in Actor embed.
+   * Currently only applies to `feature` and `ability` items.
+   * @returns {string}
+   */
+  getStatBlockIcon() {
+    if (!["feature", "ability"].includes(this.parent.type)) return null;
```
JPMeehan: While there's definitely some parallels it seems we should just be taking advantage of the polymorphism here? Almost all of the code here is ability-exclusive.

templates/embeds/actor/npc.hbs
```diff
@@ -97,4 +97,6 @@
     </div>
     {{/each}}
   </div>
+
+  {{{itemList}}}
```
JPMeehan: I feel like it would be better if the elements were here and then you were just feeding in the appropriate data, I don't think there's a compelling reason to lean on document.createElement for the embed (amongst other reasons makes it easier to edit this in the future with hot reload).

JPMeehan (commented):
Looks good and yeah the data entry can absolutely come later as part of refinement.

src/module/data/item/base-item.mjs
jalensailin: I was going back and forth since its so limited. Happy to revert it to be on the specific classes

templates/embeds/actor/npc.hbs
jalensailin: No problem, addressed in recent commit.

src/module/data/item/base-item.mjs
jalensailin: Addressed in recent commit

src/module/data/actor/npc.mjs
jalensailin: Added a sort function in a recent commit. It sorts based on the keys of `ds.CONFIG.abilities.types`

src/module/data/actor/npc.mjs
```diff
@@ -231,6 +233,58 @@ export default class NPCModel extends CreatureModel {
...
+
+  /**
+   * Orders Ability Items in the same order
+   * as the keys in ds.CONFIG.abilities.types.
+   * @returns {DrawSteelItem[]}
+   */
+  _getOrderedAbilities() {
+    const keys = Object.keys(ds.CONFIG.abilities.types);
+    return this.parent.items.documentsByType.ability.sort(
+      (a, b) => keys.indexOf(a.system.type) - keys.indexOf(b.system.type),
```
JPMeehan: Should also sort by the sort of the individual abilities so it fully matches the sheet (e.g. relative ordering of main actions).

src/module/data/actor/npc.mjs
```diff
@@ -231,6 +233,58 @@ export default class NPCModel extends CreatureModel {
...
+
+    const startFeatures = [];
+    const endFeatures = [];
+    feature.forEach(f => {
+      /** @type {EmbedDisplayFlags} */
+      const flag = f.getFlag(ds.CONST.systemID, "embedDisplay");
+      if (flag?.displayAtEnd) endFeatures.push(f);
+      else startFeatures.push(f);
+    });
+    return { startFeatures, endFeatures };
```
JPMeehan: Same comment on sort property

JPMeehan (commented):
small adjustment

src/module/data/actor/npc.mjs
jalensailin: This should be addressed in the latest commit. I tested with the following sample data:

```js
const testData = [
      { system: { type: "main" }, sort: 2 },
      { system: { type: "maneuver" }, sort: 4 },
      { system: { type: "triggered" }, sort: 3 },
      { system: { type: "main" }, sort: 1 },
      { system: { type: "main" }, sort: 0 },
      { system: { type: "maneuver" }, sort: 5 },
      { system: { type: "villain" }, sort: 8 },
      { system: { type: "triggered" }, sort: 1 },
      { system: { type: "maneuver" }, sort: 3 },
      { system: { type: "triggered" }, sort: 8 },
      { system: { type: "villain" }, sort: 7 },
      { system: { type: "none" }, sort: 8 },
]
```

src/module/data/actor/npc.mjs
jalensailin: Addressed as above

src/module/data/actor/npc.mjs
```diff
@@ -231,6 +233,59 @@ export default class NPCModel extends CreatureModel {
...
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Orders Feature Items via the `embedDisplay.displayAtEnd` flag.
+   * @returns {{ startFeatures: DrawSteelItem[], endFeatures: DrawSteelItem[] }}
+   */
+  _getOrderedFeatures() {
+    const features = this.parent.items.documentsByType.feature.sort((a, b) => a.sort - b.sort);
+    console.info("features", features);
```
JPMeehan: Hanging console then we're good

JPMeehan (approved):

krbz999:
I think we should make some stylistic choices while we're here. Like mentioned in a previous PR, the embeds are huge, and this has been exacerbated with the addition of the actor's items. I think, at the very least, we should implement column layouts and see what can be done about the white space. Compared to the official layout, ours are very padded.

image

image

Using the bugbear channeler, the embed is also *long*, excessively so. I think a double-column layout (dynamically adjusted depending on container size) and loss of extra padding would do wonders.

src/module/data/actor/npc.mjs
```diff
@@ -231,6 +233,59 @@ export default class NPCModel extends CreatureModel {
...
+  /**
+   * Orders Ability Items in the same order
+   * as the keys in ds.CONFIG.abilities.types.
+   * @returns {DrawSteelItem[]}
+   */
+  _getOrderedAbilities() {
+    const abilityTypes = Object.keys(ds.CONFIG.abilities.types);
+    return this.parent.items.documentsByType.ability.sort(
+      (a, b) => abilityTypes.indexOf(a.system.type) - abilityTypes.indexOf(b.system.type) || a.sort - b.sort,
+    );
```
krbz999: `documentsByType` is cached and `Array#sort` sorts inplace. We can use `Array#toSorted`.

src/module/data/actor/npc.mjs
krbz999:
```suggestion
```

src/module/data/actor/npc.mjs
```diff
@@ -231,6 +233,59 @@ export default class NPCModel extends CreatureModel {
...
+    console.info("features", features);
+
+    const startFeatures = [];
+    const endFeatures = [];
+    features.forEach(f => {
+      /** @type {EmbedDisplayFlags} */
+      const flag = f.getFlag(ds.CONST.systemID, "embedDisplay");
+      if (flag?.displayAtEnd) endFeatures.push(f);
+      else startFeatures.push(f);
+    });
```
krbz999:
```suggestion
    const [ endFeatures, startFeatures ] = features.partition(f => {
      return !!f.getFlag(ds.CONST.systemID, "embedDisplay.displayAtEnd");
    });
```

src/module/data/item/base-item.mjs
```diff
@@ -123,7 +125,14 @@ export default class BaseItemModel extends DrawSteelSystemModel {
 
     const embed = document.createElement("div");
     embed.classList.add("draw-steel", this.parent.type);
-    embed.innerHTML = enriched;
+    if (config.includeName) {
+      embed.innerHTML = `
+        <h5>${this.parent.name}</h5>
```
krbz999:
```suggestion
        <h5 data-no-toc>${this.parent.name}</h5>
```

krbz999 (dismissed):

src/module/data/actor/npc.mjs
```diff
@@ -231,6 +233,59 @@ export default class NPCModel extends CreatureModel {
...
+  }
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Orders Feature Items via the `embedDisplay.displayAtEnd` flag.
+   * @returns {{ startFeatures: DrawSteelItem[], endFeatures: DrawSteelItem[] }}
+   */
+  _getOrderedFeatures() {
+    const features = this.parent.items.documentsByType.feature.sort((a, b) => a.sort - b.sort);
```
krbz999: Same comment re: sorting here

src/module/data/actor/npc.mjs
jalensailin: Addressed in recent commit

src/module/data/actor/npc.mjs
jalensailin: Addressed in recent commit

src/module/data/actor/npc.mjs
jalensailin: Addressed

JPMeehan (approved):
LGTM

jalensailin:
> I think we should make some stylistic choices while we're here. Like mentioned in a previous PR, the embeds are huge, and this has been exacerbated with the addition of the actor's items. I think, at the very least, we should implement column layouts and see what can be done about the white space. Compared to the official layout, ours are very padded.
> 
> Using the bugbear channeler, the embed is also _long_, excessively so. I think a double-column layout (dynamically adjusted depending on container size) and loss of extra padding would do wonders.

I agree with your assessment here. Happy to explore these changes in another PR when I get back from my vacation (leaving in about an hour). I spoke with Joseph about it and he's good with that

PR #1945, opened 2026-06-05
[#852] Added spend values to roll data

### LINKED ISSUE #852: Provide resource expenditure as roll data
### User Experience

The Psychic Blast granted by Psychic Eruption complication works differently when compared to other heroic abilities.

Psychic Blast (Special Heroic Resource Cost)
Psionic energy bursts from your body in an iridescent shimmer.
Area, Psionic Main action
3 burst 
Each creature in the area

Effect: Using this ability costs all your Heroic Resource.

Power Roll + Your Highest Characteristic Score:
- 11 or lower: 1 psychic damage for each Heroic Resource you spend, to amaximum equal to your level
- 12-16: 1 psychic damage for each Heroic Resource you spend, to a maximum equal to your level + your highest characteristic
- 17+: 1 psychic damage for each Heroic Resource you spend

JPMeehan:
It's not the prettiest solution *but* it's fully functional.

image

Closes #852

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ Increased foundry minimum to 14.363.
 - Abilities can now have any number of HTML sections, and they can have additional info. (#475)
   - This includes promoting the "spend" field to have an HTML entry. (#329)
   - Strained & Spend message parts only display if their conditions are fulfilled. (#676)
+  - Spend effects now provide the spend value as roll data under `@spend`. (#852)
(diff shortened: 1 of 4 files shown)


krbz999 (approved):
It's not the best place to modify rolldata, but i dont feel too strongly since it's a one-and-done.

If we have to do this a lot more, we should consider using flags (cloned item) and extending `getRollData` further.

PR #1947, opened 2026-06-06
[#1901] Fixed minion interaction with heal enricher

### LINKED ISSUE #1901: Minions receiving stamina are damaged instead
### What happened?

When a selected minion receives stamina (or temp stamina) through e.g. [[/heal 5]] are instead damaged for 5. 

While Draw Steel minions can't heal or receive temp stamina, I wouldn't expect them to be damaged. I'd expect one of either of these options:
- stamina/temp stamina applies, with an alert that minions usually can't be healed/receive temp stamina.
- stamina/temp stamina does _not_ apply, with an alert that because minions can't be healed/receive temp stamina, it has been negated.

### Reproduction Steps

1. Select at least one minion token.
2. Trigger `applyDamage( { isHeal: true } )` i.e. by clicking on an enricher made by [[/heal x]]
3. Watch minion take damage instead.

FVTT 14.361, Draw Steel 1.1.0

### Bug Checklist

JPMeehan:
Closes #1901

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,7 @@ Increased foundry minimum to 14.363.
 
 ### Fixed
 
+- Minions are no longer damaged by heal enrichers, instead a warning is issued that they cannot regain stamina or gain temporary stamina. (#1901)
 - Added missing showIcon field to the details page of the Active Effect Config sheet. (#1917)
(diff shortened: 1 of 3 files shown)


krbz999 (approved):
Minions don't deserve nice things.

PR #1948, opened 2026-06-20
[#1694] Status Active Status Clarification

### LINKED ISSUE #1694: Condition Buttons and Assign Status Effect for Inactive Effects

JPMeehan:
- Inactive effects are no longer highlighted
- Toggling a condition on while it exists but disabled will delete the existing one

Closes #1694

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ Increased foundry minimum to 14.363.
 - Updated Player-Facing Compendium Content:
   - Updated "I'm No Threat", "Careful Observation" to have AEs that provide edges.
   - Added appropriate enrichers to Burning Ash and Smoke Bomb.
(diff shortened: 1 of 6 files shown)


krbz999 (commented):
Code seems fine for what it does.

How do we stand on disabling the ability to toggle a temporary effect?

The css should also be adjusted such that disabled statuses on the token HUD do not have cursor-pointer.

JPMeehan:
I think disabling toggling temporary effects is an eh improvement at the end of the day

krbz999 (approved):

PR #1952, opened 2026-06-21
Adventure Items Fix

JPMeehan:
Fixed a specific issue where pseudodocument initialization would fail because Adventures use the `BaseItem` class instead of `DrawSteelItem` class. Open to suggestions on alternative ways to check that it's BaseItem.


krbz999 (approved):

krbz999:
Need a rebase here

PR #1953, opened 2026-06-21
Adjusted default artwork logic

JPMeehan:
Made it easier for people manually replacing system-provided actor images.

### DIFF
--- a/src/module/documents/actor.mjs
+++ b/src/module/documents/actor.mjs
@@ -35,7 +35,7 @@ export default class DrawSteelActor extends BaseDocumentMixin(foundry.documents.
         },
       };
     }
-    else return super.getDefaultArtwork(data);
+    return super.getDefaultArtwork(data);
   }
 
   /* -------------------------------------------------- */
--- a/src/module/documents/actor.mjs
+++ b/src/module/documents/actor.mjs
@@ -17,6 +17,8 @@ export default class DrawSteelActor extends BaseDocumentMixin(foundry.documents.
     return super.migrateData(data);
   }
 
+  /* -------------------------------------------------- */
+
   /**
    * Is this actor a minion?
    * @returns {boolean}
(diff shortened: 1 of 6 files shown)


src/module/documents/actor.mjs
```diff
@@ -59,6 +61,20 @@ export default class DrawSteelActor extends BaseDocumentMixin(foundry.documents.
 
   /* -------------------------------------------------- */
 
+  /** @inheritdoc */
+  async _preUpdate(changed, options, user) {
+    const allowed = await super._preUpdate(changed, options, user);
+    if (allowed === false) return false;
+    // Make it easier for users to override system-provided default icons
+    if (changed.img && !foundry.utils.getProperty(changed, "prototypeToken.texture.src")) {
+      if (!this.prototypeToken.texture.src || (this.prototypeToken.texture.src.includes("systems/draw-steel/"))) {
```
krbz999: Looking at relevant locations where the concept of adjusting the prototype token in conjunction with the actor img, it looks like we want `_initializeSource` and `_preUpdate` as well. We should also adjust `Actor.getDefaultArtwork`.

Also worth considering the "show token artwork" header button (in `ActorSheet#_getHeaderControls`).

JPMeehan: I don't think _initializeSource is necessary, that's only for compendium art.

I considered adjusting the getDefaultArtwork method, but I don't think that as effectively addresses the goal of this PR: Make it easier for people to manually switch off of the system-provided art.

krbz999: Looks like I wrote preupdate when I actually meant precreate.

JPMeehan: I still don't think that's related to the purpose of the change I'm making.

krbz999 (changes requested):
When changing the actor image, I do not want the token texture to change just because it is currently an icon from the system.

I want the token texture to change if it is the default icon. Therefore we should compare to the default artwork.

JPMeehan:
So, the up/downside here is practically speaking, our "default NPC art" is almost entirely based on the Monster Role, which is *not* going to be filled in from a simple Actor.createDialog; furthermore, if someone is just trying to switch the actor art to another one of our role icons (say you're duplicating an existing monster as a base and then making some adjustments), they would need to make sure they do the art change *then* the actual data change.

krbz999:
Well `Actor.getDefaultArtwork` doesn't accept just `type`, it accepts an entire object of actor data. It is not strictly related to the creation dialog.

It should, in my opinion, determine the default artwork using whatever data is available as the API method that it is.

JPMeehan:
Fully refactored to just use getDefaultArtwork.

src/module/documents/actor.mjs
```diff
@@ -17,6 +18,28 @@ export default class DrawSteelActor extends BaseDocumentMixin(foundry.documents.
...
+      if (["leader", "solo", "minion"].includes(organization)) filePath = `systems/${systemID}/assets/roles/${organization}.webp`;
+      else if (role) filePath = `systems/${systemID}/assets/roles/${role}.webp`;
+      return {
+        img: filePath,
+        texture: {
+          src: filePath,
+        },
+      };
+    }
+    else return super.getDefaultArtwork(data);
```
krbz999:
```suggestion
    return super.getDefaultArtwork(data);
```

krbz999 (approved):

PR #1954, opened 2026-06-21
Fixed 1.1 build issue where an un-grouped token would crash the canvas

JPMeehan:
Issue internal to 1.1 build due to me not quite realizing what was pass-by-value vs. pass-by-reference


krbz999 (approved):
lmao.

PR #1957, opened 2026-06-25
Various CSS fixes for 1.1.x

### LINKED ISSUE #1682: Improve ultrawide tables
_Draw Steel: Heroes_ includes two ultrawide tables for the Green Elementalist change form and the Kits table. Both need some improvements.

### LINKED ISSUE #1926: Enforce default height and width of `ds-icon`
Generally useful to just have this apply everywhere and have applications override it if needed.

```css
ds-icon,
ds-icon svg {
  height: 100%;
  width: 100%;
}
```

### LINKED ISSUE #1934: Stylistic dividers fall apart when resizing containers
Image

jalensailin:
This PR does the following:
- Fixes the section divider in NPC embeds "splitting" in the middle when the Journal is wide (Closes #1934):
  
  image

- Adds default height/width for `ds-icons` (Closes #1926)
- Adds horizontal scroll to tables that overflow their container, i.e. ultrawide tables (Closes #1682)


src/styles/system/applications/sheets/journal-entry-sheet.css
```diff
@@ -289,6 +289,11 @@ aside[role="tooltip"] section.content {
 
   table {
     width: 100%;
+
+    /** Scroll overflowing tables **/
+    display: block;
```
JPMeehan: so, according to MDN, this is an a11y issue because screen readers rely on `display: table` to sort properly in the hierarchy. I'm not sure there's an alternative solution unfortunately without just wrapping every table? Is this maybe something we could reserve for some limited class?

JPMeehan (commented):
I also think we need some additional work for dark mode to flip the color of the inserts?

Image

JPMeehan:
We should also consider sticky headers
```css

    // When the table overflows when given a maximum height, keep the head at the top.
    thead {
      position: sticky;
      top: 0;
    }
```

src/styles/system/applications/sheets/journal-entry-sheet.css
jalensailin: After looking more into this, we have two options here.

1. Wrap ultrawide tables (or better, all tables, as you suggested, since we can't predict the exact size of the journal window) and apply `overflow: auto` to the wrapper.
2. Keep `display: block` on the `<table>` element, and add the ARIA `role` attribute to every table element, for screen reader accessibility. See this article on MDN: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/table_role#best_practices.

I think option 1 is the best option, and easiest to apply en masse. Let me know what you think

jalensailin:
> We should also consider sticky headers
> 
> ```css
>     // When the table overflows when given a maximum height, keep the head at the top.
>     thead {
>       position: sticky;
>       top: 0;
>     }
> ```

This requires us to set a maximum height on tables. I am fine with that but we should agree on a good default.

jalensailin:
> I also think we need some additional work for dark mode to flip the color of the inserts?

This is addressed in recent commit:

image

jalensailin:
I have added a maximum height for tables to 80% of its containers (`.journal-entry-pages`) width. This makes the `<thead>` sticky, and ensures there aren't overly vertically high tables. 

image

I am happy to tweak the 80% value.

src/styles/system/applications/sheets/journal-entry-sheet.css
JPMeehan: I think a mix of 1 & 2; we should have some kind of css class we can apply to tables that does the display: block that we mix with the aria-role attribute.

jalensailin: Added a css class: `.large`. When this is applied the `table` becomes `display: block` (and gets the fancy scrolling). Otherwise they retain `display: table`. I wrote a comment to remind future devs that tables with this class need to have the `role` attributes and linked to the MDN explanation. Finally, I went ahead and updated this for the "Disciple of the Green" feature description. Want me to make an MR for the fix in Heroes as well?

JPMeehan (approved):
Yeah, this is what I wanted. Looks good!

krbz999 (approved):

PR #1958, opened 2026-06-25
Object resources displayed as column in edit mode

### LINKED ISSUE #1955: Object Sheet CSS displays incorrectly
### What happened?

Image

Image

This was a mistake that was likely introduced as an unwanted side effect of my #1815 PR. These are supposed to be a flex row instead of a `column-reverse`. We also should not rely on selectors like `> * >`, instead add css class like so (as per Zhell)

```scss
// selector: .form-group.row

.draw-steel .form-group {   
  &.row {
    flex-direction: row;   
  }
}
```

### Reproduction Steps

1. Open a fresh Object Sheet
2. Toggle edit mode
3. Observe the wonky css (as above)

FVTT 14.346, Draw Steel 1.1.x

### Bug Checklist

Katiedead:
image

- I used more specific selectors for the `.form-group.paired` selectors, which are form item and labels typically.
- I corrected the `.hbs` file for the object actor form html to remove `.paired` from the class list of the div inside of `.resource-max`

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,7 @@ Increased foundry minimum to 14.363.
 - Minions are no longer damaged by heal enrichers, instead a warning is issued that they cannot regain stamina or gain temporary stamina. (#1901)
 - Added missing showIcon field to the details page of the Active Effect Config sheet. (#1917)
 - Corrected instances where the "turn end" expiration would not match to the correct combatant.
+- Corrected object resource form displaying as a column instead of a row in edit mode. (#1955)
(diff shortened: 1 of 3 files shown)


jalensailin (commented):
This looks good to me. Looks correct in non-edit mode as well:

image
 
I also checked the other Actor sheets, and those look fine in non/edit mode too.

Will wait for @JPMeehan's review but otherwise should be good to go.

JPMeehan (approved):
Looks great thanks

PR #1959, opened 2026-07-01
System Hooks

### LINKED ISSUE #1941: System Hook: ds.preRenderAbilityConfigurationDialog
### User Experience

As a Module Developer I would like a cancellable hook that fires before the ability configuration dialog renders so I can gate ability use without patching the private `_canRender` lifecycle method.

`AbilityConfigurationDialog` does not currently override `_canRender`.
Proposed addition:
```js
// New override in AbilityConfigurationDialog:
_canRender(options) {
  const allowed = Hooks.call('ds.preRenderAbilityConfigurationDialog', this);
  return super._canRender(options);
}
```
```js
Hooks.call('ds.preRenderAbilityConfigurationDialog', dialog);
// dialog: the AbilityConfigurationDialog instance (access .item for the ability, .actor for the user)
// return false to prevent the dialog from opening
```
`_canRender` is a private lifecycle method with no stable override point, making it fragile to wrap. This covers any module that conditionally blocks ability use: range checks, resource requirements, or targeting flows that need to run before the dialog opens.

JPMeehan:
Closes #1941


src/docs/Hooks.md
```diff
@@ -0,0 +1,20 @@
+The Draw Steel system adds the following hooks to the ones available in the [core software](https://foundryvtt.com/api/modules/hookEvents.html).
+
+## Prepare Data
+
+The ActiveEffect, Actor, ChatMessage, Combat, Combatant, Item, and User documents all have a hook of the format `ds.prepare${DocumentName}Data`. This hook is called after all other data preparation has completed. For example, you could use this to give all heroes an additional 10 max stamina without needing to apply an active effect.
```
krbz999: Well really this begs the question:

Why not just apply an ActiveEffect?

But I don't want such a debate to hold up this PR, so let's remove this section for now.

src/module/applications/apps/ability-configuration-dialog.mjs
```diff
@@ -108,6 +108,15 @@ export default class AbilityConfigurationDialog extends PowerRollDialog {
 
   /* -------------------------------------------------- */
 
+  /** @inheritdoc */
+  _canRender(options) {
+    const allowed = Hooks.call("ds.preRenderAbilityConfigurationDialog", this);
+    if (allowed === false) return false;
+    return super._canRender(options);
+  }
```
krbz999: I think I'd prefer we add it to a superclass.

is "preRender" a good name though? It is also called when re-rendering so can be a bit of a misnomer. How about simply `ds.canRender${this.constructor.name}`?

src/docs/Hooks.md
```diff
@@ -0,0 +1,20 @@
...
+});
+```
+
+## Drop Data
+
+The base Foundry software provides a [`dropActorSheetData`](https://foundryvtt.com/api/functions/hookEvents.dropActorSheetData.html) hook. The Draw Steel system also provides a similar `dropItemSheetData` hook.
+
+## Ability Configuration Dialog
+
+The Ability Configuration Dialog provides a `ds.preRenderAbilityConfigurationDialog` hook that if returned an explicit `false` will prevent it from rendering. Keep in mind that hooks only run synchronously; a Promise that resolves to false is not an explicit false.
```
krbz999: We don't need to lecture on how hooks work.

krbz999 (changes requested):

src/docs/Hooks.md
```diff
@@ -0,0 +1,20 @@
+The Draw Steel system adds the following hooks to the ones available in the [core software](https://foundryvtt.com/api/modules/hookEvents.html).
+
+## Prepare Data
+
+The ActiveEffect, Actor, ChatMessage, Combat, Combatant, Item, and User documents all have a hook of the format `ds.prepare${DocumentName}Data`. This hook is called after all other data preparation has completed.
```
krbz999: I should have been clearer. What I meant was that, since the result of this hook could be achieved with an AE anyway, like the omitted sentence implied, I think we should neglect to document this hook until such a time that we decide that it is actually worth keeping.

At the moment, there are no modules making use of it to my knowledge.

krbz999 (changes requested):

src/docs/Hooks.md
JPMeehan: Ah, yeah I added it because my experience with SWADE & dnd5e is you sometimes just need to do stuff in data prep, in a way that is sometimes simplified by combination with flags

krbz999 (changes requested):
Comments not yet resolved.

JPMeehan:
Alright, I'll remove the data prep hook docs but I think it's a good hook that helps the system stay flexible with minimal side effects.

krbz999 (approved):

PR #1979, opened 2026-07-12
Guard against video token images causing errors in combatant group creation

### LINKED ISSUE #1977: Omen Dragon - Incorrect test attribute
### Which compendiums is this in?

Monsters

### Documents

Corroding Breath - Omen Dragon

### What changes need to be made?

“Corroding Breath” from the Omen Dragon has a Presence test instead of an Agility one.

### LINKED ISSUE #1978: DrawSteelCombatantGroup.createFromTokens can throw if the included tokens use video files for token images
### What happened?

When creating a combat group with only minions or single type unit it doesn't like the .webm file, but providing there's a captain / other unit it lets the group get created.

### Reproduction Steps

Image

Image

_No response_

### Bug Checklist

JPMeehan:
Creating a combatant group from tokens can cause an issue with .webm files which are a legal extension for texture.src but not img

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,13 +71,15 @@ Increased foundry minimum to 14.364.
 - Player-Facing Compendium Data Fixes:
   - Corrected Back! to not have a forced movement in tier 1. (#1898)
   - Added in correct source data for several triggered actions granted by features. (#1935)
-  - Corrected Censor level 4 characteristic advancement to give Presence and not Reason. (##1968)
+  - Corrected Censor level 4 characteristic advancement to give Presence and not Reason. (#1968)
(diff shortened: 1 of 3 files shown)


krbz999 (commented):
Think I'd prefer finding the first non-video asset and using that.

JPMeehan:
Right now the logic is if they don't all match we go to default. It's unlikely there's going to be a non video asset in a given group

krbz999:
I think finding a `texture.src` property that has an image extension would be easier than nesting ternaries.

JPMeehan:
Okay, so adjust behavior so we just find an image rather than rely on them all being the same?

krbz999 (approved):

PR #1980, opened 2026-07-13
Add documentation for NPC Embeds

jalensailin:


src/docs/Enrichers.md
```diff
@@ -1,5 +1,30 @@
...
 
+## Actor Embeds
+
+NPC Actors can be embedded into enriched text via the `@Embed[uuid]` syntax. This embed displays similarly as the statblocks in the book and contains the same relevant information there:
+
+![NPC Embed Example](https://github.com/MetaMorphic-Digital/draw-steel/blob/1.1.x/assets/docs/NPC-embed-example.png)
+
+### Setting up an NPC's Features/Abilities so that its `@Embed`s display as expected
+
+- For Abilities, there is a best-effort attempt to assign the icon that it corresponds with in the PDF's statblocks. However, there are some Abilities that don't follow an obvious pattern for their icon, and thus a flag exists which can manually set the displayed icon. The flag exists under the namespace `npc.flags.draw-steel.embedDisplay`, under the property `iconOverride` (a `string` literal). Set that property to one of the following options to override the icon that is used:
```
JPMeehan: Don't lead with `npc` here, and I think the mixed use of namespace and property is confusing. If we're going to do namespace + property, then the namespace is just `draw-steel.embedDisplay`, and we're giving a data path, do `flags.draw-steel.embedDisplay.iconOverride`. I would prefer the first option as it matches what we did on the Table-of-Contents page

src/docs/Enrichers.md
```diff
@@ -1,5 +1,30 @@
...
+  - "melee"
+  - "meleeRanged"
+  - "ranged"
+  - "self"
+  - "special"
+  - "trait"
+  - "triggered-action"
+  - "villain-action"
+- For Features, there isn't sufficient information in the data model to assign the icon using logic alone. Thus, Features default to using the "trait" (star) icon. If you wish to select the icon yourself, you must use the flag described above to override the icon.
+- In statblocks, some Features come at the beginning of the list, before Abilities, and others come at the end. Thus there is another flag option for Features, under the same namespace: `npc.flags.draw-steel.embedDisplay`, under the `displayAtEnd` property. This is a `Boolean` that exists on the Feature. Set it to `true` to put that Feature after the Abilities, rather than before. The order is otherwise retained.
```
JPMeehan: Same as above

src/docs/Enrichers.md
```diff
@@ -1,5 +1,30 @@
...
+- You can set this flag on an item in Foundry via `item.setFlag("draw-steel", "embedDisplay", { iconOverride: "special", displayAtEnd: true  }`
+
+**Note:** Currently, only NPC Actors can be embedded in this way. Support for embedding additional Actor types is planned.
```
JPMeehan: Remove from this spot and add it as a parenthetical at the NPC Embed section, specifically naming that Retainer & Object actors will come in the future.

JPMeehan (commented):
Mostly good, but want some wording adjustments

src/docs/Enrichers.md
jalensailin: Addressed

src/docs/Enrichers.md
jalensailin: Addressed

src/docs/Enrichers.md
jalensailin: Addressed

jalensailin:
Addressed comments here and incorporated Amy's suggestion from Discord

src/docs/Enrichers.md
```diff
@@ -1,5 +1,28 @@
 "Enrichers" refer to specialized text syntax that can be used in any rich text editor (such as an actor description, journal entry, or chat message). Enriched content displays simply in the editor but can represent large and complex displays in the rendered text.
 
+## Actor Embeds
+
+NPC Actors can be embedded into enriched text via the `@Embed[uuid]` syntax. **Note:** Currently, only NPC Actors can be embedded in this way. Support for embedding additional Actor types is planned. This embed displays similarly as the statblocks in the book and contains the same relevant information there:
```
JPMeehan:
```suggestion
NPC Actors can be embedded into enriched text via the `@Embed[uuid]` syntax. (Only NPC Actors can be embedded in this way currently; support for Objects and Retainers is planned)

This embed displays similarly as the statblocks in the book and contains the same relevant information there:
```

JPMeehan (approved):

PR #1992, opened 2026-07-16
[#1982] Added Ancestry Malice ability category

### LINKED ISSUE #1982: Add Ancestry Malice cateogry to filter out abilities from NPC embed
As a content creator, I would like the ability to not show commonly shared abilities like Malice features that might otherwise be repeated on a journal page. Ideally there should be a config option to ignore this skip logic.

JPMeehan:
- Did some type updates now that EmbeddedCollection properly passes a generic back to Collection
- This allows us to better match the books, where ancestry-based malice abilities are displayed separately from the stat blocks.

Closes #1982

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,8 +23,8 @@
 ### Added
 
 - New Player-Facing Compendium Content:
-  - Added Armor & Implement imbues to the Effects compendium
-- Added an "Ancestry Malice" category for abilities that are from a monster's ancestry, not from its unique statblock. Ancestry malice abilities are skipped in the actor embed unless the embed is supplied the `showAll=true` option.
+  - Added Armor, Implement, and Weapon imbues to the Effects compendium
+- Added an "Ancestry Malice" category for abilities that are from a monster's ancestry, not from its unique statblock. Ancestry malice abilities are skipped in the actor embed unless the embed is supplied the `showAll=true` option.  (#1982)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@
 
 - New Player-Facing Compendium Content:
   - Added Armor & Implement imbues to the Effects compendium
-- Added an "Ancestry Malice" category for abilities which will be skipped by the embed. Using `showAllItems=true` in the embed call will include the abilities anyways.
+- Added an "Ancestry Malice" category for abilities that are from a monster's ancestry, not from its unique statblock. Ancestry malice abilities are skipped in the actor embed unless the embed is supplied the `showAll=true` option.
 
 ### Changed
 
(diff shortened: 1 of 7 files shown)


krbz999 (commented):
Fine idea but I think the flag should exist on the actor instead, as an array of ids. That way, copying an item from one actor to another would not have it hidden on the next actor, too.

JPMeehan:
The counter argument is that's desirable, malice abilities are common through the monster band and you'd want to skip them if you made a new goblin and grabbed the ones from our Monster Features compendium.

krbz999:
Counter-counter argument is that bulk updates for a collection of actors is simpler than bulk updates of their embedded documents.

Should we have a way to set these flags in the UI?

If it were on an actor, we could use an effect.

JPMeehan:
There's no UI logic right now — the other part of my decision here is we're otherwise using this item flag. Would you like it more if there was a helper app?

src/docs/Enrichers.md
```diff
@@ -23,7 +23,8 @@ This embed displays similarly as the statblocks in the book and contains the sam
   - "villain-action"
 - For Features, there isn't sufficient information in the data model to assign the icon using logic alone. Thus, Features default to using the "trait" (star) icon. If you wish to select the icon yourself, you must use the flag described above to override the icon.
 - In statblocks, some Features come at the beginning of the list, before Abilities, and others come at the end. Thus there is another flag option for Features, under the same namespace: `draw-steel.embedDisplay`, under the `displayAtEnd` property. This is a `Boolean` that exists on the Feature. Set it to `true` to put that Feature after the Abilities, rather than before. The order is otherwise retained.
-- You can set this flag on an item in Foundry via `item.setFlag("draw-steel", "embedDisplay", { iconOverride: "special", displayAtEnd: true  }`
+- You can set this flag on an item in Foundry via `item.setFlag("draw-steel", "embedDisplay", { iconOverride: "special", displayAtEnd: true  })`
+- Abilities with the "Ancestry Malice" category will not be shown unless you use `showAllItems=true` in the embed configuration (i.e. @Embed&ZeroWidthSpace;[uuid showAllItems=true])
```
krbz999: I like just `showAll=true`.

CHANGELOG.md
```diff
@@ -20,14 +20,30 @@
 
 ## 1.1.1
 
+### Added
+
+- New Player-Facing Compendium Content:
+  - Added Armor & Implement imbues to the Effects compendium
+- Added an "Ancestry Malice" category for abilities which will be skipped by the embed. Using `showAllItems=true` in the embed call will include the abilities anyways.
```
krbz999:
```suggestion
- Added an "Ancestry Malice" category for abilities that are from a monster's ancestry, not from its unique statblock. Ancestry malice abilities are skipped in the actor embed unless the embed is supplied the `showAll=true` option.
```

krbz999 (approved):

PR #1997, opened 2026-07-16
Add imbue implement AEs

cdverrett94:
There are a couple of these implements (Seeking and Thought Sending) that need to be ability modifiers, but it would require doing 2 AEs for them to do the magic OR psionic. So i've just left it as one base AE for now


JPMeehan (approved):
LGTM

Issue #583, opened 2025-06-12
Implement Summoning


JPMeehan:
- Conduit
  - Gods' Machine
  - Divine Dragon
- Elementalist
  - Source of Earth
  - Heart of the Wode
- Beastheart
- Summoner
- Familiar Perk

### Basic Flow

- use ability with some data for summoning
- Ability stores UUID of compendium actor. Scan `game.actors` for a matching actor (or cache?)
  - with the right `compendiumSource`
  - with a `flags.draw-steel.summonSource` value matching the item's UUID
- If no actor is found, perform import from compendium (requires Actor Creation rights)
- Then prompt placement. 
  - Summoner minions scale with the summoner, may need to perform updates/create effects. Need framework for this scaling.
- After placement, add to combat tracker and potentially combat groups (Group creation is unfortunately locked behind GM permission and will require sockets). Check flags again for smooth creation.

KJTailor:
Might be worth while to implement on monsters too? there are quite few that summon minions into battle, eg. Goblin Monarch, Gnoll Abyssal Summoner, Gnoll Carnage, Demon Lumbering Egress

JPMeehan:
Some notes on the Heroes book summons
- Divine Dragon characteristics are tied to the Conduit's
- Gods' Machine might conjure an object? No characteristics. Also has the same Elementalist bit of repeating a previous power roll.
- Source of Earth can pop in and out during an encounter but maintains stamina for the fight, meaning a linked actor.
- Familiar only returns during respites and is thus the simplest/doesn't need automation support

JPMeehan:
Summoner notes
- Primary summons are minions, which means need smooth addition into squads 
- Summoning range is dynamic (5 + Reason)
- Get 2 minions at the start of encounter, 3 at the start of turn
- Can use surges with their attacks
- All summoning flows through the "Call Forth" ability
- Level 2 gives the ability to summon a "fixture", which improves at higher levels. Need to decide if the fixture itself has advancements or we just make versions for each level.
- Summon Champion at 9 is not really a minion

PR #2019, opened 2026-07-22
[#583] Summoning Data Model & Sheet

### LINKED ISSUE #2020: Implement basic summoning
Summoning from a fixed list, unrelated to actor choice advancements, like the Elementalists "Source of Earth" or the Gnoll Summoners ability to conjure Abyssal Hyenas

JPMeehan:
Implements fixed-actor summoning logic as a new Special Effect.

image

Implementation questions:
- How about abilities with Spend logic that replace what's summoned
- Call Forth as a "Portfolio" summon could be another subtype, as could the companion. The common thread is that effect sudokus figure out *which* actor to summon, and then the ability model method does the actual handling of reusing/importing the actor.

### DIFF
--- a/src/module/data/item/ability.mjs
+++ b/src/module/data/item/ability.mjs
@@ -891,4 +891,36 @@ export default class AbilityModel extends BaseItemModel {
     }
     return false;
   }
+
+  /* -------------------------------------------------- */
+
+  /**
+     * Places summons.
+     * @param {string} uuid               The UUID of the actor to summon. If this points to a compendium actor a copy will be imported.
+     * @param {Object} options
+     * @param {number} [options.count=1]  How many tokens to summon.
+     * @returns {Promise<DrawSteelTokenDocument[] | null>} Returns null if the user did not have permissions.
+     */
+  async performSummon(uuid, { count = 1 }) {
+    /** @type {DrawSteelActor} */
+    const sourceActor = await fromUuid(uuid);
+
+    if (!sourceActor) return void ui.notifications.error("");
+
+    let worldActor = game.actors.find(a => (a._stats.compendiumSource === uuid) && (a.getFlag(systemID, "summonSource") === this.document.uuid));
+
+    if (!worldActor) {
+      // Ensure the user has permission to drop the actor and create a Token.
+      if (!game.user.can("ACTOR_CREATE")) {
+        ui.notifications.warn("DRAW_STEEL.Actor.Summoning.Errors.ACTOR_CREATE", { localize: true });
+        return null;
+      }
+
+      worldActor = game.actors.importFromCompendium(sourceActor.pack, sourceActor.id, {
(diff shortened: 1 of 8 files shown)


src/module/data/item/ability.mjs
```diff
@@ -884,4 +884,36 @@ export default class AbilityModel extends BaseItemModel {
...
+
+  /* -------------------------------------------------- */
+
+  /**
+     * Places summons.
+     * @param {string} uuid               The UUID of the actor to summon. If this points to a compendium actor a copy will be imported.
+     * @param {Object} options
+     * @param {number} [options.count=1]  How many tokens to summon.
+     * @returns {Promise<DrawSteelTokenDocument[] | null>} Returns null if the user did not have permissions.
+     */
```
```suggestion
  /**
   * Places summons.
   * @param {string} uuid               The UUID of the actor to summon. If this points to a compendium actor a copy will be imported.
   * @param {Object} options
   * @param {number} [options.count=1]  How many tokens to summon.
   * @returns {Promise<DrawSteelTokenDocument[] | null>} Returns null if the user did not have permissions.
   */
```

src/module/data/item/ability.mjs
```diff
@@ -884,4 +884,36 @@ export default class AbilityModel extends BaseItemModel {
...
+  async performSummon(uuid, { count = 1 }) {
+    /** @type {DrawSteelActor} */
+    const sourceActor = await fromUuid(uuid);
+
+    if (!sourceActor) return void ui.notifications.error("");
```
krbz999: Missing warning.

src/module/data/item/ability.mjs
```diff
@@ -884,4 +884,36 @@ export default class AbilityModel extends BaseItemModel {
...
+
+    let worldActor = game.actors.find(a => (a._stats.compendiumSource === uuid) && (a.getFlag(systemID, "summonSource") === this.document.uuid));
```
krbz999: Do we want to add a registry for tracking summons, having actors assign themselves in their `prepareData` method? It would be more efficient than searching the full actor collection.

That would be fine for a different PR of course.

src/module/data/pseudo-documents/special-effect/summon-effect.mjs
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Places summons.
+   * @returns {Promise<DrawSteelTokenDocument[] | null>} Returns null if the user did not have permissions.
+   */
+  async performSummon() {
+    if (!this.pool.size) return void ui.notifications.error("");
```
krbz999: Missing warning.

The model might also have uuids pointing to actors that do not exist. We should grab them with `fromUuidSync` first and filter.

krbz999 (approved):

PR #2026, opened 2026-07-26
[#1962] DSDialog styling

### LINKED ISSUE #1962: `DSDialog` Styling
The next large styling overhaul here will be the `DSDialog` sheets:

#### Tasks: 
- [ ] Create a Style Guide (in Journal) for `DSDialog` which contains every Form form elements (inputs/fields/buttons/etc).
  - This can be done on a step-by-step basis as you style each of the following element.
- [ ] Style each form element:
  - [ ] `<input type="number">`
  - [ ] `<input type="text">`
  - [ ] `<input type="checkbox">`
  - [ ] `<select>`
  - [ ] `<button>`
  - [ ] `HTMLColorPickerElement`
  - [ ] `HTMLFilePickerElement`
  - [ ] `HTMLMultiCheckboxElement`
  - [ ] `HTMLMultiSelectElement
  - [ ] `HTMLProseMirrorElement`
  - [ ] `HTMLRangePickerElement`
  - [ ] `HTMLStringTagsElement`
  - [ ] Any custom elements  / layout in our forms
  - [ ] Any other elements I couldn't think of

Having styling for all of these elements will allow us to *also* plug them in to to other locations, like the Compendium Browser, Character Sheets, TokenHUD, and ChatMessages, in the future.

We haven't exactly decided how this styling should look, beyond just making it look nicer than it is currently. That can be a part of the discussion on this issue.

Katiedead:
# Adds styling for various form objects in both light and dark mode

No additional HTML changes are required anywhere. These are scoped to `.draw-steel` and do not need classes applied to them. This resolves #1962 

## Examples
image
image


krbz999:
We should prevent it from applying to the actor and item sheets.

Using the `draw-steel` class would be preferred, so maybe we remove that class where we don't yet want it?

JPMeehan:
I think a `:not` makes the most sense? Plays nicer with subclassing if people want to strip the styling since DEFAULT_OPTIONS.classes only appends

jalensailin:
Let's minimize the gap between the label and the input for stacked `form-group`s (see Immunities/Weaknesses). We can also center-align their labels.

src/styles/system/applications/sheets/actor-sheet.css
```diff
@@ -30,9 +30,9 @@
       .tags {
         gap: 5px;
         .tag {
-          background: var(--draw-steel-c-faint);
-          border: 1px solid var(--draw-steel-c-beige);
-          color: var(--draw-steel-c-dark);
+          /* background: var(--draw-steel-c-faint); */
+          /* border: 1px solid var(--draw-steel-c-beige); */
+          /* color: var(--draw-steel-c-dark); */
```
krbz999: We'll (probably?) keep these for now since we'll target a different selector.

krbz999 (approved):
All looks good to me, here's the steps we want to finalize it:
- Target `.application.draw-steel:not(.ds-sheet)`.
- Make adjustments for `div.form-group.stacked` elements: less gap between `label` and `div.form-fields` and center the `label`.

JPMeehan:
> All looks good to me, here's the steps we want to finalize it:
> 
>     * Target `.application.draw-steel:not(.ds-sheet)`.
> 
>     * Make adjustments for `div.form-group.stacked` elements: less gap between `label` and `div.form-fields` and center the `label`.

also need to add ds-sheet to the DSDocumentSheet.DEFAULT_OPTIONS

Katiedead:
> > All looks good to me, here's the steps we want to finalize it:
> > ```
> > * Target `.application.draw-steel:not(.ds-sheet)`.
> > 
> > * Make adjustments for `div.form-group.stacked` elements: less gap between `label` and `div.form-fields` and center the `label`.
> > ```
> 
> also need to add ds-sheet to the DSDocumentSheet.DEFAULT_OPTIONS

Clarification, do you want all of the form changes to target `.application.draw-steel:not(.ds-sheet)` so no form changes touch the character sheets? Like so:

image

If so then that is what I've done with this last commit.

JPMeehan:
That is mostly correct but those field set backgrounds don't look quite right, is that what they look like before this PR? Basically want to have these new changes hit everything *except* our primary actor and item sheets

Katiedead:
Missed it as it wasn't fully grouped with the others. Corrected now.
image

JPMeehan:
Wait, what happened with the tags at the top? Those had *some* styling on live (Dragon Knight/Aristocrat)

Katiedead:
> Wait, what happened with the tags at the top? Those had _some_ styling on live (Dragon Knight/Aristocrat)

Let me go through and double check I have everything I changed behind the namespacing. Thought it was all in one spot, sorry.

Katiedead:
Want me to revert this as well? I had changed the svg coloring for the effects to better match the light / dark mode of the character sheets. I can just leave the change commented out or leave the change active. It's mainly for the light mode which I think was overlooked. This is the only other thing that should be different I think.

If you're happy keeping it, then this is ready to be merged unless I overlooked another thing.

Default
image

image


With change
image

image

JPMeehan:
I think I'm happy with that change, we'll mess around with it more later I'm sure.

krbz999 (changes requested):
Some issues with contrast. In dark mode, these text inputs are almost entirely invisible.

In this same screenshot, I also disabled some of the inputs to see the difference. It doesn't appear as if there is any.

The same goes for light mode.

Design question: is it intentional that text/number inputs do not have the corners that selects do?

Image

Image

krbz999 (approved):

PR #2031, opened 2026-08-02
Fixed with captain effect field setup

JPMeehan:
- The form input would not return to null when picking blank option
- Local Document Field was treating options as mandatory
- Unclear why there was this or usage in the model attachment

### DIFF
--- a/src/module/data/fields/local-document-field.mjs
+++ b/src/module/data/fields/local-document-field.mjs
@@ -1,16 +1,7 @@
 /**
- * @import { StringFieldOptions, DataFieldContext } from "@client/data/_types.mjs"
+ * @import { DataFieldContext } from "@client/data/_types.mjs"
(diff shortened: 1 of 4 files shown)


src/module/data/fields/local-document-field.mjs
```diff
@@ -22,7 +22,7 @@ export default class LocalDocumentField extends foundry.data.fields.DocumentIdFi
    * @param {LocalDocumentFieldOptions} options   Options which configure the behavior of the field.
    * @param {DataFieldContext} [context]          Additional context which describes the field.
    */
-  constructor(model, options, context = {}) {
+  constructor(model, options = {}, context = {}) {
```
krbz999, line 16: fix [options] in the `@param` too

krbz999 (approved):

PR #2034, opened 2026-08-02
[#2024] Adjusted size update propagation

### LINKED ISSUE #2024: Actor size not propogating to token depth
### What happened?

Hello! It looks like when changing the size of an actor, it's not actually propagated to that creature's depth, despite the Active Effects docs saying it does. 

Relevant codelines:
https://github.com/MetaMorphic-Digital/draw-steel/blob/d9fcf1234d035cbcba107f61c920e703502b1cae/src/module/data/actor/base-actor.mjs#L255-L262

https://github.com/MetaMorphic-Digital/draw-steel/blob/d9fcf1234d035cbcba107f61c920e703502b1cae/src/docs/Active-Effects.md?plain=1#L248

### Reproduction Steps

Change an actor's size value to a different integer.

Draw Steel 1.1.0

### Bug Checklist

JPMeehan:
I'm unsure on which way to go; I think it's more useful to keep this prototype token update than not, but maybe we should *only* rely on the tokenActiveEffectChanges solution we have in CreatureModel#prepareDerivedData?

Closes #2024

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,7 @@ Increased foundry minimum to 14.366.
 - Retainers no longer show level-up when equal in level to their mentor. (#2015)
 - Dropping a class on a retainer will now prompt to delete the existing one. (#2030)
 - The "With Captain" select for minions can now be cleared.
+- Updating an actor's size will now also update their prototype token's depth, in addition to height and width. (#2024)
 
 ## 1.1.1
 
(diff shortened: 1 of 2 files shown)


krbz999:
Well if we can rely on just one implementation, we should do that. Gonna move to 1.2.0 instead if there's anything to discuss about this.

JPMeehan:
Yeah I'm mixed on which is more correct, I think it's a little weird to so heavily rely on the persistent size updates but it's not like there's a meaningful performance cost.

krbz999 (approved):

Issue #1869, opened 2026-04-19
Display Potency Results


JPMeehan:
As a user, I would like to easily be able to see the potency comparison for any power roll effects.

JPMeehan:
Related to #509

JPMeehan:
Should be implemented as part of #1936

Issue #1320, opened 2025-10-30
Implement Squad Actions


JPMeehan:
As a director, I would like minion attacks that target a single individual to scale with the number of other minions joining based on the minion's free strike value, to better adjust for possible immunities.

JPMeehan:
This probably needs to wait until we design a comprehensive usage dialog, which I don't think is on the docket until 0.11

JPMeehan:
I think this requires #1628

PR #2045, opened 2026-08-10
[#1628] Implemented ability target part

### LINKED ISSUE #1628: Implement per-target chat message parts
As a player, I would like to see per-target ability results that then have a button to specifically apply the results of the ability to that target with the full ability to evaluate forced movement and potency adjustments based on that target.
___
Original: Hang buttons off chat messages
The chat buttons are currently very large and unwieldy. The implementation of anchor positioning in the last year means that as of v14 we can reliably use it to construct hanging elements. Moving various chat buttons to hang off of the chat message will free up visual real estate.

JPMeehan:
Implements a new part that handles results by target rather than our current method of grouping by tier result. Currently just rewires the existing buttons to apply to only the target rather than fully implementing a "Confirm Effects" type logic. This unblocks significant other work such as #1320 as well as #1869.

image

Closes #1628

### DIFF
--- a/src/module/config.mjs
+++ b/src/module/config.mjs
@@ -2108,6 +2108,9 @@ export const MessagePart = {
   savingThrow: {
     documentClass: pseudoDocuments.messageParts.SavingThrowPart,
   },
+  targetResult: {
+    documentClass: pseudoDocuments.messageParts.TargetResult,
+  },
   test: {
     documentClass: pseudoDocuments.messageParts.TestPart,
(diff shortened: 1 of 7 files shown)


JPMeehan:
I did not implement any kind of double-apply protection in this PR, I kinda figured that would just come from the PR that actually implements a holistic apply button, which I think is the one after potency checks?

krbz999 (approved):

PR #2047, opened 2026-08-12
[#363] Flying or burrowing applies the status effect

### LINKED ISSUE #363: Consider relationship of statuses and movement types
Core provides two "special" statuses, Burrowing and Flying, out of the box. We should consider how these interact with the token flags for movement type; it's clear that something like "teleporting" shouldn't be a status, but it may be separately useful to track what a creature is up to. For example, prone (also covered in #246) might be used as a gate for Crawling.

JPMeehan:
Foundry (and the Draw Steel Rules) assign special meaning to being flying or burrowing. Using these movement modes will apply the appropriate status effect to mark an actor as flying/burrowing.

Closes #363


krbz999 (dismissed):
At the end of the movement, the token might not be flying anymore; the movement action can reset to the default.

When moving again, one can end up with both flying and burrowing by first flying, then burrowing in two separate operations.

JPMeehan:
Should we only evaluate the last waypoint? I feel like flying + burrowing is just a question of people properly toggling things off, which is easier to remember because they'll have the token icon.

krbz999:
The ideal would be to have the `flying/burrowing` statuses during any relevant movement segment.

I'm not really sure, I feel like this implementation is a half-measure.

JPMeehan:
The point of the statuses, both I think within Foundry as well as the DS rules, is to track creatures that are flying/burrowed in between turns/when they're not actively moving. For example, Foundry doesn't provide a native way to cleanly distinguish height from standing on a table/at the top of a hill and height from flying, which we would care about for the "High Ground" bonus and for falling respectively.

krbz999:
I don't love it. But perhaps we simply give it a try and await feedback post-release.

PR #2056, opened 2026-08-17
[#1871] Track maximum HR within turn cycle

### LINKED ISSUE #1871: Property to track max HR within a turn
### User Experience

Hi,

It would be nice to have a property available that tracks the maximum your heroic resource reached during the current turn, or maybe since the end of your last turn.  

The use case is for class features like the Fury's Growing Ferocity and Null's Discipline Mastery where the benefits of the feature are not based on the current heroic resource, but the highest value of the heroic resource in the current turn (maybe).  

ChaosOS and I had a [brief conversation about it in the discord](https://discord.com/channels/332362513368875008/1342298358664138805/1495466123540304054).   I've reread the feature and I think that the text is clear that you reset the max counter to the current value at the end of the character's own turn.  I'm less clear on if the counter resets at the end of each turn in between or not.  But regardless of whether the reset is implemented each turn or just on your turn, I guess anytime you gain HR it would check if your current HR is now higher than the current maximum and update the maximum if needed.  

Thanks!

JPMeehan:
Some classes, like the Fury and Null, care about piling up resources. However, the benefits of you stockpiling your heroic resource don't expire until the end of your turn, which means we need to track the highest those resources go and then only reset the value at the end of turn. To do this I've implemented a generalized _onEndTurn logic for actor data models.

By itself, this PR doesn't actually accomplish much, but it sets up for #702 where we'd actually have data on AEs that suppresses them based on the actor conditions.

A follow-up question is that since we fixed the general combat order logic, it might make more sense to have our _onStartTurn trigger care about the combat tracker defined start of turn, rather than the initiative value (# of turns remaining this round) decreasing.

Closes #1871

### DIFF
--- a/src/module/data/actor/hero.mjs
+++ b/src/module/data/actor/hero.mjs
@@ -313,7 +313,7 @@ export default class HeroModel extends CreatureModel {
   /** @inheritdoc */
   async _onEndTurn(combatant) {
     await super._onEndTurn(combatant);
-    await this.parent.update({ "system.hero.primary.tracking": this.hero.primary.value });
+    await this.parent.update({ "system.hero.primary.tracking": this.hero.primary.value }, { render: false });
   }
 
   /* -------------------------------------------------- */
--- a/src/module/data/actor/hero.mjs
+++ b/src/module/data/actor/hero.mjs
@@ -62,6 +62,7 @@ export default class HeroModel extends CreatureModel {
     schema.hero = new fields.SchemaField({
       primary: new fields.SchemaField({
         value: new fields.NumberField({ initial: 0, integer: true, nullable: false }),
+        tracking: new fields.NumberField({ initial: 0, integer: true, nullable: false }),
       }),
       epic: new fields.SchemaField({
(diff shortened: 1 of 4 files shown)


krbz999 (commented):
Is this really a value we need to store in the schema? Could it be assigned at each turn change?

JPMeehan:
I guess my question is "assigned where?"; I'd rather not just have some floating in-memory assignment in case someone disconnects mid-combat.

src/module/data/actor/hero.mjs
```diff
@@ -298,6 +310,14 @@ export default class HeroModel extends CreatureModel {
 
   /* -------------------------------------------------- */
 
+  /** @inheritdoc */
+  async _onEndTurn(combatant) {
+    await super._onEndTurn(combatant);
+    await this.parent.update({ "system.hero.primary.tracking": this.hero.primary.value });
```
krbz999: Perhaps with `render: false`

krbz999 (approved):

PR #2057, opened 2026-08-17
[#348] Automatically toggle with captain effects

### LINKED ISSUE #348: Implement "With Captain" condition for Active Effects

JPMeehan:
Main design decision was to just use the `disabled` property because we auto-delete `expired` effects as part of combat cleanup.

Closes #348


krbz999 (approved):
I've pushed a commit to avoid iterating over the same actor more than once (which would never happen but could be possible).

PR #2066, opened 2026-08-24
Reconfigure Button

### LINKED ISSUE #1513: Add button to repick item
As a user, I would like a header button to redo the advancement that granted an item.

Lusunati:
As described in #1513, added header button that reconfigures an advancement an item belongs to. Is not present unless it belongs to an advancement.

image

closes #1513

### DIFF
--- a/src/module/applications/sheets/item-sheet.mjs
+++ b/src/module/applications/sheets/item-sheet.mjs
@@ -757,7 +757,7 @@ export default class DrawSteelItemSheet extends DSDocumentSheet {
    * 
    */
   static #canReconfigAdvancement() {
-    const advancementSource = this.item.parent?.items.get(this.item.getFlag("draw-steel.advancement", "parentId")
+    const advancementSource = this.item.parent?.items.get(this.item.getFlag("draw-steel.advancement", "parentId");
     if (!advancementSource) return false;
     return !!advancementSource.pseudoCollections.Advancement.get(this.item.getFlag("draw-steel.advancement", "advancementId");
   }
--- a/src/module/applications/sheets/item-sheet.mjs
+++ b/src/module/applications/sheets/item-sheet.mjs
@@ -757,7 +757,9 @@ export default class DrawSteelItemSheet extends DSDocumentSheet {
    * 
    */
   static #canReconfigAdvancement() {
-    return (this.item.parent && (this.item.flags["draw-steel"] != undefined));
+    const advancementSource = this.item.parent?.items.get(this.item.getFlag("draw-steel.advancement", "parentId")
+    if (!advancementSource) return false;
+    return !!advancementSource.pseudoCollections.Advancement.get(this.item.getFlag("draw-steel.advancement", "advancementId");
   }
 
   /**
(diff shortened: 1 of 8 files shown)


src/module/applications/sheets/item-sheet.mjs
```diff
@@ -744,6 +750,27 @@ export default class DrawSteelItemSheet extends DSDocumentSheet {
...
+  }
+
+  /**
+   * Finds which advanceemnt an item came from, calls the advancement's reconfigure.
+   * @this DrawSteelItemSheet
+   * @private
+   */
+  static async #rewindAdvancement() {
+    const advancementSource = this.item.parent.items.get(this.item.flags["draw-steel"].advancement.parentId);
+    const advancement = advancementSource.pseudoCollections.Advancement.get(this.item.flags["draw-steel"].advancement["advancementId"]);
```
JPMeehan:
```suggestion
    const advancementSource = this.item.parent.items.get(this.item.getFlag("draw-steel.advancement", "parentId");
    const advancement = advancementSource.pseudoCollections.Advancement.get(this.item.getFlag("draw-steel.advancement", "advancementId");
```

`getFlag` is a wrapper for the optional chaining here that ensures no errors.

src/module/applications/sheets/item-sheet.mjs
```diff
@@ -744,6 +750,27 @@ export default class DrawSteelItemSheet extends DSDocumentSheet {
...
   }
 
+  /**
+   * If an item has an advancement that can be reconfigured.
+   * @this DrawSteelItemSheet
+   * @returns {bool}
+   * 
+   */
+  static #canReconfigAdvancement() {
+    return (this.item.parent && (this.item.flags["draw-steel"] != undefined));
```
JPMeehan:
```suggestion
    const advancementSource = this.item.parent?.items.get(this.item.getFlag("draw-steel.advancement", "parentId")
    if (!advancementSource) return false;
    return !!advancementSource.pseudoCollections.Advancement.get(this.item.getFlag("draw-steel.advancement", "advancementId");
```

We want to directly check for the attributes we care about, which means doing this access. It's all very efficient O(1) lookups so we don't care about redoing this work

JPMeehan (approved):
Looks good just some notes on accessing flags. Also looks like your formatter caught some inconsistencies in the .hbs files so I'll take those.

src/module/applications/sheets/item-sheet.mjs
```diff
@@ -744,6 +750,29 @@ export default class DrawSteelItemSheet extends DSDocumentSheet {
...
   }
 
+  /**
+   * If an item has an advancement that can be reconfigured.
+   * @this DrawSteelItemSheet
+   * @returns {bool}
+   * 
+   */
+  static #canReconfigAdvancement() {
+    const advancementSource = this.item.parent?.items.get(this.item.getFlag("draw-steel.advancement", "parentId")
```
JPMeehan:
```suggestion
    const advancementSource = this.item.parent?.items.get(this.item.getFlag("draw-steel.advancement", "parentId");
```

Issue #2021, opened 2026-07-23
Implement Portfolio Summoning


JPMeehan:
Summoning that pulls from an Actor Choice advancement, like the Summoner or Beastheart

PR #2071, opened 2026-08-25
[#2021] Add Portfolio Summoning

JPMeehan:
Portfolios now fully integrate into the summoning API by adding another special effect for them + doing more work in data prep. If we want we can loosen up from strict `item.dsid` matching to registering portfolios in a follow-up PR; might make sense for handling the start-of-turn summons which aren't technically part of Call Forth.

image

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -1082,7 +1082,13 @@
           "count": {
             "label": "Count",
             "hint": "How many signature minions to summon?"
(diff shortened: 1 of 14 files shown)


krbz999 (approved):
Calling fromUuid on first render is a bit of a heavy one which we might want to improve in the future.

Issue #2073, opened 2026-08-26
Summoner Compendium


Lusunati:
### User Experience

As a player I'd like the summoner implemented as a compendium so i can easily create and play one.

Lusunati:
#2075 implements summoner ability keywords.

Lusunati:
Current thinking for Summon Portfolios in the "Summons & Companions" Compendium
```
Compendium Root/
├─ Summoner/
│  ├─ Portfolio/
│  │  ├─ X-Essence Minion/
│  │  │  ├─ Minion
│  │  ├─ Fixture
│  │  ├─ Champion
```

Lusunati:
#2077 adds signature minions
#2080 adds 3-cost minions

Lusunati:
Remaining Tasks (Last Dregs of Data Input):
- [ ] Implement Devil Detective
- [ ] Implement Rival Summoner
- [ ] Enrich any titles that enable one time summoning or portfolio additions once summon enricher is added.

PR #2080, opened 2026-08-27
Summoner Compendium: 3-Essence Minions

Lusunati:
As per #2073, added the Signature Minions for each portfolio from Speedrace's compendium, changed to fit the style guide, added icons to the features.

Potencies changed to use regular enrichers, including in abilities and restored highest attribute to power rolls as these should be changed automatically by the summoner.


src/packs/summons/Summoner_K3NKz0H5uCvXk35S/Demon_Portfolio_HS8VOC5sBgutnpR7/3_Essence_Minions_grTBqFuzUL7VN4Ju/npc_Twisted_Bengrul_a9YKT6xoII95LRt3.json
```diff
@@ -0,0 +1,569 @@
...
+      "ownership": {
+        "default": 0,
+        "bULgfeSVOL2yapd0": 3
+      },
+      "folder": null,
+      "sort": 0,
+      "_key": "!actors.items!a9YKT6xoII95LRt3.xxpvEdqJkB4YNJyq"
+    },
+    {
+      "name": "Mind Twist",
```
JPMeehan, line 290: - Duplicate applied effect should be cleaned up
- Need the applied effect on all tiers
- Icon for effect should probably be picked out from the magic/unholy folder
- Uncheck the "apply to actor" so it doesn't show up on the minion's sheet

Image

JPMeehan (changes requested):
Clean up the Twisted Bengrul, otherwise other improvements will need some feature dev to accomodate

JPMeehan:
Also need the twisted condition attached to all three tiers of the Mind Twist ability

JPMeehan (approved):

PR #2083, opened 2026-08-28
Summoner Integration - Minion/NPC Free Strike overrides.

### LINKED ISSUE #2081: Allow adjusting free strike damage types without signature
Many summoner minions don't have a signature ability but do have features that modify the damage type of their free strikes. A data path or flag for this would be an appropriate fix via AE.

Lusunati:
Now using active effects via `monsters.freeStrikeType `and using the minion sheet setup, players can set a damage type for Free Strikes.

Closes #2081


package-lock.json
JPMeehan, line 1: Not sure why this got deleted

JPMeehan (commented):
Other than deleting package-lock looks good. Could you also lint `src/module/config.mjs`? Some of the new keyword logic from a previous PR didn't have appropriate spaces

package-lock.json
Lusunati: Done so and reverted the change

JPMeehan (approved):
Looks good

PR #2085, opened 2026-08-30
Add applied effect tier 1 copy

### LINKED ISSUE #1142: Button/Toggle to add Active Effect to ALL roll tiers
### User Experience

As a game master I would like to be able to, with a single click, apply a given status or active effect to all tiers of a power roll. Doing so would drastically speed up the creation of abilities with rider effects, especially where there may be a need to override the end duration, potency condition, or properties across the board. 

In the current implementation, I have to go down for each tier and re-apply the effect (especially if it's a custom one). 

Ideas for implementation/improvement:
- Add a button next to the "+" for adding effect to "copy all effects to all tiers". This has the added benefit of being able to change your mind later by deleting it from a given tier.
- Alternatively, add a checkbox (boolean) for a give effect to propagate it to all tiers.

mrafaj:
Adds a button to the "Power Roll Effect: Applied Effect" sheet that will copy the applied effects for Tier 1 to the current Tier. This should speed up creating abilities that apply effects on their power roll, since most power rolls apply the same effects on all 3 tiers (often even with the same duration).

Also includes a minor compendium fix for "Blood for Blood" applied effects (this was used to test the button).

Picture of button as added (with tooltip shown):

Foundry_Virtual_Tabletop_NT0exdozUA

### DIFF
--- a/src/module/applications/sheets/pseudo-documents/power-roll-effect-sheet.mjs
+++ b/src/module/applications/sheets/pseudo-documents/power-roll-effect-sheet.mjs
@@ -15,7 +15,7 @@ export default class PowerRollEffectSheet extends PseudoDocumentSheet {
   static DEFAULT_OPTIONS = {
     actions: {
       addAppliedEffect: this.#addAppliedEffect,
-      addTierOneEffect: this.#addTierOneEffect,
+      copyTierOneEffect: this.#copyTierOneEffect,
       deleteAppliedEffectEntry: this.#deleteAppliedEffectEntry,
       editAppliedEffect: this.#editAppliedEffect,
     },
@@ -112,13 +112,15 @@ export default class PowerRollEffectSheet extends PseudoDocumentSheet {
     }
   }
 
+  /* -------------------------------------------------- */
+
   /**
    * Add an entry to `applied.effects`.
    * @this PowerRollEffectSheet
    * @param {PointerEvent} event        The initiating click event.
    * @param {HTMLButtonElement} target  The capturing HTML element which defined a [data-action].
    */
-  static async #addTierOneEffect(event, target) {
+  static async #copyTierOneEffect(event, target) {
     const path = target.dataset.path;
     if (path.includes("1")) return;
     this.pseudoDocument.update({ [`${path}`]: this.pseudoDocument.applied.tier1.effects });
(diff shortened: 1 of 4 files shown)


templates/sheets/pseudo-documents/power-roll-effect-sheet/details-tiers.hbs
```diff
@@ -33,6 +33,7 @@
       </select>
     </div>
     <button type="button" class="icon fa-solid fa-plus" data-action="addAppliedEffect" data-path="{{effects.name}}"></button>
+    <button type="button" class="icon fa-solid fa-copy" data-action="addTierOneEffect" data-path="{{effects.name}}" data-tooltip="DRAW_STEEL.POWER_ROLL_EFFECT.APPLIED.CopyTierOneEffect" data-tooltip-direction="UP"></button>
```
JPMeehan: I think you can expand some of the tier context in `AppliedPowerRollEffect_#tierRenderingContext` so this button only shows on tiers 2 & 3 (probably use `#unless` but `#if` is find, just keep track of which boolean value you're expecting)

src/module/applications/sheets/pseudo-documents/power-roll-effect-sheet.mjs
```diff
@@ -111,6 +112,18 @@ export default class PowerRollEffectSheet extends PseudoDocumentSheet {
     }
   }
 
+  /**
```
JPMeehan, line 117: Copy the -- divider so it's above this block

JPMeehan (commented):
Most of the way there but I'd rather just not show the button for the tier 1 section.

templates/sheets/pseudo-documents/power-roll-effect-sheet/details-tiers.hbs
mrafaj: Done.

src/module/applications/sheets/pseudo-documents/power-roll-effect-sheet.mjs
mrafaj: Done.

JPMeehan (approved):
LGTM

PR #2086, opened 2026-08-30
[#2063] Added swap melee/ranged button to ability configuration dialog

### LINKED ISSUE #2063: Toggle melee/ranged usage in roll UI
### User Experience

As a player and game master, I'd love it if it were possible for abilities with a melee/ranged usage to be able to toggle between those in the ability roll UI itself.

This would help to not only highlight which mode it currently _is_ in, but also reduce the number of clicks needed to swap between modes if you accidentally start the roll in the wrong mode. 

This is helpful especially when running for players that take a kit that does _not_ have a damage bonus to one or the other, as I find we often forget to swap between in the few scenarios where they _are_ making the other strike type, which leads to them either having less or more damage than expected.

JPMeehan:
image

image

Closes #2063


templates/apps/ability-configuration-dialog/ability.hbs
```diff
@@ -1,5 +1,14 @@
 
 <section class="ability-config">
+  {{#if meleeRanged.show}}
+  <div class="form-group meleeRanged">
+    <span class="{{#if meleeRanged.isMelee}}label{{/if}}">{{localize "DRAW_STEEL.Item.ability.Distance.Melee"}}</span>
+    <button type="button" data-action="swapMeleeRanged" data-tooltip="{{ifThen meleeRanged.isMelee "DRAW_STEEL.Item.ability.SwapUsage.ToRanged" "DRAW_STEEL.Item.ability.SwapUsage.ToMelee"}}">
+      <i class="fa-solid {{ifThen meleeRanged.isMelee "fa-sword" "fa-bow-arrow"}}"></i>
```
krbz999:
```suggestion
      <i class="fa-solid {{ifThen meleeRanged.isMelee "fa-sword" "fa-bow-arrow"}}" inert></i>
```

src/module/applications/apps/ability-configuration-dialog.mjs
```diff
@@ -290,4 +296,17 @@ export default class AbilityConfigurationDialog extends PowerRollDialog {
...
+  /* -------------------------------------------------- */
+
+  /**
+   * Switch if this ability is going to roll as melee or ranged.
+   * @this AbilityConfigurationDialog
+   * @param {PointerEvent} event   The originating click event.
+   * @param {HTMLElement} target   The capturing HTML element which defined a [data-action].
+   */
+  static async #swapMeleeRanged(event, target) {
+    await this.item.update({ "system.damageDisplay": this.item.system.damageDisplay === "melee" ? "ranged" : "melee" });
```
krbz999: I don't believe we clone the item. Maybe that'd be a good idea to do rather than spurious updates.

krbz999 (approved):
Seems fine. I think I'd prefer a different layout of the buttons; a row of anchor elements you select one of, disabling the rest.

JPMeehan:
image
Updated visuals

PR #2087, opened 2026-08-30
[#2050] Implemented MapLocationJournalPage

### LINKED ISSUE #2050: Customize Journal Numbers
As a premium content creator, I would like more control over the numbering of journal entries in the journal contents listing

JPMeehan:
Implements the requested functionality to customize TOC labels

Closes #2050

### DIFF
--- a/src/module/applications/sheets/journal-entry-sheet.mjs
+++ b/src/module/applications/sheets/journal-entry-sheet.mjs
@@ -68,4 +68,22 @@ export default class DrawSteelJournalEntrySheet extends foundry.applications.she
     li.append(anchor);
     return li;
   }
+
+  /* -------------------------------------------------- */
+
+  /** @inheritDoc */
+  _preparePageData() {
+    const pages = super._preparePageData();
+    let adjustment = 0;
+    for (const descriptor of Object.values(pages)) {
+      const page = this.document.pages.get(descriptor.id);
+      const numbering = page?.system.adjustTOCNumbering?.(descriptor.number);
+      if (numbering) {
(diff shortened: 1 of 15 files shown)


src/module/applications/sheets/journal-entry-sheet.mjs
```diff
@@ -68,4 +68,22 @@ export default class DrawSteelJournalEntrySheet extends foundry.applications.she
     li.append(anchor);
     return li;
   }
+
+  /* -------------------------------------------------- */
+
+  /** @inheritDoc */
```
krbz999, line 74:
```suggestion
  /** @inheritdoc */
```

src/module/applications/sheets/journal-entry-sheet.mjs
```diff
@@ -68,4 +68,22 @@ export default class DrawSteelJournalEntrySheet extends foundry.applications.she
...
+  _preparePageData() {
+    const pages = super._preparePageData();
+    let adjustment = 0;
+    for (const descriptor of Object.values(pages)) {
+      const page = this.document.pages.get(descriptor.id);
+      const numbering = page?.system.adjustTOCNumbering?.(descriptor.number);
```
krbz999, line 80: In ryuutama, I went with `prepareCode`. We'd also have to be explicit that it must return a truthy value (eg `"0"` and not `0`).

krbz999 (changes requested):
Not seeing much happening here.

Image

krbz999 (commented):
For now let's restrict this to just the ToC changes, not the Notes.

PR #2089, opened 2026-08-30
[#2021] Apply effects to fresh summons

JPMeehan:
Summon Choice advancements can now specify level-gated effects to potentially apply to the summons, based on the summoner's level. This handles both "You use your own characteristics where a minion’s stat block refers to an R or uses a potency (such as m<w)." as well as the 4, 7, and 10 Minion Improvement features.

image
image

### DIFF
--- a/src/module/canvas/layers/tokens.mjs
+++ b/src/module/canvas/layers/tokens.mjs
@@ -69,7 +69,10 @@ export default class DrawSteelTokenLayer extends foundry.canvas.layers.TokenLaye
       await tokenDocument.actor.updateEmbeddedDocuments("ActiveEffect", oldEffects);
       await tokenDocument.actor.createEmbeddedDocuments("ActiveEffect", newEffects, { keepId: true });
     } else {
-      tokenDocument.delta.updateSource(actorUpdates);
+      if (!foundry.utils.isEmpty(actorUpdates)) {
+        tokenDocument.updateSource({ delta: {} });
+        tokenDocument.delta.updateSource(actorUpdates);
+      }
       // Foundry will automatically increment but we need to add in our indices
       if (actor.prototypeToken.appendNumber) {
         const regex = new RegExp(/\((\d+)\)$/);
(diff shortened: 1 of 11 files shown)


src/module/canvas/layers/tokens.mjs
```diff
@@ -69,7 +69,10 @@ export default class DrawSteelTokenLayer extends foundry.canvas.layers.TokenLaye
       await tokenDocument.actor.updateEmbeddedDocuments("ActiveEffect", oldEffects);
       await tokenDocument.actor.createEmbeddedDocuments("ActiveEffect", newEffects, { keepId: true });
     } else {
-      tokenDocument.delta.updateSource(actorUpdates);
+      if (!foundry.utils.isEmpty(actorUpdates)) {
+        tokenDocument.updateSource({ delta: {} });
+        tokenDocument.delta.updateSource(actorUpdates);
+      }
```
JPMeehan, line 75: At some point the v14 delta nullability changes silently broke this piece of code. We didn't implement token placement like this outside of API/Macros so nobody had noticed.

src/module/data/item/ability.mjs
```diff
@@ -956,6 +957,19 @@ export default class AbilityModel extends BaseItemModel {
...
-    return canvas.tokens.placeActor(worldActor, { count });
+    const actorUpdates = { effects: [] };
+
+    const replacementData = this.parent.getRollData();
+
+    for (const e of effects) {
+      const data = game.items.fromCompendium(e, { keepId: true, clearFolder: true });
+      for (const change of data.system.changes) {
+        if (typeof change.value !== "string") continue;
+        change.value = DrawSteelActiveEffect.implementation._replaceDataRefs(change.value, replacementData);
```
```suggestion
        change.value = DrawSteelActiveEffect._replaceDataRefs(change.value, replacementData);
```

krbz999 (approved):

src/module/data/item/ability.mjs
krbz999: This needs to be updated.

PR #2091, opened 2026-08-30
Basic Roll Card Styling

jalensailin:
### This MR does the following:

 - Spruces up the basic roll cards to have a bit more DS character.
 - Adds the `expanded` class to roll `dice-roll` elements so that the inline tooltip is always displayed.
   - This tooltip is also rearranged so that only the dice formula and the dice results are shown (hiding the total since it is already shown)
   - Removes the ability to collapse these inline tooltips.

### Examples:
   
**Note:** I am sure that I missed roll cards / states that I just don't know how to trigger. Please let me know what these are if they don't look good with the new styling. 

image

image

image

image

image

### DIFF
--- a/src/styles/system/applications/applications.css
+++ b/src/styles/system/applications/applications.css
@@ -29,6 +29,7 @@
   }
 
   dl.power-roll-display {
+    align-items: center;
     gap: 4px;
 
     dt.tier1,
--- a/src/styles/system/applications/sidebar/tabs/chat.css
+++ b/src/styles/system/applications/sidebar/tabs/chat.css
@@ -9,7 +9,7 @@
 
     /* "standard" chat messages that use parts */
     &.draw-steel {
-      >section:not(:last-of-type) {
+      > section:not(:last-of-type) {
         border-bottom: var(--sidebar-separator);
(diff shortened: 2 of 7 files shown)


JPMeehan:
I think we're missing a saving throw? Need a save ends effect and then it's an option in the effect's context menu

JPMeehan:
Current saving throw aesthetic. We can improve that.

image

JPMeehan:
OK, on review the "issue" is a lack of flavor being passed to the saving throw part, which we can separately fix. Otherwise, my gut here is we should *keep* the toggled expansion for projects & saving throws, there's not a lot being gained by forcing them to be expanded if we're otherwise keeping this formatting.

image

JPMeehan:
So the bug was elsewhere and now fixed in #2100 Here's what saving throws look like now

image

JPMeehan (commented):
OK, playing around I think my only change is to keep the expansion toggle for Saving Throws and Projects unless we also want to redo that roll HTML like I did with the damage rolls, and even if we chose to do that it can be a separate PR

JPMeehan (approved):

PR #2096, opened 2026-08-30
[#680] Origin Roll Data Parsing

### LINKED ISSUE #680: Consider parsing roll data in effect application
As a user, I would like my applied effects to vary in their change values based on actor data.

Proposed implementation: Attempt to parse each `change` value as roll data, if it successfully parses then evaluate and fill in the total.

This ticket would only cover sync evaluation; if there are any effects that need non-deterministic values (unlikely?) that should be a separate handling.

JPMeehan:
Added new property to use the origin's roll data to parse the change values, useful for things where the applied effect scales with the source's level or characteristics. Will add in compendium and documentation updates if the code is good.

Closes #680


src/module/applications/ux/enrichers/apply-effect.mjs
```diff
@@ -148,6 +149,13 @@ async function onClickAnchor() {
   const originActor = origin?.documentName === "Actor" ? origin : origin?.actor;
   let combatant = this.dataset.originDuration ? game.combat?.getCombatantsByActor(originActor)[0] ?? "" : null;
 
+  if (this.dataset.originRollData) {
+    const replacementData = origin.getRollData();
+    const changeUpdates = foundry.utils.deepClone(tempEffect.system._source.changes);
+    for (const change of changeUpdates) if (typeof change.value === "string") change.value = DrawSteelActiveEffect._replaceDataRefs(change.value, replacementData);
+    tempEffect.updateSource({ "system.changes": changeUpdates });
+  }
```
krbz999, line 156:
```suggestion
  if (this.dataset.originRollData) {
    const replacementData = origin.getRollData();
    const changeUpdates = foundry.utils.deepClone(tempEffect.system._source.changes);
    for (const change of changeUpdates)
      if (typeof change.value === "string")
        change.value = DrawSteelActiveEffect._replaceDataRefs(change.value, replacementData);
    tempEffect.updateSource({ "system.changes": changeUpdates });
  }
```

src/module/data/pseudo-documents/power-roll-effects/applied-effect.mjs
```diff
@@ -198,6 +199,13 @@ export default class AppliedPowerRollEffect extends BasePowerRollEffect {
       ? await DrawSteelActiveEffect.fromStatusEffect(effectId)
       : this.item.effects.get(effectId).clone({}, { keepId: noStack, addSource: true });
 
+    if (config.properties.has("originRollData")) {
+      const replacementData = this.document.getRollData();
+      const changeUpdates = foundry.utils.deepClone(tempEffect.system._source.changes);
+      for (const change of changeUpdates) if (typeof change.value === "string") change.value = DrawSteelActiveEffect._replaceDataRefs(change.value, replacementData);
+      tempEffect.updateSource({ "system.changes": changeUpdates });
+    }
```
krbz999, line 206:
```suggestion
    if (config.properties.has("originRollData")) {
      const replacementData = this.document.getRollData();
      const changeUpdates = foundry.utils.deepClone(tempEffect.system._source.changes);
      for (const change of changeUpdates)
        if (typeof change.value === "string")
          change.value = DrawSteelActiveEffect._replaceDataRefs(change.value, replacementData);
      tempEffect.updateSource({ "system.changes": changeUpdates });
    }
```

krbz999 (changes requested):
There's not really a reason this cannot be async.

In addition, having done this exact feature in dnd5e, I think the correct implementation would be additional data on the AE or its changes.

Lastly, there's duplicated code in this PR which could be factored out to a static method.

JPMeehan:
- I agree the process can be async, but nothing that's being done right now
- I don't strongly feel we need to support this on a change level, and the enricher + PRE solution is equivalent to the AE level given how it would be called
- Could definitely refactor to a single static method.

src/module/documents/active-effect.mjs
```diff
@@ -33,6 +34,23 @@ export default class DrawSteelActiveEffect extends foundry.documents.ActiveEffec
 
   /* -------------------------------------------------- */
 
+  /**
+   * Convert roll data properties in change data.
+   * @param {EffectChangeData[]} [changes]    Effect change data in json-serializable format.
+   * @param {object} [replacementData]        The target of the effect to be applied.
+   * @returns {Promise<EffectChangeData[]>}      A promise that resolves to the modified change data.
```
krbz999:
```suggestion
   * @param {EffectChangeData[]} [changes]    Effect change data in json-serializable format.
   * @param {object} [replacementData]        The target of the effect to be applied.
   * @returns {Promise<EffectChangeData[]>}   A promise that resolves to the modified change data.
```

krbz999 (approved):

PR #2099, opened 2026-08-31
[#2060] Store and forward spend amounts to enrichment

### LINKED ISSUE #2060: System Data: store per-spend-effect amounts on ability chat messages
### User Experience

As a module developer I would like individual spend amounts preserved on the ability chat message so I can reference them individually. Since Spend is now an effect type, and there can be multiple on one ability such as on Heart of the Beast for the Beastheart, it would be useful if they were stored individually on the chat message instead of being totaled together like they are now.

Proposed addition alongside the existing resourceSpend block:
```
// AbilityModel.use()
messageData.system.ability.spendAmounts = this.power.effects.documentsByType.spend.map(e => fd.spend?.[e.id] ?? 0);
```
Note: `spendAmounts` would need to be added to the appropriate message data model schema, the exact implementation is up to you, as long as the data is accessible on the resulting message.

JPMeehan:
- Ability Use messages now store the amount of each spend
- This info is forwarded to enrichment
- This required a slight adjustment to our `enrichHTML` utility method to only override properties rather than fully clobber the incoming `rollData`. This should be safe within our own logic but theoretically has side effects if people were passing in roll data they expected to be fully clobbered.

image

Closes #2060

### DIFF
--- a/src/docs/Roll-Data.md
+++ b/src/docs/Roll-Data.md
@@ -81,7 +81,7 @@ Most actors support the following roll attributes
 
 ### Statuses
 Value indicates if actor currently has status (1/yes, 0/no). Use with multiplication.
-+ Asleep:`@statuses.sleep`
++ Asleep: `@statuses.sleep`
 + Bleeding: `@statuses.bleeding`
 + Burning: `@statuses.burning`
 + Dazed: `@statuses.dazed`
@@ -144,5 +144,5 @@ Value indicates if actor currently has status (1/yes, 0/no). Use with multiplica
 ### Spend Info
 
 The amount of heroic resource/malice spent on an ability is available in both the Power Roll Effect as well as the text for the spend.
-- In the power roll effect, use `@spend.0` to reference the amount spent on the first spend effect, `@spend.0` for the second, etc.
+- In the power roll effect, use `@spend.0` to reference the amount spent on the first spend effect, `@spend.1` for the second, etc.
 - In the text of the individual effect, `@spend` is the amount spent on that effect.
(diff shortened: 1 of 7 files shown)


src/docs/Roll-Data.md
```diff
@@ -80,7 +80,7 @@ Most actors support the following roll attributes
 </details>
 
 ### Statuses
-Value indicates if actor currently has status (1/yes, 0/no)
+Value indicates if actor currently has status (1/yes, 0/no). Use with multiplication.
 + Asleep:`@statuses.sleep`
```
krbz999:
```suggestion
Value indicates if actor currently has status (1/yes, 0/no). Use with multiplication.
+ Asleep: `@statuses.sleep`
```

src/docs/Roll-Data.md
```diff
@@ -139,5 +139,10 @@ Value indicates if actor currently has status (1/yes, 0/no)
 + Ability Power Roll Characteristic: `@item.powerRoll.characteristics`
 + Ability Power Roll Characteristic formula: `@item.powerRoll.formula`
 + Ability Heroic Resource/Malice Cost:`@item.resource`
-+ Ability Additional Heroic Resource/Malice Cost: `@item.spend.value`
 + Number of targets: `@item.target.value`
+
+### Spend Info
+
+The amount of heroic resource/malice spent on an ability is available in both the Power Roll Effect as well as the text for the spend.
+- In the power roll effect, use `@spend.0` to reference the amount spent on the first spend effect, `@spend.0` for the second, etc.
```
krbz999: Check that latter one.

krbz999 (approved):

PR #2101, opened 2026-09-01
[#583] Added summoning button to ability use messages

JPMeehan:
I think we're in the home stretch. This implements a generalized way for special effects to add buttons, taken advantage of by our new summoning effects. Like most data-action tricks, modules just need to add to the part's `static ACTIONS`

image

### DIFF
--- a/templates/sidebar/chat/parts/ability-use.hbs
+++ b/templates/sidebar/chat/parts/ability-use.hbs
@@ -1,6 +1,7 @@
 <div class="message-part-html">
   {{{ctx.embed.outerHTML}}}
 </div>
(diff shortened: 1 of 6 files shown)


templates/sidebar/chat/parts/ability-use.hbs
```diff
@@ -1,6 +1,7 @@
 <div class="message-part-html">
   {{{ctx.embed.outerHTML}}}
 </div>
+{{log ctx.buttons}}
```
krbz999: Excuse me

krbz999 (approved):

PR #2109, opened 2026-09-03
[#2102] Shifted start of turn heroic resource calculation

### LINKED ISSUE #2102: Shift HR production to actor property
1. Set up HR/turn as a non-persisted field that is initialized by the class value 
2. Level 7 features can have an AE that gives +1 to that value
3. Persistent can be a temporary effect that reduces that value (implemented via the button method introduced in #2101)

JPMeehan:
- Implemented `hero.primary.turnGain` as a non-persisted formula, which is initialized as the value from the hero's class
- This allows AEs to target the property to increase or decrease it, cleanly implementing the various level 7 "+1 bonus" features on all classes.
- This also enables Persistent effects to cleanly reduce the elementalist's HR gain, with a button. Button itself dependent on #2101 which creates the general button infrastructure, but the API method of `applyPersist` can be tested independently. This does not automate the unique damage trigger that can end the effect early.

image

Closes #2102


src/module/data/pseudo-documents/special-effect/persistent-effect.mjs
```diff
@@ -34,4 +38,47 @@ export default class PersistentSpecialEffect extends BaseSpecialEffect {
...
+   */
+  async applyPersist() {
+    /** @type {DrawSteelActor} */
+    const parent = this.document.actor;
+    if (!parent) throw new Error("No parent actor found to create Persistent AE!");
+
+    return foundry.documents.ActiveEffect.create({
+      name: this.label,
+      img: this.document.img,
+      description: "<p>If you take damage equal to our greater than [[lookup 5*@R evaluate]]{5 times your Reason score} in one turn, you stop maintaining any persistent abilities.</p>",
```
JPMeehan: I'm slightly unsure how to handle this description, it's HTML so maybe I should defer to an escaped i18n string?

krbz999: ~~our~~ or

and should go into the i18n file if possible.

I believe `_loc` does some(?) cleaning.

krbz999 (approved):

PR #2120, opened 2026-09-04
[#1093] Ability Modifier matching by DSID

### LINKED ISSUE #1093: Ability modifiers for specific abilities
As a user, I would like features that improve core maneuvers and actions, e.g. the Fury's bonuses to the Knockback maneuver, to be able to apply those bonuses automatically.

Implementation idea: New ability modifier filter for `dsid`

JPMeehan:
Added new filter field to Ability Modifiers so they can target specific abilities, e.g. "You gain an edge on the Grab and Knockback maneuvers"

image

Closes #1093


krbz999 (commented):
I think this might also be made redundant by a filter as per my comment in #2121.

PR #2121, opened 2026-09-04
[#702] Added suppression logic for HR levels

### LINKED ISSUE #702: Implement Effect Conditionals

JPMeehan:
Several classes have features that are based on how much of their HR they're holding onto. 

> As you advance in your chosen null tradition, you gain certain benefits in combat, including benefits based on the amount of discipline you have. Benefits based on how much discipline you have last until the end of your turn, even if a benefit would become unavailable to you because of the amount of discipline you spend during your turn.

This also sets up the Beastheart, whose companion has Rampage which is updated slightly differently (which can be handled at the data model level).

> Your companion has a resource called rampage. Whenever you or your companion spends ferocity, your companion gains rampage equal to the ferocity spent. Your companion loses their rampage and its effects at the end of an encounter.

image

Closes #702

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -1944,6 +1944,7 @@
     "ActiveEffect": {
       "Name": "Effect",
       "Filters": "Filters",
+      "Requirements": "Requirements",
       "StatusConditions": "Status Conditions",
       "ComplexStatus": "This condition cannot be toggled here because it is added by a non-standard effect. Open the actor sheet to adjust.",
       "Source": "Source",
@@ -2029,6 +2030,12 @@
             },
             "roll": {
               "label": "Saving Throw Formula"
+            }
+          },
+          "requirements": {
+            "resource": {
+              "label": "Resources",
+              "hint": "If a hero has less of their heroic resource than this the effect will be suppressed."
             }
           },
           "project": {
--- a/src/module/data/effect/base-effect.mjs
+++ b/src/module/data/effect/base-effect.mjs
@@ -6,6 +6,10 @@ import enrichHTML from "../../utils/enrich-html.mjs";
 import { setOptions } from "../helpers.mjs";
 
 /**
+ * @import { CoreResource } from "../actor/_types";
+ */
+
+/**
  * A data model used by default effects with properties to control the expiration behavior.
  */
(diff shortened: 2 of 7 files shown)


src/module/data/effect/base-effect.mjs
```diff
@@ -38,6 +42,10 @@ export default class BaseEffectModel extends foundry.data.ActiveEffectTypeDataMo
       roll: new FormulaField({ initial: "1d10 + @combat.save.bonus" }),
     });
 
+    schema.requirements = new fields.SchemaField({
```
krbz999: Were we to implement a filter class like done in dnd5e, black flag, ryuutama, et al, this would be made redundant.

Perhaps we are not ready for that just yet, or perhaps this is so often used that it deserves its own field and handling but I don't think that's the case.

lang/en.json
```diff
@@ -2031,6 +2032,12 @@
               "label": "Saving Throw Formula"
             }
           },
+          "requirements": {
+            "resource": {
+              "label": "Resources",
+              "hint": "If a hero has less of their heroic resource than this the effect will be suppressed."
```
krbz999:
```suggestion
              "hint": "The effect is suppressed if a hero's heroic resource is below this value."
```

PR #2123, opened 2026-09-04
[#604] Factored in Highground

### LINKED ISSUE #604: Add automation for "High Ground" bonuses
Multifold task:

- Implement high ground
Quote from heroes document p. 174
> HIGH GROUND
> When you use an ability against a creature or object while standing on ground and occupying a space that is fully above the space the target takes up, with the bottom of your space higher than the top of the target’s space, you gain an edge on the power roll against that target. You can get this the benefit while climbing if your speed has the Climb keyword.
- Implement elevation change via regions
The module [Terrain Mapper](https://github.com/caewok/fvtt-terrain-mapper) does something like this already, but given the inherent Verticality of Draw Steel, maybe that should be core?

Lusunati:
High Ground Advantage checks the difference between two elevations is bigger or equal to the actor's size. If it's higher for the originating attack and the actor isn't flying it adds an edge. Resolves #604

### DIFF
--- a/src/module/data/item/ability.mjs
+++ b/src/module/data/item/ability.mjs
@@ -772,7 +772,8 @@ export default class AbilityModel extends BaseItemModel {
       // Flanking checks
       if (this.keywords.has("melee") && this.keywords.has("strike") && token.isFlanking(target)) modifiers.edges += 1;
       // High Ground Check - targeting (Target's Size) Squares Lower gets an edge.
-      if ((token.document.elevation - target.document.elevation) >= targetActor.system.combat.size.value) modifiers.edges += 1;
+      
+      if (((token.document.elevation - target.document.elevation) >= targetActor.system.combat.size.value) && !(this.actor.statuses.has("flying"))) modifiers.edges += 1;
     }
 
     // Modifiers requiring both a controlled token and the targeted token to have an actor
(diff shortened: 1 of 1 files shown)


src/module/data/item/ability.mjs
```diff
@@ -771,6 +771,9 @@ export default class AbilityModel extends BaseItemModel {
     if (token) {
       // Flanking checks
       if (this.keywords.has("melee") && this.keywords.has("strike") && token.isFlanking(target)) modifiers.edges += 1;
+      // High Ground Check - targeting (Target's Size) Squares Lower gets an edge.
+      
+      if (((token.document.elevation - target.document.elevation) >= targetActor.system.combat.size.value) && !(this.actor.statuses.has("flying"))) modifiers.edges += 1;
```
JPMeehan: We should use the actual `target.document.depth` instead of the actor size since it is possible for a token to be extra tall or short (e.g. I believe high ground is valid for objects)

JPMeehan (approved):

src/module/data/item/ability.mjs
Lusunati: Alright, tried to implement that - hopefully it slots 1-1?

PR #2135, opened 2026-09-05
#2073 Summoner Class, Circle of Blight and Some Active Effects

Lusunati:
Circle of Blight needs it's Fixture Ability Updated when we can use the relevant enricher.


JPMeehan (approved):
LGTM

PR #2144, opened 2026-09-07
#2133 Growing ferocity on existing classes

### LINKED ISSUE #2133: Add tracking text to Growing Ferocity style features
`[[lookup @hero.primary.tracking]]` will produce an enricher tracking the "current" heroic resource amount through the end of turn for the table value. We should add it to the top of growing ferocity style functions so people can see it "live"

Lusunati:
Implements #2133


JPMeehan (approved):
LGTM

Issue #1930, opened 2026-05-29
Falling Automation


JPMeehan:
A creature without appropriate movement types and/or statuses should automatically fall at the end of its turn.

Lusunati:
Potentially implemented in #2156

PR #2156, opened 2026-09-13
[#1930] Falling automation

Lusunati:
If a token cannot climb, fly or hover innately (and conditions for falling are met if flying) - the token's height is now set to 0. 

Since there is no way to account for edge cases such as a stopped climb without a climb speed (such as clinging to a wall from might), it might be before this automation can be implemented, tokens should have some way to toggle their climbing/swimming state. (isTraversing perhaps).


src/module/data/actor/base-actor.mjs
JPMeehan, line 1: Do the other classes properly call `super`?

src/module/data/actor/base-actor.mjs
```diff
@@ -368,9 +368,30 @@ export default class BaseActorModel extends DrawSteelSystemModel {
...
+
+      const tokens = this.parent.getActiveTokens();
+      
+      tokens.forEach((token) => {
+        if (!token.visible || token.document.isSecret) {
+          return;
+        }
+        if (token.document.elevation > 0) {
+          token.document.update({ elevation: 0 });
+        }
```
JPMeehan: 1. Doesn't account for multi-level maps and surfaces at all. There's some token helper method to determine the nearest surface, I believe 5e has this part figured out
2. Unfortunately this probably needs to be a bigger lift; the recommended way in Foundry to handle "standing on a slope/tall object" is to just raise the elevation *without* building out surfaces, which we can't truly distinguish. A better solution is to create a chat message with a button that handles the falling (+ then we can append a roll part for the damage)

JPMeehan (changes requested):

src/module/data/actor/base-actor.mjs
Lusunati: Seems so! I tested it with an NPC, a D.T.O and a Hero. This is all called already from `combat.mjs`, so all it needs to call super for is if it changes how _onEndTurn() works to begin with (like the hero type does)

src/module/data/actor/base-actor.mjs
Lusunati: Checking that out presently, thanks.

src/module/data/actor/base-actor.mjs
```diff
@@ -368,9 +372,74 @@ export default class BaseActorModel extends DrawSteelSystemModel {
...
+          await this.processFall(token, distance);
+        }
+        
+      });
+    }
+  }
+
+  /* -------------------------------------------------- */
+  /**
+   * Taking a Surface from elsewhere, post a chat message trying to get permission to fall. If granted post applicable damage.
```
JPMeehan: This seems to reflect an old version of the method

src/module/data/actor/base-actor.mjs
```diff
@@ -368,9 +372,74 @@ export default class BaseActorModel extends DrawSteelSystemModel {
...
+   * @param {DrawSteelToken} token the Token that is falling.
+   * @param {number} dist The distance straight down the actor is falling.
```
JPMeehan:
```suggestion
   * @param {DrawSteelToken} token The Token that is falling.
   * @param {number} dist          The distance straight down the actor is falling.
```

src/module/data/actor/base-actor.mjs
```diff
@@ -368,9 +372,74 @@ export default class BaseActorModel extends DrawSteelSystemModel {
...
+
+    let fallDamage = 0;
+    if (dist > 2)
+    {
+      fallDamage = Math.min(2 * dist, 50);
+    }
+    
+    const roll = new DamageRoll(String(fallDamage), {
+    });
+    await roll.evaluate();
```
JPMeehan:
```suggestion
    const fallDamage = dist >= 2 ? 0 : Math.min(2 * dist, 50);
    
    const roll = new DamageRoll(String(fallDamage), {});
```
1. Code styling adjustments to minimize new lines
2. Falling is "2 or more" squares of falling.

src/module/data/actor/base-actor.mjs
```diff
@@ -368,9 +372,74 @@ export default class BaseActorModel extends DrawSteelSystemModel {
...
+      type: "standard",
+      speaker: DrawSteelChatMessage.getSpeaker({ actor: this.parent }),
+      "system.parts": [{ type: "falling",
+        rolls: [roll],
+        flavor: _loc("DRAW_STEEL.ChatMessage.PARTS.falling.Label"),
+        fallerUuid: token.document.uuid,
+        fallerDistance: dist,
+        fallerDamage: fallDamage,
+      }],
+
```
JPMeehan:
```suggestion
      flags: { core: { canPopout: true } },
```
Always add in the popout logic to system created chat messages

src/module/data/pseudo-documents/message-parts/fall.mjs
```diff
@@ -0,0 +1,92 @@
...
+    fall: this.#handleFall,
+  };
+  /* -------------------------------------------------- */
+  
+  /** @inheritdoc */
+  static defineSchema() {
+    return Object.assign(super.defineSchema(), {
+      fallerUuid: new DocumentUUIDField({ nullable: false, type: "Token" }),
+      fallerDistance: new NumberField(),
+      fallerDamage: new NumberField(),
```
JPMeehan: Structurally, I'm not sure why the damage needs to be stored, we already have the constructed roll and can reference its total

src/module/data/pseudo-documents/message-parts/fall.mjs
```diff
@@ -0,0 +1,92 @@
...
+    applyDamage: (event) => DamageRoll.applyDamageCallback(event),
+    fall: this.#handleFall,
+  };
+  /* -------------------------------------------------- */
+  
+  /** @inheritdoc */
+  static defineSchema() {
+    return Object.assign(super.defineSchema(), {
+      fallerUuid: new DocumentUUIDField({ nullable: false, type: "Token" }),
+      fallerDistance: new NumberField(),
```
JPMeehan, line 33: This should probably have some options passed into it, e.g. integer & min. Might even be a case of our requiredInteger helper?

src/module/data/pseudo-documents/message-parts/fall.mjs
```diff
@@ -0,0 +1,92 @@
...
+ 
+  /** @inheritdoc */
+  static get TYPE() {
+    return "falling";
+  }
+  
+  /** @inheritdoc */
+  static ACTIONS = {
+    ...super.ACTIONS,
+    applyDamage: (event) => DamageRoll.applyDamageCallback(event),
```
JPMeehan: Don't need to restate the applyDamage action since it's in `super.ACTIONS`, except that actually I would look at what `TargetResultPart` does to ensure that the damage is applied to the relevant (falling) actor

src/module/data/pseudo-documents/message-parts/fall.mjs
JPMeehan, line 1: Need the dividers ---

src/module/data/actor/base-actor.mjs
```diff
@@ -368,9 +372,74 @@ export default class BaseActorModel extends DrawSteelSystemModel {
...
+        //This is where our logic for getting permission to fall comes in.
+        if (distance > 0) {
+          await this.processFall(token, distance);
+        }
+        
+      });
+    }
+  }
+
+  /* -------------------------------------------------- */
```
JPMeehan, line 410:
```suggestion
  /* -------------------------------------------------- */

```

src/module/data/pseudo-documents/message-parts/fall.mjs
```diff
@@ -0,0 +1,92 @@
...
+  get token() {
+    return fromUuidSync(this.fallerUuid) ?? null;
+  }
+
+  /** @inheritdoc */
+  async _prepareContext(context) {
+    await super._prepareContext(context);
+    context.ctx.fallWarning = _loc("DRAW_STEEL.ChatMessage.PARTS.falling.explain", { victim: this.token.name, distance: this.fallerDistance });
+    
+    context.ctx.buttons.push(
```
JPMeehan, line 60: Should only be able to fall if you own the token

src/module/data/pseudo-documents/message-parts/fall.mjs
```diff
@@ -0,0 +1,92 @@
...
+    );
+
+  }
+
+  /**
+   * Perform a token's fall.
+   * @this AbilityUsePart
+   * @param {PointerEvent} event   The originating click event.
+   */
+  static async #handleFall(event) {
```
JPMeehan, line 77: I think it would be better if this added a new part to the existing parts of the parent message

src/module/documents/token.mjs
```diff
@@ -105,6 +105,90 @@ export default class DrawSteelTokenDocument extends foundry.documents.TokenDocum
     else if (movement.passed.waypoints.every(wp => wp.action === "burrow")) this.actor?.toggleStatusEffect("burrow", { active: true });
   }
 
+  /* -------------------------------------------------- */
```
JPMeehan, line 111:
```suggestion
  /* -------------------------------------------------- */

```

src/module/documents/token.mjs
```diff
@@ -105,6 +105,90 @@ export default class DrawSteelTokenDocument extends foundry.documents.TokenDocum
...
+  /**
+   * Find the supporting surface this token rests on or would fall onto, and the level it comes to rest on. A scene that
+   * defines any movement surface uses those surfaces as its only floors. As a heuristic to accommodate older scenes
+   * without levels or surfaces, the base of each level is considered a floor in scenes with no surfaces.
+   * @param {object} [options]
+   * @param {TokenCoordinates} [options.position]  The position to evaluate against. Defaults to the token's source
+   *                                               position.
+   * @returns {{ elevation: number, region: RegionDocument|null, level: Level }|null}
```
JPMeehan: define both options & return type & extract to an interface in the nearest _types file

src/module/documents/token.mjs
```diff
@@ -105,6 +105,90 @@ export default class DrawSteelTokenDocument extends foundry.documents.TokenDocum
...
+  /* -------------------------------------------- */
+
+  /**
+   * Resolve the level a token comes to rest on when landing on a surface region at a given elevation. Checks every
+   * level the surface region belongs to. The result is the single candidate whose elevation range is home to the
+   * landing elevation, or the current level when there is no unambiguous home.
+   * @param {RegionDocument} region  The landed surface's region.
+   * @param {number} elevation       The landing elevation.
+   * @param {string} levelId         The token's current level ID.
+   * @returns {Level|null}           The level the token rests on.
```
JPMeehan, line 179: I suspect Level & RegionDocument are not actually imported into this file at the top

JPMeehan (changes requested):

src/module/data/pseudo-documents/message-parts/fall.mjs
```diff
@@ -0,0 +1,110 @@
...
+    if (!this.actorTarget.isOwner) return;
```
JPMeehan: My intention was to gate at the button display, so not pushing the button — if you use optional chaining it has the benefit of double checking the actor still exists (people delete unlinked tokens all the time)

src/module/data/pseudo-documents/message-parts/fall.mjs
```diff
@@ -0,0 +1,110 @@
...
+   * @type {DrawSteelToken | null}
+   */
+  get token() {
+    return fromUuidSync(this.fallerUuid) ?? null;
+  }
+
+  /** @inheritdoc */
+  async _prepareContext(context) {
+    await super._prepareContext(context);
+    context.ctx.fallWarning = _loc("DRAW_STEEL.ChatMessage.PARTS.falling.explain", { victim: this.token.name, distance: this.fallerDistance });
```
JPMeehan: `this.token` can return null if the token was deleted

src/module/data/pseudo-documents/message-parts/fall.mjs
```diff
@@ -0,0 +1,110 @@
...
+   * @this AbilityUsePart
+   * @param {PointerEvent} event   The originating click event.
+   */
+  static async #handleFall(event) {
+    if (!this.actorTarget.isOwner) return;
+    const token = this.token;
+    const surface = await token._findSupportingSurface();
+    token.update({ elevation: surface.elevation });
+
+    console.log(this.parent.parts.get());
```
JPMeehan: Zap it

src/module/data/pseudo-documents/message-parts/fall.mjs
```diff
@@ -0,0 +1,110 @@
...
+
+    const eventText = "<br>" + _loc("DRAW_STEEL.ChatMessage.PARTS.falling.aftermath", { victim: this.token.name, distance: this.fallerDistance, damage: this.rolls[0].formula });
+    const updates = { content: eventText };
+
+    const parts = { ...this.parent.parts };
+    const contentPart = new ContentPart({ type: "content", flavor: _loc("DRAW_STEEL.ChatMessage.PARTS.falling.takeFall", { distance: this.fallerDistance }) });
+    parts[contentPart._id] = contentPart;
+    updates.system = { parts };
```
JPMeehan: You should definitely be able to use the `static create` method here, I forget at the moment if the parent is the model or the document

src/module/data/pseudo-documents/message-parts/fall.mjs
```diff
@@ -0,0 +1,110 @@
...
+  /**
+   * Perform a token's fall.
+   * @this AbilityUsePart
+   * @param {PointerEvent} event   The originating click event.
+   */
+  static async #handleFall(event) {
+    if (!this.actorTarget.isOwner) return;
+    const token = this.token;
+    const surface = await token._findSupportingSurface();
+    token.update({ elevation: surface.elevation });
```
JPMeehan: One extra thing we should be using `move` not just a plain update and using the `forced` movement action

src/module/data/actor/base-actor.mjs
```diff
@@ -368,9 +372,70 @@ export default class BaseActorModel extends DrawSteelSystemModel {
...
+    const fallDamage = dist <= 2 ? 0 : Math.min(2 * dist, 50);
+    
+    const roll = new DamageRoll(String(fallDamage), {});
+    await roll.evaluate();
+
+    await DrawSteelChatMessage.create({
+      title: _loc("DRAW_STEEL.ChatMessage.PARTS.falling.Label"),
+      type: "standard",
+      speaker: DrawSteelChatMessage.getSpeaker({ actor: this.parent }),
+      content: "<br>" + _loc("DRAW_STEEL.ChatMessage.PARTS.falling.aftermath", { victim: this.name, distance: fallDamage, damage: roll.formula }),
```
JPMeehan: Probably don't need a `<br>`, would rather just use some css to manage padding/margins. Maybe try wrapping in `<p>` tags, idk if that's automatic.

src/module/data/pseudo-documents/message-parts/fall.mjs
```diff
@@ -0,0 +1,105 @@
...
+   * Perform a token's fall.
+   * @this AbilityUsePart
+   * @param {PointerEvent} event   The originating click event.
+   */
+  static async #handleFall(event) {
+    const token = this.token;
+    const surface = await token._findSupportingSurface();
+
+    await token.move({ action: "forced", elevation: surface.elevation });
+    await ContentPart.create({ type: "content", flavor: _loc("DRAW_STEEL.ChatMessage.PARTS.falling.takeFall", { distance: this.fallerDistance }) }, { parent: game.messages.get(this.message.id) });
```
JPMeehan:
```suggestion
    await ContentPart.create({ type: "content", flavor: _loc("DRAW_STEEL.ChatMessage.PARTS.falling.takeFall", { distance: this.fallerDistance }) }, { parent: this.message });
```
Already have pointer to message, no need to use id to fetch it

JPMeehan (commented):
Last, minor changes. I'll test when I'm home then should be good to merge.

JPMeehan (approved):

PR #2158, opened 2026-09-15
Changed grid on dl items to block display inside of the chat messages.

### LINKED ISSUE #2124: Improve whitespace usage and wrapping for special effects
Special effects take up a lot of whitespace for their labels. On the item sheet this isn't a crisis but it's very inefficient in the chat log

Image

Image

Katiedead:
Fixes #2124 
image


JPMeehan:
I think we probably want that ability metadata (range, targets) to still leverage some of this, especially with the glyphs

PR #2159, opened 2026-09-15
Properly scoped typography to .draw-steel

Katiedead:
Scoped typography file to `.draw-steel`.


src/styles/variables/typography.css
```diff
@@ -12,7 +12,7 @@
   --font-h1-size: 2em;
   --font-h2-size: 1.5em;
   --font-h3-size: 1.25em;
-  --font-h4-size: 1em;
+  --font-h4-size: 1.5em;
```
krbz999, line 15: Intentional?

PR #2162, opened 2026-09-16
[#2073] Summoner titles

Lusunati:
Adding Titles for #2073


JPMeehan (commented):
I think the formatting of the actual titles doesn't match what we've been doing? Especially since I think we've preferred embeds so you can properly pick which benefit as an advancement

JPMeehan (approved):

Issue #2127, opened 2026-09-04
The Beastheart Compendium


Lusunati:
### Powers
- [x] Core Powers
- [x] Level 1 Powers
- [x] Level 2 Powers
- [x] Level 3 Powers
- [x] Level 5 Powers
- [x] Level 6 Powers
- [x] Level 8 Powers
- [x] Level 9 Powers
### Features
- [x] Echelon 1 Features
- [x] Echelon 2 Features 
- [x] Echelon 3 Features
- [x] Echelon 4 Features
### Companions
- [x] Basilisk
- [x] Bear
- [x] Boar
- [x] Condor
- [x] Deinonychus
- [x] Drake
- [ ] Elemental Spark
- [ ] Gummy Ball
- [ ] Hellhound
- [ ] Lightbender
- [ ] Panther
- [ ] Spider
- [ ] Sporeling
- [ ] Wolf

Lusunati:
#2128 - Core and Level 1 Powers

PR #2166, opened 2026-09-16
[#2127] Beastheart Wire-up

Lusunati:
Adding and "Wiring Up" the beastheart. The only thing that cannot presently be set up is the companion selection and the companion's advancements. This will have to be done at a later date.


JPMeehan (approved):
LGTM

PR #2167, opened 2026-09-17
[#2039] Make changes to dice for tests apply to ability power rolls

### LINKED ISSUE #2039: Bonuses to all power rolls for an actor
We need to carve out a data path for adjustments to *all* power rolls from a target, affecting both tests and abilities.

- Fate Domain "Blessing of Fate and Destiny"
- Censor "Fulfill your Destiny"

eddbc:
This is a quick change for issue [#2039] that makes it so that effects that change the dice for a characteristic test also affect power rolls from abilities that use that characteristic. 

It moves the existing logic used by tests for getting the updated formula to somewhere it can be referenced by both abilities and tests. This should make it so that compendium changes can implement abilities such as the Conduits 'Blessing of Fate and Destiny'.

I've not worked on Foundry modules or systems before, so it's possible I missed something obvious, but from some quick testing it appears to do the job.


JPMeehan:
Except the Fury doesn't have bonuses to all power rolls with Might and Agility, they only have these bonuses to tests. A new set of keys is necessary.

eddbc:
Ahhh, yeah you're right. I saw that the Fury had some roll3kh abilities but missed that they only applied to tests.

Looking at it a bit closer, it looks like there are a few different ways roll3kh/kl rules are done.

Conduit does it for all power rolls
Fury does it for tests with a specific characteristic or skill
Tactician does it for ability rolls against a marked target

If a way to add the effect to ability rolls specifically was added, then "Blessing of Fate and Destiny" for the conduit could be done using a combination of that and the existing characteristic modifiers, and also the tactician might be able to be done as well combining that with whatever logic is used for applying edges/banes to marked targets? The only one that would be left still unsupported would be "Menagerie" for the Fury that targets tests with the tracking skill.

JPMeehan:
Yes, a new section of ability modifier handling for the power roll would be appropriate in combination with some non-persisted fields.

eddbc:
I've taken another pass at this, and done it in a way that adds power.roll.dice.number and power.roll.dice.mode change keys to ability modifiers. I've done some quick testing and the ability rolls and tests now get the bonuses separately. It does mean that for "Blessing of Fate and Destiny" it needs separate effects applying for the bonus/penalty to tests and abilities, which is a little clunky, but doing it another way would have probably needed bigger changes to how these modifiers are handled

JPMeehan:
Could you also update src/docs/Active-Effects in this PR + the compendium update? Structure here looks good

eddbc:
Added compendium changes for the Conduit. I was hoping that this change would also be enough to get the Tactician's "Warmaster" ability to be able to work, but I think that will require additional work, maybe something related to filters?

JPMeehan:
Take a look at getTargetModifiers in the AbilityModel, as that would be the appropriate place to fetch the Warmaster modification

src/docs/Active-Effects.md
```diff
@@ -276,6 +277,8 @@ Ability modifiers support a far more limited set of keys, but still use the same
 |`potency`|Give a bonus to all potency values|
 |`power.roll.banes`|Add banes that apply to the ability's power roll regardless of target|
 |`power.roll.edges`|Add edges that apply to the ability's power roll regardless of target|
+|`power.roll.dice.number`|Adjust the number of dice rolled, defaulting to keep highest 2. (Use the UPGRADE type)|
+|`power.roll.dice.mode`|When the number of dice is more than 2, set to `kh` or `kl` to detemine if Keep Highest 2 or Keep Lowest 2 should be used|
```
JPMeehan:
```suggestion
|`power.roll.dice.mode`|When the number of dice is more than 2, set to `kh` or `kl` to detemine if Keep Highest 2 or Keep Lowest 2 should be used. `kh` is the default.|
```

JPMeehan (approved):

PR #2173, opened 2026-09-19
[#2007] Minion stamina marks

### LINKED ISSUE #2007: Show stamina marks for minion thresholds
### User Experience

As a player and game master, I would like if minion squads showed marks at each of the minion thresholds, now that the winded/dying thresholds are shown by default on other creatures.

Lusunati:
Adds stamina marks for minions in a manner akin to Winded and Dying creatures, as per #2007


src/module/canvas/placeables/token.mjs
```diff
@@ -240,4 +241,41 @@ export default class DrawSteelToken extends foundry.canvas.placeables.Token {
     const posY = index === 0 ? height - bh : 0;
     bar.position.set(0, posY);
   }
+  /**
```
JPMeehan, line 247: Divider

JPMeehan (approved):
Looks good just one stylistic adjustment

PR #2174, opened 2026-09-19
[#1899] Critical Threshold Modifiers

### LINKED ISSUE #1899: Crit Range Expansion code (for Devastating Imbued Weapon etc)
### User Experience

The Devastating Weapon enhancement and a few other things increase the crit range of abilities, it would be great if we had an Attribute Key we could adjust for that, thanks!

Lusunati:
Added a datapath for the Critical Threshold. It defaults to 19, and can be modified using active effects. Subtracting makes it easier to crit, adding makes it harder. Useful for the relevant imbue, as well as any other powers in Heroes that utilize it.


src/module/data/item/ability.mjs
```diff
@@ -90,6 +90,7 @@ export default class AbilityModel extends BaseItemModel {
         reactive: new fields.BooleanField(),
         formula: new FormulaField({ blank: true, initial: "@chr", placeholder: "@chr" }),
         characteristics: new fields.SetField(setOptions()),
+        criticalThreshold: new fields.NumberField({ initial: 19, min: 2, integer: true }),
```
JPMeehan: Probably also want nullable: false/required: true. I think non-persisted is correct but that makes me wonder if this should go in the UI or rely on programmatic adjustments (I guess programmatic adjustments is fine it's a *very* niche field but I do see design space for individual abilities that are more likely to crit)

src/module/data/item/ability.mjs
```diff
@@ -625,7 +626,7 @@ export default class AbilityModel extends BaseItemModel {
           targets.push(context.target);
           delete context.target;
         }
-        const roll = new PowerRoll(dialogConfig.context.formula, {}, { flavor: _loc(PowerRoll.TYPES.ability.label), ...context });
+        const roll = new PowerRoll(dialogConfig.context.formula, {}, { flavor: _loc(PowerRoll.TYPES.ability.label), criticalThreshold: this.power.roll.criticalThreshold, ...context });
```
JPMeehan: Line length is getting long, I think just break up the options object across multiple lines.

JPMeehan (commented):
Small adjustments

src/module/data/item/ability.mjs
Lusunati: For the moment it's a few abilities and can be simulated using this for 3PP. If there's a strong need, it can always be retrofitted later i suppose.

JPMeehan: Yeah, just make the nullability changes and if people want to make an individual ability they can do it programmatically without a field.

JPMeehan (approved):

PR #2177, opened 2026-09-21
[#2073] Devil detective (summoner retainer)

Lusunati:
The devil detective, sourced and enriched. Awaiting the Summon Enricher to complete.


JPMeehan (approved):
LGTM

PR #2178, opened 2026-09-21
[#2073] Rival summoners

Lusunati:
Adding the rival summoners for #2073. like #2177, needs to be enriched with the Summon Enricher before i'd be pleased to say it's Properly Done.


JPMeehan (approved):
LGTM

